PR #21, opened 2024-02-15
Check for empty token - error early

riya-n:


gnalh (approved):

src/main.rs
```diff
@@ -109,6 +109,11 @@ async fn run(cli: Cli) -> anyhow::Result<()> {
         |s| s,
     );
 
+    if token.trim().is_empty() {
+        log::error!("Trunk API token is required.");
+        return Ok(());
```
zaycev, line 114: Why do we return ok for error case?

zaycev:
Weird, this suggests that empty values are not  possible https://github.com/clap-rs/clap/issues/1740

PR #31, opened 2024-03-20
Add ability to wrap test execution + check for quarantining

gnalh:
In order to be able to quarantine tests we need to wrap the test execution and inspect the results of the test. If a test fails and that test is quarantined then we will override the exit code, otherwise we will return the original exit code. 

Example:
```
cargo run test --junit-paths=py/*.xml --org-url-slug trunk-staging-org --token=[REDACTED] --dry-run -- pytest py/test.py --junitxml=py/test.xml
```


.github/workflows/pull_request.yml
```diff
@@ -45,7 +45,7 @@ jobs:
 
       - name: Upload results using action from ${{ matrix.target }}
         env:
-          TRUNK_API_ADDRESS: https://api.trunk-staging.io/v1/metrics/createBundleUpload
+          TRUNK_API_ADDRESS: https://api.trunk-staging.io
```
gnalh, line 48: We'll need to update our own repo to match once this is live.

src/constants.rs
```diff
@@ -1,5 +1,7 @@
 pub const ALLOW_LIST: &[&'static str] = &[r".*\.xml$"];
 
+pub const EXIT_SUCCESS: i32 = 0;
+pub const EXIT_FAILURE: i32 = 1;
```
gnalh, line 4: Ran into a bunch of typing issues using std::ExitCode directly. It's typed as `u8`, but `ExitStatus` (which is returned by running the Command) is `i32`. I opted to define these since it reduces any concern of precision loss when converting from one form to another.

src/main.rs
```diff
@@ -24,6 +27,13 @@ struct Cli {
 enum Commands {
     #[clap(name = "upload")]
     Upload {
+        #[arg(
+            long,
+            required = false,
+            value_delimiter = ' ',
+            help = "Test command to invoke."
+        )]
+        command: Vec<String>,
```
gnalh, line 36: We can either have this as a specific arg or as a catch all for all parameters after the upload command. I leaned more towards a specific `command` arg since it is more clear what the intent is. The other option is to have a separate `test` command that mimics the `upload` command but also takes in a command. Thoughts?

gnalh: Opted to just have a separate `test` command.

src/main.rs
```diff
@@ -20,50 +23,66 @@ struct Cli {
...
-        #[arg(long, help = "Print files which will be uploaded to stdout.")]
-        print_files: bool,
-        #[arg(long, help = "Run metrics CLI without uploading to API.")]
-        dry_run: bool,
-    },
+    Upload(UploadArgs),
+    Test(TestArgs),
 }
 
-const DEFAULT_API_ADDRESS: &str = "https://api.trunk.io/v1/metrics/createBundleUpload";
```
zaycev, line 66: I think this should be safe to rollout. I don't think any of our customers are setting this externally.

src/main.rs
```diff
@@ -74,15 +93,17 @@ const RETRY_COUNT: usize = 5;
 async fn main() -> anyhow::Result<()> {
     setup_logger()?;
     let cli = Cli::parse();
-    if let Err(e) = run(cli).await {
-        log::error!("Error: {:?}", e);
-        std::process::exit(exitcode::SOFTWARE);
+    match run(cli).await {
```
zaycev, line 96: Maybe I'm missing something but if we already have result which can be either success or failure, why do we also need to return exit code? What would `Ok(EXIT_FAILURE)` mean? Does it mean we successfully captured failure of a test run and we want to propagate it to the caller of cli?

src/main.rs
```diff
@@ -212,7 +235,85 @@ async fn run(cli: Cli) -> anyhow::Result<()> {
...
+        ref token,
+        ref repo_root,
+        ref repo_url,
+        ref repo_head_sha,
+        ref repo_head_branch,
+        ref repo_head_commit_epoch,
+        tags: _,
+        print_files: _,
+        dry_run: _,
+    } = upload_args;
```
zaycev, line 258: If you wan tot avoid adding `ref` you can also take `upload_args` by reference, e.g. `} = &upload_args`. I don't think it matters though.

src/main.rs
```diff
@@ -212,7 +235,85 @@ async fn run(cli: Cli) -> anyhow::Result<()> {
...
+        repo_head_branch.clone(),
+        repo_head_commit_epoch.clone(),
+    )?;
+
+    if junit_paths.is_empty() {
+        return Err(anyhow::anyhow!("No junit paths provided."));
+    }
+
+    let api_address = from_non_empty_or_default(
+        std::env::var("TRUNK_API_ADDRESS").ok(),
```
zaycev, line 273: Since we started using `TRUNK_API_ADDRESS` in more than one place, consider moving it to constants.

src/constants.rs
zaycev: I think this is fine. You can also alias it for better clarity. e.g. `pub type TrunkExitCode = i32`.

src/runner.rs
```diff
@@ -0,0 +1,63 @@
...
+                let metadata = metadata(&file.original_path)?;
+                let time = metadata.modified()?;
+                // skip files that were last modified before the test started
+                if time <= start {
+                    continue;
+                }
+                let file = std::fs::File::open(&file.original_path)?;
+                let reader = std::io::BufReader::new(file);
+                let junitxml = junit_parser::from_reader(reader)?;
+                for suite in junitxml.suites {
```
zaycev, line 44: Feel free to do it in the next PR, parsing XML is totally something we should test before rolling this out.

src/runner.rs
```diff
@@ -0,0 +1,63 @@
...
+use std::process::Command;
+use std::process::Stdio;
+use std::time::SystemTime;
+
+pub async fn run_test_command(
+    repo: &BundleRepo,
+    command: &String,
+    args: Vec<&String>,
+    output_paths: Vec<&String>,
+) -> anyhow::Result<RunResult> {
```
zaycev, line 15: If this function fails to process output of the command, what will happen? Do we ever want to override exit code of the child command?

src/main.rs
gnalh: We need to pipe the original exit code up so that we surface the expected results to the caller. We don't want to hide it if the results weren't quarantined.

zaycev:
Overall – LGTM.

src/runner.rs
gnalh: We should only override the exit code if the only failures parsed were quarantined.

src/main.rs
```diff
@@ -20,69 +23,87 @@ struct Cli {
...
 }
 
-const DEFAULT_API_ADDRESS: &str = "https://api.trunk.io/v1/metrics/createBundleUpload";
+const DEFAULT_ORIGIN: &str = "https://api.trunk.io";
 // Tokio-retry uses base ^ retry * factor formula.
 // This will give us 8ms, 64ms, 512ms, 4096ms, 32768ms
 const RETRY_BASE_MS: u64 = 8;
 const RETRY_FACTOR: u64 = 1;
-const RETRY_COUNT: usize = 5;
+const RETRY_COUNT: usize = 3;
```
gnalh, line 90: 5 was causing very long delays when there were issues with the api.

src/main.rs
gnalh: Perfect, thanks!

src/runner.rs
gnalh: Running into issues getting tests running on my machine. Will address in a follow-up.

src/clients.rs
```diff
@@ -43,6 +47,42 @@ pub async fn get_bundle_upload_location(
...
+pub async fn get_quarantine_bulk_test_status(
+    origin: &str,
+    api_token: &str,
+    org_slug: &str,
+    repo: &Repo,
+    test_identifiers: &[Test],
+) -> anyhow::Result<QuarantineBulkTestStatus> {
+    let client = reqwest::Client::new();
+    let resp = match client
+        .post(format!("{}/v1/metrics/getQuarantineBulkTestStatus", origin,))
```
riya-n, line 59:
```suggestion
        .post(format!("{}/v1/metrics/getQuarantineBulkTestStatus", origin))
```

src/clients.rs
```diff
@@ -43,6 +47,42 @@ pub async fn get_bundle_upload_location(
...
+        .header(reqwest::header::CONTENT_TYPE, "application/json")
```
riya-n, line 60: is there intentionally no timeout?

.github/workflows/pull_request.yml
riya-n: should we still call this TRUNK_API_ADDRESS? maybe TRUNK_API_ORIGIN would be more appropriate?

src/main.rs
riya-n: I've brought this up with Vlad before and he suggested keeping it at 5 because this gives the backend enough time to auto-resolve issues, expensive queries to finish, etc. Not sure if this is still a concern though cc @zaycev

src/main.rs
```diff
@@ -212,7 +235,90 @@ async fn run(cli: Cli) -> anyhow::Result<()> {
...
+            &run_result.failures,
+        )
+    })
+    .await
+    .unwrap_or(QuarantineBulkTestStatus {
+        group_is_quarantined: false,
+    });
+    // use the exit code from the command if the group is not quarantined
+    // override exit code to be exit_success if the group is quarantined
+    let exit_code = if !run_result.exit_code != 0 && !quarantine_results.group_is_quarantined {
```
riya-n, line 302:
```suggestion
    let exit_code = if !run_result.exit_code != EXIT_SUCCESS && !quarantine_results.group_is_quarantined {
```

src/clients.rs
gnalh: Good catch, we should add trunk check to this repo so it's caught there.

src/clients.rs
gnalh: I was looking at an example without one. Seems like timeouts are inconsistent here. I'll add one.

src/main.rs
gnalh: The delay is very long with 5 (10+ seconds). I think we need a lower number.

.github/workflows/pull_request.yml
gnalh: `API_ADDRESS` is a pattern we use elsewhere.

gnalh:
This is ready for a re-review.

src/main.rs
riya-n: Yeah, I'd mentioned that to Vlad as well. Will let him comment on this

riya-n (approved):

src/main.rs
zaycev: > 5 was causing very long delays when there were issues with the api.

That was the original point for adding this delay - wait enough for transient issue to resolve. What are the down sites of it?

gnalh: It feels broken when nothing happens for 20 seconds.

zaycev (approved):

zaycev:
LGTM

PR #35, opened 2024-03-22
Update versioning

gnalh:
The version defined in Cargo.toml was incorrect. Update it to match what we expect the new version to be.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "trunk-analytics-cli"
 edition = "2021"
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "trunk-analytics-cli"
-edition = "2021"
-version = "0.4.0"
+edition = "2024"
+version = "0.0.5"
```
gnalh, line 4: Next release should be 0.0.5

PR #36, opened 2024-03-25
Update build version

gnalh:


Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "trunk-analytics-cli"
 edition = "2021"
-version = "0.4.0"
+version = "0.4.1"
```
gnalh, line 4:
```suggestion
version = "0.5.0"
```

riya-n (approved):

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "trunk-analytics-cli"
 edition = "2021"
-version = "0.4.0"
+version = "0.5.0"
```
gnalh, line 4: Increase minor version since we are adding new functionality in a backwards compatible way.

pat-trunk-io: Can cargo.toml be generated from git tag?

gnalh: That's a great question, I'm not sure.

PR #37, opened 2024-03-25
Update release runner type

gnalh:
Our self hosted runners don't have `gh` installed. Not sure how the first release went through.


riya-n (approved):

dapirian:
did we have a ci step to setup trunk tools? because we install `gh` and add it to PATH via trunk tools, possibly for CI too

gnalh:
> did we have a ci step to setup trunk tools? because we install `gh` and add it to PATH via trunk tools, possibly for CI too

Addressed [here](https://github.com/trunk-io/analytics-cli/pull/38)

PR #44, opened 2024-03-27
Improve error handling

gnalh:
We should always either return exit_success or the underlying exit code. We should not obfuscate it when an unhandled exception occurs.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -6,10 +6,10 @@ use tokio_retry::strategy::ExponentialBackoff;
 use tokio_retry::Retry;
 use trunk_analytics_cli::bundler::BundlerUtil;
 use trunk_analytics_cli::clients::get_quarantine_bulk_test_status;
-use trunk_analytics_cli::constants::{EXIT_SUCCESS, TRUNK_API_ADDRESS_ENV};
+use trunk_analytics_cli::constants::{EXIT_FAILURE, EXIT_SUCCESS, TRUNK_API_ADDRESS_ENV};
 use trunk_analytics_cli::runner::run_test_command;
 use trunk_analytics_cli::scanner::{BundleRepo, EnvScanner, FileSet, FileSetCounter};
-use trunk_analytics_cli::types::{BundleMeta, QuarantineBulkTestStatus, META_VERSION};
+use trunk_analytics_cli::types::{BundleMeta, QuarantineBulkTestStatus, RunResult, META_VERSION};
 use trunk_analytics_cli::utils::{from_non_empty_or_default, parse_custom_tags};
 
 #[derive(Debug, Parser)]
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -283,20 +283,31 @@ async fn run_test(test_args: TestArgs) -> anyhow::Result<i32> {
...
-    })
     .await
-    .unwrap_or(QuarantineBulkTestStatus {
-        group_is_quarantined: false,
+    .unwrap_or(RunResult {
+        exit_code: EXIT_FAILURE,
+        failures: Vec::new(),
     });
+
+    let quarantine_results = match run_result.failures.is_empty() {
```
zaycev, line 292: Match seems redundant.
```rust
let quarantine_results =  if  run_result.failures.is_empty()
{
  ...
} else 
{
 ...
}
```

src/main.rs
```diff
@@ -310,13 +321,11 @@ async fn run_test(test_args: TestArgs) -> anyhow::Result<i32> {
         run_result.exit_code
     };
 
-    let upload_exit_code = run_upload(upload_args).await.unwrap_or(EXIT_SUCCESS);
-    // use the upload exit code if the command exit code is exit_success
-    if exit_code == EXIT_SUCCESS {
-        Ok(upload_exit_code)
-    } else {
-        Ok(exit_code)
+    let upload_exit_code = run_upload(upload_args).await.unwrap_or(EXIT_FAILURE);
```
zaycev, line 324: `unwrap` will lose context of an error which we want to log. I would use match for each case instead. Something like this

```rust
match run_upload(...)
{
    Ok(EXIT_SUCCESS) => EXIT_SUCCESS,
    Ok(code)         => { log:error("...");      EXIT_FAILURE },
    Err(err).        => { log:error("...", err); EXIT_FAILURE },
}
```

zaycev: Or if we don't need exit code:

```rust
match run_upload(...)
{
    Ok(EXIT_SUCCESS) => (),
    Ok(code)         => { log:error("...", code); },
    Err(err).        => { log:error("...", err);  },
}
```

src/runner.rs
```diff
@@ -20,32 +20,75 @@ pub async fn run_test_command(
         .stderr(Stdio::inherit())
         .spawn()
         .unwrap();
-    let result = child.wait()?;
```
zaycev, line 23: nit. This is another shortcut syntax you can use (up to you):

```rust
let Ok(result) = child.wait() else { 
  log::error!("Error waiting for execution: {}", e); 
  return Ok(RunResult{ ... }); 
}
```

src/runner.rs
```diff
@@ -20,32 +20,75 @@ pub async fn run_test_command(
...
-        let file_sets = output_paths
+        let file_sets = match output_paths
             .iter()
             .map(|path| {
                 FileSet::scan_from_glob(&repo.repo_root, path.to_string(), &mut file_counter)
             })
-            .collect::<anyhow::Result<Vec<FileSet>>>()?;
+            .collect::<anyhow::Result<Vec<FileSet>>>()
+        {
+            Ok(file_sets) => file_sets,
```
zaycev, line 43: Same here. `let Ok(file_sets) = ...` might be more compact.

src/runner.rs
```diff
@@ -20,32 +20,75 @@ pub async fn run_test_command(
...
                 if time <= start {
-                    log::debug!(
+                    log::info!(
                         "Skipping file because of lack of modification: {}",
                         file.original_path
                     );
                     continue;
                 }
-                let file = std::fs::File::open(&file.original_path)?;
+                let file = match std::fs::File::open(&file.original_path) {
```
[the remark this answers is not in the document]
zaycev: Actually, nvm. This is not going to work since you need error object.

src/runner.rs
zaycev: nvm. This is not going to work since you need error val.

src/runner.rs
zaycev: nvm. This is not going to work since you need error val.

PR #48, opened 2024-04-08
find remote reference when HEAD detached

gnalh:
Scan the references to find the matching reference name. This is equivalent to `git for-each-ref --contains e0c6eaa19870c1ae30601bed5bee604c7436c9a2`

After:
```
2024-04-08T05:37:42 [INFO] - Found git_sha: Some("e0c6eaa19870c1ae30601bed5bee604c7436c9a2")
2024-04-08T05:37:42 [INFO] - Found git_branch: Some("refs/remotes/pull/47/merge")
```

Before
```
2024-04-08T05:39:17 [INFO] - Found git_sha: Some("e0c6eaa19870c1ae30601bed5bee604c7436c9a2")
2024-04-08T05:39:17 [INFO] - Found git_branch: None
```

Repro:
```
git -c protocol.version=2 fetch --no-tags --prune --no-recurse-submodules --depth=1 origin +e0c6eaa19870c1ae30601bed5bee604c7436c9a2:refs/remotes/pull/47/merge
git checkout --progress --force refs/remotes/pull/47/merge
```

### DIFF
--- a/src/scanner.rs
+++ b/src/scanner.rs
@@ -174,7 +174,8 @@ impl BundleRepo {
                                 && git_head.id().is_some()
                                 && id.unwrap().to_string() == git_head.id().unwrap().to_string()
                             {
-                                git_head_branch = r.name().to_path().to_str().map(|s| s.to_string());
+                                git_head_branch =
+                                    r.name().to_path().to_str().map(|s| s.to_string());
                                 break;
                             };
                         }
--- a/src/scanner.rs
+++ b/src/scanner.rs
@@ -1,4 +1,5 @@
 use regex::Regex;
+use std::format;
 use serde::Serialize;
 
 use crate::constants::{ALLOW_LIST, ENVS_TO_GET};
@@ -161,8 +162,25 @@ impl BundleRepo {
                 .string_by_key(GIT_REMOTE_ORIGIN_URL_CONFIG)
                 .map(|s| s.to_string());
             let mut git_head = git_repo.head()?;
+
+            let reference = git_repo.references().unwrap();
+            let git_head_id = git_head.id().unwrap();
+            let mut git_head_branch = git_head.referent_name().map(|s| s.as_bstr().to_string());
+            reference.remote_branches()?.for_each(|r| {
+                match r {
+                    Ok(r) => {
+                        let target = r.target();
+                        let id = target.try_id();
+                        if id.is_some() && id.unwrap().to_string() == git_head_id.to_string() {
+                            git_head_branch = Some(r.name().to_path().to_str().unwrap().to_string());
+                        };
+                    }
+                    Err(e) => {
+                        log::info!("reference {:?}", e);
+                    }
+                }
+            });
             let git_head_sha = git_head.id().map(|id| id.to_string());
-            let git_head_branch = git_head.referent_name().map(|s| s.as_bstr().to_string());
             let git_head_commit_time = git_head.peel_to_commit_in_place()?.time()?;
             git_head_commit_message = git_head.peel_to_commit_in_place().map_or(None, |commit| {
                 commit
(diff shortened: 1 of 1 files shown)


src/scanner.rs
```diff
@@ -161,8 +162,25 @@ impl BundleRepo {
...
+            let git_head_id = git_head.id().unwrap();
+            let mut git_head_branch = git_head.referent_name().map(|s| s.as_bstr().to_string());
+            reference.remote_branches()?.for_each(|r| {
+                match r {
+                    Ok(r) => {
+                        let target = r.target();
+                        let id = target.try_id();
+                        if id.is_some() && id.unwrap().to_string() == git_head_id.to_string() {
+                            git_head_branch = Some(r.name().to_path().to_str().unwrap().to_string());
+                        };
```
gnalh, line 176: I'm sure there's a more elegant solution here and could use some rust pointers.

src/scanner.rs
```diff
@@ -161,8 +162,26 @@ impl BundleRepo {
...
+                let git_head_id = git_head.id().unwrap();
+                let reference = git_repo.references().unwrap();
+                reference.remote_branches()?.for_each(|r| match r {
+                    Ok(r) => {
+                        let target = r.target();
+                        let id = target.try_id();
+                        if id.is_some() && id.unwrap().to_string() == git_head_id.to_string() {
+                            git_head_branch =
+                                Some(r.name().to_path().to_str().unwrap().to_string());
+                        };
```
gnalh, line 177: I'm sure there's a more elegant solution here. If there is, any pointers would be greatly appreciated.

src/scanner.rs
```diff
@@ -161,8 +162,30 @@ impl BundleRepo {
...
+                    match r {
+                        Ok(r) => {
+                            let target = r.target();
+                            let id = target.try_id();
+                            if id.is_some()
+                                && git_head.id().is_some()
+                                && id.unwrap().to_string() == git_head.id().unwrap().to_string()
+                            {
+                                git_head_branch =
+                                    Some(r.name().to_path().to_str().unwrap().to_string());
```
riya-n, line 178:
```suggestion
                                    r.name().to_path().to_str().map(|s| s.to_string());
```
which I believe would either give you `None` or `Some(s.to_string())`

src/scanner.rs
```diff
@@ -161,8 +162,30 @@ impl BundleRepo {
...
+                                break;
+                            };
+                        }
+                        Err(e) => {
+                            log::warn!("unable to match reference {:?}", e);
```
riya-n, line 183: Why do we need to warn?

riya-n (approved):
Looks good, but not sure we need to match or warn. Just a suggestion, but we could do something like this as well
```
if git_head_branch.is_none() {
  for rb in git_repo.references()?.remote_branches()? {
    git_head_branch = rb.map_or(None, |r| {
      let id = r.target().try_id().map(|id| id.to_string());
      let git_head_id = git_head.id().map(|id| id.to_string());
      if id.is_some() && git_head_id.is_some() && id == git_head_id {
        return r.name().to_path().to_str().map(|s| s.to_string());
      }
      None
    });
  }
}
```

src/scanner.rs
gnalh: We can lower it to info. I think right now we want to be verbose on logging to help with debugging.

src/scanner.rs
gnalh: Nice, that's handy.

src/scanner.rs
```diff
@@ -161,8 +162,30 @@ impl BundleRepo {
...
+                                && git_head.id().is_some()
+                                && id.unwrap().to_string() == git_head.id().unwrap().to_string()
+                            {
+                                git_head_branch =
+                                    r.name().to_path().to_str().map(|s| s.to_string());
+                                break;
+                            };
+                        }
+                        Err(e) => {
+                            log::info!("Unexpected error when trying to find reference {:?}", e);
```
gnalh, line 183:
```suggestion
                            log::debug!("Unexpected error when trying to find reference {:?}", e);
```

PR #50, opened 2024-04-09
send classname + file when checking quarantine status

gnalh:


riya-n (approved):

src/runner.rs
```diff
@@ -103,6 +103,8 @@ pub async fn run_test_command(
                             failures.push(Test {
                                 parent_name: parent_name.clone(),
                                 name: name.clone(),
+                                classname: case.classname.clone(),
```
riya-n, line 106: nit, optional: `class_name`?

PR #53, opened 2024-04-16
Update CLI to parse code owners

riya-n:
https://linear.app/trunk/issue/TRUNK-10994/update-cli-to-parse-code-owners

tested here: https://github.com/trunk-io/analytics-cli/actions/runs/8700187796/job/23860008954

### DIFF
--- a/src/scanner.rs
+++ b/src/scanner.rs
@@ -59,10 +59,8 @@ impl FileSet {
         };
 
         // Parse codeowners.
-        // TODO: check correct path to locate in.
         let owners_of_paths =
             codeowners::locate(".").map(|path| codeowners::from_path(path.as_path()));
-        log::info!("Owners of paths: {:?} and {}", owners_of_paths, path_to_scan);
 
         let mut files = Vec::new();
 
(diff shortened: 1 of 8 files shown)


src/scanner.rs
```diff
@@ -58,6 +58,9 @@ impl FileSet {
             glob_path.clone()
         };
 
+        // Parse codeowners.
+        let codeowners = codeowners::locate(".").map(|path| codeowners::from_path(path.as_path()));
```
riya-n, line 62: It might be better to have a list of possible directories for the `CODEOWNERS` file (e.g., `.github/`) instead of searching through everything - not great for big repos

riya-n: For GitHub CODEOWNERS file location: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-file-location
Bitbucket: https://confluence.atlassian.com/bitbucketserver/code-owners-1296171116.html
GitLab: https://docs.gitlab.com/ee/user/project/codeowners/#codeowners-file

Given the docs above, we could first try searching for the `CODEOWNERS` file in a list of directories like `.github/`, `.bitbucket/`, `.gitlab/`, `docs/` and only searching in `"."` if we haven't found a file as yet.

gnalh: Let's look in those 4 directories and have a manual override if you are doing something else. I want to avoid scanning the file system as much as possible.

gnalh (approved):

PR #59, opened 2024-06-03
Add createRepo public endpoint

riya-n:
- this should create a trunk repo for orgs that don't already have them
- calls the createRepo public endpoint: https://github.com/trunk-io/trunk/pull/14243
- tested upload on a test repo and seems to work fine
- TODO: wait for endpoint to be on prod, then merge this and update cli version on uploader

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -272,7 +272,13 @@ async fn run_upload(
 
     let remote_urls = vec![repo.repo_url.clone()];
     Retry::spawn(default_delay(), || {
-        create_trunk_repo(&api_address, &token, &org_url_slug, &repo.repo, &remote_urls)
+        create_trunk_repo(
+            &api_address,
+            &token,
+            &org_url_slug,
+            &repo.repo,
+            &remote_urls,
+        )
     })
     .await?;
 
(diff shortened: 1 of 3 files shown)


gnalh:
Is this endpoint available on prod yet?

gnalh (approved):

src/main.rs
```diff
@@ -268,7 +266,19 @@ async fn run_upload(
     }
 
     Retry::spawn(default_delay(), || {
-        trunk_analytics_cli::clients::put_bundle_to_s3(&upload.url, &bundle_time_file)
+        put_bundle_to_s3(&upload.url, &bundle_time_file)
+    })
+    .await?;
+
+    let remote_urls = vec![repo.repo_url.clone()];
```
zaycev, line 273: Is this a correct order? Should we create repo first and then create upload?

src/types.rs
```diff
@@ -56,6 +65,12 @@ pub struct QuarantineBulkTestStatus {
     pub quarantine_results: Vec<QuarantineResult>,
 }
 
+#[derive(Debug, Serialize, Clone, Deserialize)]
+pub struct TrunkRepo {
```
zaycev, line 69: If we aren't using response object, maybe we can have empty schema for now.

src/main.rs
gnalh: The order shouldn't matter since the other endpoints don't depend on a repo existing.

src/types.rs
riya-n: Yeah, we aren't using it, but I think it's fine to leave this as a sanity check that the repoId exists in the response

PR #67, opened 2024-07-08
Track cli version

riya-n:
Ticket: https://linear.app/trunk/issue/TRUNK-11730/track-analytics-cli-version-in-upload
- this small change will make it easier to just track the cli version, without having to parse it in the etl

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -18,7 +18,7 @@ use trunk_analytics_cli::utils::{from_non_empty_or_default, parse_custom_tags};
 
 #[derive(Debug, Parser)]
 #[command(
-    version = std::env!("CARGO_PKG_VERSION"),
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -206,12 +206,7 @@ async fn run_upload(
     });
     let meta = BundleMeta {
         version: META_VERSION.to_string(),
-        cli_version: format!(
```
zaycev, line 209: should we just split this into 3 fields?

riya-n: Not sure why we had this in the first place, we don't use any of the other fields in the ETL it seems (we weren't even using `cli_version`). I think we log all 3 fields when running the CLI so might've just copied over that string

riya-n:
We will instead extract the `cli_version` from the existing format in the ETL

PR #68, opened 2024-07-09
Don't fail on failed uploads

riya-n:
Ticket: https://linear.app/trunk/issue/TRUNK-11948/never-fail-customers-ci-pipeline-on-failed-uploads-or-analytics

### DIFF
--- a/src/clients.rs
+++ b/src/clients.rs
@@ -37,9 +37,7 @@ pub async fn create_trunk_repo(
         Err(e) => return Err(anyhow::anyhow!(e).context("Failed to validate trunk repo")),
     };
 
-    if resp.status() != reqwest::StatusCode::OK
-        && resp.text().await?.contains("Organization not found")
-    {
+    if resp.status() == reqwest::StatusCode::NOT_FOUND {
         return Err(anyhow::anyhow!(
             "Organization not found. Please double check the provided organization url slug: {}",
             org_slug
@@ -73,16 +71,15 @@ pub async fn get_bundle_upload_location(
         Err(e) => return Err(anyhow::anyhow!(e).context("Failed to create bundle upload")),
     };
 
-    if resp.status() != reqwest::StatusCode::OK {
-        if resp.text().await?.contains("Organization not found") {
-            return Err(
-                anyhow::anyhow!("Organization not found. Please double check the provided organization url slug: {}", org_slug)
-                    .context("Failed to create bundle upload"),
-            );
-        } else {
-            log::info!("Failed to create bundle upload. We will try again at a later time.");
-            return Ok(None);
-        }
+    if resp.status() == reqwest::StatusCode::NOT_FOUND {
+        return Err(anyhow::anyhow!(
+            "Organization not found. Please double check the provided organization url slug: {}",
+            org_slug
+        )
+        .context("Failed to create bundle upload"));
+    } else if resp.status() != reqwest::StatusCode::OK {
+        log::info!("Failed to create bundle upload. We will try again at a later time.");
+        return Ok(None);
     }
 
     resp.json::<Option<BundleUploadLocation>>()
--- a/src/clients.rs
+++ b/src/clients.rs
@@ -39,7 +39,7 @@ pub async fn create_trunk_repo(
 
     if resp.status() == reqwest::StatusCode::NOT_FOUND {
         return Err(anyhow::anyhow!(
-            "Organization not found. Please double check the provided organization url slug: {}",
+            "Organization not found. Please double check the provided organization token and url slug: {}",
             org_slug
         )
         .context("Failed to validate trunk repo"));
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -95,7 +95,7 @@ const DEFAULT_ORIGIN: &str = "https://api.trunk.io";
 // This will give us 8ms, 64ms, 512ms, 4096ms, 32768ms
 const RETRY_BASE_MS: u64 = 8;
 const RETRY_FACTOR: u64 = 1;
-const RETRY_COUNT: usize = 3;
+const RETRY_COUNT: usize = 5;
```
riya-n, line 98: previous discussion on retry_count: https://github.com/trunk-io/analytics-cli/pull/31#discussion_r1532611386

src/clients.rs
```diff
@@ -34,14 +34,17 @@ pub async fn create_trunk_repo(
...
+        Err(e) => return Err(anyhow::anyhow!(e).context("Failed to validate trunk repo")),
     };
 
-    if resp.status() != reqwest::StatusCode::OK {
-        return Err(
-            anyhow::anyhow!("{}: {}", resp.status(), status_code_help(resp.status()))
-                .context("Failed to create trunk repo"),
-        );
+    if resp.status() != reqwest::StatusCode::OK
+        && resp.text().await?.contains("Organization not found")
```
zaycev, line 41: Should we check with status `404` instead of text?

src/clients.rs
```diff
@@ -71,15 +74,20 @@ pub async fn get_bundle_upload_location(
...
-            anyhow::anyhow!("{}: {}", resp.status(), status_code_help(resp.status()))
-                .context("Failed to create bundle upload"),
-        );
+        if resp.text().await?.contains("Organization not found") {
+            return Err(
+                anyhow::anyhow!("Organization not found. Please double check the provided organization url slug: {}", org_slug)
+                    .context("Failed to create bundle upload"),
+            );
+        } else {
+            log::info!("Failed to create bundle upload. We will try again at a later time.");
```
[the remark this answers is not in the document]
zaycev: This message isn've very descriptive. How would user know what didn't work?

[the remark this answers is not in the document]
riya-n: If we are hitting this case, it is probably an error on our end.

src/clients.rs
riya-n: Yeah, it seems like the only case where we return 404 from the public_api endpoints is if org isn't found from orgslug

src/clients.rs
```diff
@@ -71,15 +74,20 @@ pub async fn get_bundle_upload_location(
     };
 
     if resp.status() != reqwest::StatusCode::OK {
-        return Err(
-            anyhow::anyhow!("{}: {}", resp.status(), status_code_help(resp.status()))
-                .context("Failed to create bundle upload"),
-        );
+        if resp.text().await?.contains("Organization not found") {
+            return Err(
+                anyhow::anyhow!("Organization not found. Please double check the provided organization url slug: {}", org_slug)
```
zaycev, line 79: Slug and token.

src/clients.rs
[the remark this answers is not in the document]
zaycev: Right, but how do we know. If cli receives unexpected response, I don't see a reason to hide it.

src/clients.rs
zaycev: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status

(1) If status is 5xx the issue is on our side,  we should explain what happened, recommend actions and return 0.
(2) If status is 4xx the issue is on client side, we should explain what happened, recommend actions and return 1.
(3) Otherwise, if the response is unexpected handle it as (1).

src/clients.rs
```diff
@@ -70,16 +71,24 @@ pub async fn get_bundle_upload_location(
         Err(e) => return Err(anyhow::anyhow!(e).context("Failed to create bundle upload")),
     };
 
-    if resp.status() != reqwest::StatusCode::OK {
-        return Err(
-            anyhow::anyhow!("{}: {}", resp.status(), status_code_help(resp.status()))
-                .context("Failed to create bundle upload"),
+    if resp.status() == reqwest::StatusCode::NOT_FOUND {
```
zaycev, line 74: `resp.status().is_client_error()`

zaycev: or:
```rust
if resp.status().is_success()
    {
        resp.json::<Option<BundleUploadLocation>>()
            .await
            .context("Failed to get response body as json")
    }
    
    if resp.status().is_client_error()
    {
        return Err(anyhow::anyhow!(
            "Organization not found. Please double check the provided organization token and url slug: {}",
            org_slug
        )
        .context("Failed to create bundle upload"));
    }
    
    log::warn!(
        "Failed to create bundle upload. {}: {}",
        resp.status(),
        status_code_help(resp.status())
    );
    return Ok(None);
```

zaycev (approved):

src/clients.rs
riya-n: oo that is handy, thanks

PR #81, opened 2024-08-13
Search for CODEOWNERS in repo root

dfrankland:
https://github.com/softprops/codeowners/blob/e1bf125a2076fbacab73af310daac9a344aa95e5/src/lib.rs#L163

### DIFF
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -3,7 +3,7 @@ pub const ALLOW_LIST: &[&str] = &[r".*\.xml$", r".*\.junit$"];
 pub const EXIT_SUCCESS: i32 = 0;
 pub const EXIT_FAILURE: i32 = 1;
 
-pub const CODEOWNERS_LOCATIONS: &[&str] = &[".github/", ".bitbucket/", ".gitlab/", "docs/"];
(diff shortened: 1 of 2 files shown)


mmatheson (approved):

src/constants.rs
```diff
@@ -3,7 +3,20 @@ pub const ALLOW_LIST: &[&str] = &[r".*\.xml$", r".*\.junit$"];
...
+//      .github
+//      root
+//      docs
+// GitLab: https://docs.gitlab.com/ee/user/project/codeowners/#codeowners-file
+//      root
+//      docs
+//      .gitlab
+// BitBucket: https://support.atlassian.com/bitbucket-cloud/docs/set-up-and-use-code-owners/#Set-up-your-CODEOWNERS-file
+//      .bitbucket
+pub const CODEOWNERS_LOCATIONS: &[&str] = &[".github", ".bitbucket", ".", "docs", ".gitlab"];
```
dfrankland, line 19: We could add logic for checking the provider and looking only for those directories, but to me that seems a bit over-engineered. Let me know if y'all think otherwise.

mmatheson (approved):

src/constants.rs
federicoweber: agree I don't believe that will be needed.

federicoweber (approved):

PR #82, opened 2024-08-14
Add sentry to binary, create a sentry release in release pipeline

mmatheson:


pat-trunk-io (approved):
Lgtm

For c++ we also need to upload symbols for stack trace. Is that required for rust?

src/main.rs
```diff
@@ -102,8 +102,9 @@ const RETRY_BASE_MS: u64 = 8;
 const RETRY_FACTOR: u64 = 1;
 const RETRY_COUNT: usize = 5;
 
-#[tokio::main]
-async fn main() -> anyhow::Result<()> {
+// "the Sentry client must be initialized before starting an async runtime or spawning threads"
+// https://docs.sentry.io/platforms/rust/#async-main-function
+fn main() {
```
dfrankland, line 107: Optional: this can still return a `anyhow::Result<()>` to leverage the try sigil (`?`)

src/main.rs
```diff
@@ -112,15 +113,21 @@ async fn main() -> anyhow::Result<()> {
...
-            log::error!("Error: {:?}", e);
-            std::process::exit(exitcode::SOFTWARE);
-        }
-    }
+    tokio::runtime::Builder::new_multi_thread()
+        .enable_all()
+        .build()
+        .unwrap()
+        .block_on(async {
+            setup_logger().expect("Setup Logger");
```
dfrankland, line 121: Nit: this would be a confusing error to see for a user. I think we can make it slightly better with a message like "unable to initialize". Also, do we want to hard fail on the logger not being able to be setup?

dfrankland (approved):

src/main.rs
mmatheson: We do not, your above comment helped me fix this one too :)

src/main.rs
```diff
@@ -102,17 +102,32 @@ const RETRY_BASE_MS: u64 = 8;
...
-        Err(e) => {
-            log::error!("Error: {:?}", e);
-            std::process::exit(exitcode::SOFTWARE);
-        }
-    }
+// "the Sentry client must be initialized before starting an async runtime or spawning threads"
+// https://docs.sentry.io/platforms/rust/#async-main-function
+fn main() -> anyhow::Result<()> {
+    let _guard = sentry::init((
+        "https://4814eaf1df0e8a1e3303bb7e2f89095a@o681886.ingest.us.sentry.io/4507772986982400",
```
zaycev, line 109: nit. I would move this to constants.rs

src/main.rs
```diff
@@ -102,17 +102,32 @@ const RETRY_BASE_MS: u64 = 8;
...
+        sentry::ClientOptions {
+            release: sentry::release_name!(),
+            ..Default::default()
+        },
+    ));
+
+    tokio::runtime::Builder::new_multi_thread()
+        .enable_all()
+        .build()
+        .unwrap()
```
zaycev, line 119: `.unwrap()` -> `.expect("Failed to build tokio runtime")`

zaycev:
> Lgtm
> 
> For c++ we also need to upload symbols for stack trace. Is that required for rust?

@mmatheson I think you may also need to specify `panic = 'unwind'` in Cargo for the new profile.

PR #83, opened 2024-08-14
do not hard fail on failure to parse codeowners file

dfrankland:
`codeowners` uses `unwrap`, so we need to fork the library to allow us to handle errors. If we cannot parse a codeowners file a debug log will be emitted:
```
2024-08-14T16:43:26 [DEBUG] - Found CODEOWNERS file `/Users/dylan/GitHub/flake-factory/./CODEOWNERS`, but couldn't parse it.
```


cli/src/scanner.rs
```diff
@@ -73,7 +73,16 @@ impl FileSet {
                 }
             }
         }
-        let codeowners = codeowners_file.map(|path| codeowners::from_path(path.as_path()));
+        let codeowners = codeowners_file.and_then(|path| {
+            if let Ok(owners) = codeowners::from_path(path.as_path()) {
+                return Some(owners);
+            }
+            log::debug!(
+                "Found CODEOWNERS file `{}`, but couldn't parse it.",
```
zaycev, line 81: Should be `error!`?

zaycev:
You may need to update workflows to use new paths. Specifically `.github/workflows/release.yml`, but might be others as well.

cli/src/scanner.rs
zaycev: Also, let's log the parsing error.

dfrankland: I fine with that. Do we want to show this error to the users though? We plan to follow up by uploading the CODEOWNERS file regardless so that we can eventually post-process

dfrankland:
I'll double-check the `release.yml` workflow, but it should work fine. `cargo` will intelligently build and run things

cli/src/scanner.rs
dfrankland: Also, the parsing error isn't useful—it's just a `PatternError` from the `glob` crate. That won't help anyone understand why their file isn't correct.

dfrankland:
I updated the log to look like this:
```
2024-08-15T07:55:02 [ERROR] - Found CODEOWNERS file `/Users/dylan/GitHub/flake-factory/./CODEOWNERS`, but couldn't parse it: Pattern syntax error near position 3: invalid range pattern
```

Also, I updated the code to use `file.is_file()` rather than `file.exists()`... Since macOS fs is case-insensitive, our test run on CI would hang indefinitely since the CLI would find the new `codeowners` directory and try to open it as a file 🤦

PR #85, opened 2024-08-16
[TRUNK-12396] update cli to work with the new apis

federicoweber:
This updates the quarantine flow to work with the new APIs

### DIFF
--- a/cli/src/runner.rs
+++ b/cli/src/runner.rs
@@ -1,9 +1,9 @@
 use crate::{
-    clients::get_quarantine_bulk_test_status,
+    clients::get_quarantining_config,
     codeowners::CodeOwners,
     constants::{EXIT_FAILURE, EXIT_SUCCESS},
     scanner::{BundleRepo, FileSet, FileSetCounter},
-    types::{QuarantineBulkTestStatus, QuarantineRunResult, RunResult, Test},
+    types::{QuarantineBulkTestStatus, QuarantineConfig, QuarantineRunResult, RunResult, Test},
 };
 use junit_parser;
 use std::{
@@ -202,42 +202,44 @@ pub async fn run_quarantine(
     repo: &BundleRepo,
     delay: std::iter::Take<ExponentialBackoff>,
 ) -> anyhow::Result<QuarantineRunResult> {
-    let quarantine_results = if run_result.failures.is_empty() {
-        QuarantineBulkTestStatus {
-            group_is_quarantined: false,
-            quarantine_results: Vec::new(),
-        }
-    } else {
+    let mut quarantine_results = QuarantineBulkTestStatus {
+        group_is_quarantined: false,
+        quarantine_results: Vec::new(),
+    };
+    let quarantine_config: QuarantineConfig = if !run_result.failures.is_empty() {
         log::info!("Quarantining failed tests");
         let result = Retry::spawn(delay, || {
-            get_quarantine_bulk_test_status(api_address, token, org_url_slug, &repo.repo)
+            get_quarantining_config(api_address, token, org_url_slug, &repo.repo)
         })
         .await;
 
         result.unwrap_or_else(|e| {
             log::error!("Failed to get quarantine results: {:?}", e);
-            QuarantineBulkTestStatus {
-                group_is_quarantined: false,
-                quarantine_results: Vec::new(),
+            QuarantineConfig {
+                is_preview_mode: false,
+                quarantined_tests: Vec::new(),
             }
         })
+    } else {
+        log::info!("No failed tests to quarantine");
+        QuarantineConfig {
+            is_preview_mode: false,
+            quarantined_tests: Vec::new(),
+        }
     };
 
-    // Print out failed tests and their quarantine status
-    let quarantined: HashSet<_> = quarantine_results
-        .quarantine_results
-        .iter()
-        .map(|qr| &qr.name)
-        .collect();
-
+    // quarantine the failed tests
+    let quarantined: HashSet<_> = quarantine_config.quarantined_tests.iter().collect();
     for failure in &run_result.failures {
-        if quarantined.contains(&failure.name) {
+        if quarantined.contains(&failure.id.as_ref().unwrap()) {
             log::error!(
                 "{} -> {} [QUARANTINED] (id: {:?})",
                 failure.parent_name,
                 failure.name,
                 failure.id.clone().unwrap()
             );
+
+            quarantine_results.quarantine_results.push(failure.clone());
         } else {
             log::error!(
                 "{} -> {} (id: {:?})",
@@ -247,13 +249,19 @@ pub async fn run_quarantine(
             );
         }
     }
+    quarantine_results.group_is_quarantined = run_result.failures.iter().all(|failure| {
+        quarantine_results
+            .quarantine_results
+            .iter()
+            .any(|quarantined_failure| quarantined_failure.id == failure.id)
+    });
 
     // use the exit code from the command if the group is not quarantined
     // override exit code to be exit_success if the group is quarantined
     let exit_code = if !quarantine_results.group_is_quarantined {
         log::info!("Not all test failures were quarantined, returning exit code from command.");
         run_result.exit_code
-    } else if run_result.exit_code != EXIT_SUCCESS {
+    } else if run_result.exit_code != EXIT_SUCCESS && !quarantine_config.is_preview_mode {
         log::info!("All test failures were quarantined, overriding exit code to be exit_success");
         EXIT_SUCCESS
     } else {
--- a/cli/src/runner.rs
+++ b/cli/src/runner.rs
@@ -16,6 +16,7 @@ use tokio_retry::{strategy::ExponentialBackoff, Retry};
 
 pub async fn run_test_command(
     repo: &BundleRepo,
+    org_slug: &str,
     command: &String,
     args: Vec<&String>,
     output_paths: &[String],
@@ -27,7 +28,7 @@ pub async fn run_test_command(
     let (file_sets, ..) = build_filesets(repo, output_paths, team, codeowners)?;
     let failures = if exit_code != EXIT_SUCCESS {
         let start = SystemTime::now();
-        extract_failed_tests(&file_sets, Some(start)).await?
+        extract_failed_tests(repo, org_slug, &file_sets, Some(start)).await?
     } else {
         Vec::new()
     };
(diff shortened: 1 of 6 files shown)


cli/src/runner.rs
```diff
@@ -179,54 +202,66 @@ pub async fn run_quarantine(
...
-            log::error!("{} -> {} [QUARANTINED]", failure.parent_name, failure.name);
+        if quarantined.contains(&failure.id.as_ref().unwrap()) {
+            log::error!(
+                "{} -> {} [QUARANTINED] (id: {:?})",
+                failure.parent_name,
+                failure.name,
+                failure.id.clone().unwrap()
+            );
+
+            quarantine_results.quarantine_results.push(failure.clone());
```
dfrankland, line 242: We end up just clone things from `RunResult` and nothing else needs access to it, so `run_quarantine` should accept a value like `run_result: RunResult` in the arguments

cli/src/runner.rs
```diff
@@ -179,54 +202,66 @@ pub async fn run_quarantine(
...
+                failure.id.clone().unwrap()
+            );
         }
     }
+    quarantine_results.group_is_quarantined = run_result.failures.iter().all(|failure| {
+        quarantine_results
+            .quarantine_results
+            .iter()
+            .any(|quarantined_failure| quarantined_failure.id == failure.id)
+    });
```
dfrankland, line 257: I think this logic is hard to follow and needlessly clones a lot of things. From what I can see this boils down very simply to the following:
```suggestion
    let total_failures = run_result.failures.len();
    quarantine_results.quarantine_results = run_result
        .failures
        .into_iter()
        .filter_map(|failure| {
            let quarantine_failure = quarantined.contains(&failure.id);
            log::error!(
                "{} -> {}{}(id: {})",
                failure.parent_name,
                failure.name,
                if quarantine_failure {
                    " [QUARANTINED] "
                } else {
                    " "
                },
                failure.id
            );
            if quarantine_failure {
                Some(failure)
            } else {
                None
            }
        })
        .collect();
    quarantine_results.group_is_quarantined =
        quarantine_results.quarantine_results.len() == total_failures;
```

Using `into_iter` relies on the change to accepting `run_result` as a value and not a reference. The reason for using `into_iter` instead of `iter` is to prevent the need for `clone`.

cli/src/runner.rs
```diff
@@ -179,54 +202,66 @@ pub async fn run_quarantine(
...
-    let quarantine_results = if run_result.failures.is_empty() {
-        QuarantineBulkTestStatus {
-            group_is_quarantined: false,
-            quarantine_results: Vec::new(),
-        }
-    } else {
+    let mut quarantine_results = QuarantineBulkTestStatus {
+        group_is_quarantined: false,
+        quarantine_results: Vec::new(),
+    };
```
dfrankland, line 208: nit: can we define this below, where we use/mutate it?

nit: Also, we should add a `#[derive(Default)]` to `QuarantineBulkTestStatus` that way we can simply do the following

```suggestion
    let mut quarantine_results = QuarantineBulkTestStatus::default();
```

cli/src/runner.rs
```diff
@@ -179,54 +202,66 @@ pub async fn run_quarantine(
...
+                is_preview_mode: false,
+                quarantined_tests: Vec::new(),
             }
         })
+    } else {
+        log::info!("No failed tests to quarantine");
+        QuarantineConfig {
+            is_preview_mode: false,
+            quarantined_tests: Vec::new(),
+        }
```
dfrankland, line 228: nit: similarly here we should use `#[derive(Default)]` on `QuarantineConfig` so we don't have to repeat ourselves and we can refactor this code to be much more simple:

```suggestion
        if let Err(ref err) = result {
            log::error!("Failed to get quarantine results: {:?}", err);
        }

        result.unwrap_or_default()
    } else {
        log::info!("No failed tests to quarantine");
        QuarantineConfig::default()
```

cli/src/runner.rs
```diff
@@ -179,54 +202,66 @@ pub async fn run_quarantine(
...
+                "{} -> {} [QUARANTINED] (id: {:?})",
+                failure.parent_name,
+                failure.name,
+                failure.id.clone().unwrap()
+            );
+
+            quarantine_results.quarantine_results.push(failure.clone());
         } else {
-            log::error!("{} -> {}", failure.parent_name, failure.name);
+            log::error!(
```
dfrankland, line 244: an aside: seems odd to me to make these error logs 🤷

cli/src/runner.rs
```diff
@@ -179,54 +202,66 @@ pub async fn run_quarantine(
...
-        .iter()
-        .map(|qr| &qr.name)
-        .collect();
-
+    // quarantine the failed tests
+    let quarantined: HashSet<_> = quarantine_config.quarantined_tests.iter().collect();
     for failure in &run_result.failures {
-        if quarantined_names.contains(&failure.name) {
-            log::error!("{} -> {} [QUARANTINED]", failure.parent_name, failure.name);
+        if quarantined.contains(&failure.id.as_ref().unwrap()) {
```
zaycev, line 234: Please don't use unwrap. If you need to get value from option or result, use `expect("Something unexpected happened")`.

cli/src/runner.rs
```diff
@@ -102,7 +103,25 @@ pub fn build_filesets(
     Ok((file_sets, file_counter))
 }
 
+pub fn get_run_info_id(test: &Test, org_slug: &str, repo: &BundleRepo) -> String {
```
zaycev, line 106: Please add test.

cli/src/runner.rs
zaycev: I agree. How would this work if we have someone with 100k tests failed due to some missconfiguration. I think this should be controller by flag if we even need this.

zaycev:
I would highly recommend to use `log::error` only to log errors. Test failure is not an error, it's expected output.

cli/src/runner.rs
zaycev: Also, there are a lot of unnecessary allocations for each ID. Consider re-using buffer which holds input data:
```rust
pub fn get_run_info_id(buf: &mut Vec<u8>, test: &Test, org_slug: &str, repo: &BundleRepo) -> String {
    buf.clear();
    let mut buf = Vec<u8>::new();
    buf.extend_from_slice(org_slug.as_bytes());
    buf.push(b'/');
    buf.extend_from_slice(repo_full_name.as_bytes());
    buf.push(b'/');
    buf.extend_from_slice(test.file.as_deref().unwrap_or("").as_bytes());
    buf.push(b'#');
    buf.extend_from_slice(test.class_name.as_deref().unwrap_or("").as_bytes());
    buf.push(b'#');
    buf.extend_from_slice(test.parent_name.as_bytes());
    buf.push(b'#');
    buf.extend_from_slice(test.name.as_bytes());
    buf.push(b'#');
    buf.extend_from_slice("JUNIT_TESTCASE".as_bytes());
    uuid::Uuid::new_v5(&uuid::Uuid::NAMESPACE_URL, &buf[..]).to_string()
}
```

You can wrap this into `TestIdBuilder` type of you want, which you can create before traversing tests:

```
            let mut id_builder = TestIdBuilder::default();
            for suite in junitxml.suites {
            ...
                      let mut test = Test {
                            parent_name: parent_name.clone(),
                            name: name.clone(),
                            class_name: case.classname.clone(),
                            file: case.file.clone(),
                            id: Some(id_builder.build_test_id(&test, org_slug, repo)),
                        };
```

```

File: cli/src/runner.rs [hunk 8]
```diff
@@ -157,12 +176,16 @@ pub async fn extract_failed_tests(
                     if failure {
                         let name = case.original_name;
                         log::debug!("Test failed: {} -> {}", parent_name, name);
-                        failures.push(Test {
+                        let mut test = Test {
                             parent_name: parent_name.clone(),
```
zaycev, line 180: Why do we need to clone these fields if you are already moving cases out of `suite.cases` ?

cli/src/runner.rs
```diff
@@ -157,12 +176,16 @@ pub async fn extract_failed_tests(
...
                             name: name.clone(),
                             class_name: case.classname.clone(),
                             file: case.file.clone(),
-                        });
+                            id: None,
```
zaycev, line 184: id: Some(...),

cli/src/runner.rs
```diff
@@ -157,12 +176,16 @@ pub async fn extract_failed_tests(
                     if failure {
                         let name = case.original_name;
                         log::debug!("Test failed: {} -> {}", parent_name, name);
-                        failures.push(Test {
+                        let mut test = Test {
```
zaycev, line 179: Does not have to be mut.

cli/src/runner.rs
zaycev: Also, consider just adding constructor for `Test` type.

cli/src/runner.rs
federicoweber: I hear you, but I think it is much easier to parse the log

federicoweber: Going to move it to info as both you and Vladimir had the same comment

junit.xml
dfrankland, line 1: accidental inclusion?

cli/src/types.rs
```diff
@@ -49,28 +47,23 @@ pub struct Test {
...
-    pub quarantine_results: Vec<QuarantineResult>,
+    pub quarantine_results: Vec<Test>,
+}
+
+#[derive(Debug, Serialize, Clone, Deserialize)]
+pub struct QuarantineConfig {
+    #[serde(rename = "isPreview")]
+    pub is_preview_mode: bool,
+    #[serde(rename = "testIds")]
+    pub quarantined_tests: Vec<String>,
```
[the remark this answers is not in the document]
dfrankland: As a heads up, you can still make `quarantined_tests` a `HashSet` so you don't need to convert it to one later

cli/src/runner.rs
```diff
@@ -172,61 +178,65 @@ pub async fn extract_failed_tests(
...
     };
 
-    // Print out failed tests and their quarantine status
-    let quarantined_names: HashSet<_> = quarantine_results
-        .quarantine_results
-        .iter()
-        .map(|qr| &qr.name)
+    // quarantine the failed tests
+    let mut quarantine_results = QuarantineBulkTestStatus::default();
+    let quarantined: HashSet<_> = quarantine_config.quarantined_tests.iter().collect();
```
dfrankland, line 206: if you define `quarantined_tests` as a `HashSet` you won't have to convert it to one here

dfrankland (approved):

cli/src/runner.rs
federicoweber: ah shoot forgot abut that

PR #87, opened 2024-08-23
always return the exit code from a test command

dfrankland:
Never ever succeed when a test command fails

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -367,42 +367,44 @@ async fn run_test(test_args: TestArgs) -> anyhow::Result<i32> {
         failures: Vec::new(),
     });
 
-    if *use_quarantining {
-        let quarantine_run_result = run_quarantine(
-            run_result,
-            &api_address,
-            token,
-            org_url_slug,
-            &repo,
-            default_delay(),
+    let run_exit_code = run_result.exit_code;
+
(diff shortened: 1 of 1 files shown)


cli/src/main.rs
```diff
@@ -367,42 +367,44 @@ async fn run_test(test_args: TestArgs) -> anyhow::Result<i32> {
...
+    {
         Ok(EXIT_SUCCESS) => (),
         Ok(code) => log::error!("Error uploading test results: {}", code),
+        // TODO(TRUNK-12558): We should fail on configuration error _prior_ to running a test
         Err(e) => log::error!("Error uploading test results: {:?}", e),
-    }
+    };
 
-    Ok(EXIT_SUCCESS)
+    Ok(exit_code)
```
riya-n, line 407: wouldn't this fail on the customer's end? I thought we wanted to avoid that. I am likely missing some context here though

jc3m (approved):

cli/src/main.rs
dfrankland: If the customer's test command fails we need to bubble up their error exit code

acatxnamedvirtue (approved):
don't really know if I understand rust or the context well enough, but it seems reasonable to me

PR #92, opened 2024-09-06
Support drone CI env vars

mmatheson:
Grabbing env based on these docs: https://docs.drone.io/pipeline/environment/reference/


cli/src/constants.rs
```diff
@@ -276,6 +276,78 @@ pub const ENVS_TO_GET: &[&str] = &[
     "BUILD_TRIGGEREDBY_DEFINITIONNAME",
     "BUILD_TRIGGEREDBY_BUILDNUMBER",
     "BUILD_TRIGGEREDBY_PROJECTID",
+    //     "DRONE CI
```
riya-n, line 279: do we need to capture so many? or I suppose we could remove the ones we don't end up using later

riya-n (approved):

PR #93, opened 2024-09-06
store start time before running tests

swimmadude66:
Should fix the issue of output being ignored because the output was modified before the start time


swimmadude66:
closing in favor of #94

PR #94, opened 2024-09-06
Fix start time calculations and coalesce the logic

gnalh:
When wrapping test execution with `test` we had 3 issues:
1. The start time was calculated after the test had been executed. This meant that every file was skipped when doing the quarantining check.
2. The `test` subcommand invokes `upload`, which caused quarantining checks to happen twice.
3. `test` would skip files that it identified as stale, but the `upload` step wouldn't. This caused debugging to be more difficult.

This PR addresses all of the above.

Next steps:
* write some end to end tests for this logic
* return the default exit code when there are no failures found

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -197,13 +197,14 @@ async fn run_upload(
     let quarantine_run_results = if use_quarantining && quarantine_results.is_none() {
         Some(
             run_quarantine(
-                RunResult {
+                &RunResult {
                     exit_code: if failures.is_empty() {
                         EXIT_SUCCESS
                     } else {
                         EXIT_FAILURE
                     },
                     failures,
+                    exec_start: None,
                 },
                 &api_address,
                 &token,
@@ -368,6 +369,7 @@ async fn run_test(test_args: TestArgs) -> anyhow::Result<i32> {
         RunResult {
             exit_code: EXIT_FAILURE,
             failures: Vec::new(),
+            exec_start: None,
         }
     });
 
@@ -376,7 +378,7 @@ async fn run_test(test_args: TestArgs) -> anyhow::Result<i32> {
     let quarantine_run_result = if *use_quarantining {
         Some(
             run_quarantine(
-                run_result,
+                &run_result,
                 &api_address,
                 token,
                 org_url_slug,
@@ -397,7 +399,7 @@ async fn run_test(test_args: TestArgs) -> anyhow::Result<i32> {
     match run_upload(
         upload_args,
         Some(command.join(" ")),
-        quarantine_run_result,
+        None, // don't re-run quarantine checks
         codeowners,
     )
     .await
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -403,7 +403,6 @@ async fn run_test(test_args: TestArgs) -> anyhow::Result<i32> {
         None, // don't re-run quarantine checks
         codeowners,
         run_result.exec_start,
-
     )
     .await
     {
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -189,8 +189,9 @@ async fn run_upload(
 
     let tags = parse_custom_tags(&tags)?;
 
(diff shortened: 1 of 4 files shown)


cli/src/types.rs
```diff
@@ -3,10 +3,8 @@ use std::collections::HashSet;
 use regex::Regex;
 use serde::{Deserialize, Serialize};
 
-use crate::{
-    codeowners::CodeOwners,
-    scanner::{BundleRepo, FileSet},
-};
+use crate::codeowners::CodeOwners;
+use crate::scanner::{BundleRepo, FileSet};
```
gnalh, line 7: `trunk fmt`

cli/src/scanner.rs
```diff
@@ -1,6 +1,8 @@
+use std::format;
+use std::time::SystemTime;
+
 use regex::Regex;
 use serde::Serialize;
-use std::format;
```
gnalh, line 3: `trunk fmt`

cli/src/runner.rs
```diff
@@ -1,18 +1,17 @@
...
+use tokio_retry::Retry;
+
+use crate::clients::get_quarantining_config;
+use crate::codeowners::CodeOwners;
+use crate::constants::{EXIT_FAILURE, EXIT_SUCCESS};
+use crate::scanner::{BundleRepo, FileSet, FileSetCounter};
+use crate::types::{
+    QuarantineBulkTestStatus, QuarantineConfig, QuarantineRunResult, RunResult, Test,
 };
-use tokio_retry::{strategy::ExponentialBackoff, Retry};
```
gnalh, line 15: `trunk fmt`

cli/src/runner.rs
```diff
@@ -107,39 +113,10 @@ pub async fn extract_failed_tests(
...
-                if time <= start {
-                    log::info!(
-                        "Skipping file `{}` because modified time `{}` is before start time `{}`",
-                        file.original_path,
-                        DateTime::<Utc>::from(time),
-                        DateTime::<Utc>::from(start),
-                    );
-                    continue;
-                }
-            }
```
gnalh, line 142: This moves the timestamp checking into the glob scan as an optional parameter. This way it applies to both upload and test subcommands.

cli/src/runner.rs
```diff
@@ -23,18 +22,22 @@ pub async fn run_test_command(
     team: Option<String>,
     codeowners: &Option<CodeOwners>,
 ) -> anyhow::Result<RunResult> {
+    let start = SystemTime::now();
     let exit_code = run_test_and_get_exit_code(command, args).await?;
     log::info!("Command exit code: {}", exit_code);
-    let (file_sets, ..) = build_filesets(repo, output_paths, team, codeowners)?;
+    let (file_sets, ..) = build_filesets(repo, output_paths, team, codeowners, Some(start))?;
     let failures = if exit_code != EXIT_SUCCESS {
-        let start = SystemTime::now();
```
gnalh, line 30: This is the main fix. The start time needs to proceed the test execution.

swimmadude66 (approved):

riya-n (approved):

cli/src/main.rs
```diff
@@ -396,8 +400,9 @@ async fn run_test(test_args: TestArgs) -> anyhow::Result<i32> {
     match run_upload(
         upload_args,
         Some(command.join(" ")),
-        quarantine_run_result,
+        None, // don't re-run quarantine checks
         codeowners,
+        run_result.exec_start,
```
dfrankland, line 405: nit: I think it's better to clone `run_result.exec_start` in this function—before `run_result` is passed to `run_quarantine`—rather than pass `run_result` as a reference and clone it... If I think about it more though, I feel like we should just pass `failures` and `exit_code` directly to `run_quarantine` rather than the entire `run_result`

dfrankland (approved):

cli/src/main.rs
gnalh: Done.

PR #95, opened 2024-09-16
add integration test suite for CLI

dfrankland:
Use junit mocking lib and mock server to test a full upload and quarantine flow

### DIFF
--- a/cli/src/scanner.rs
+++ b/cli/src/scanner.rs
@@ -2,7 +2,7 @@ use std::format;
 use std::time::SystemTime;
 
 use regex::Regex;
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
 
 use crate::codeowners::CodeOwners;
 use crate::constants::{ALLOW_LIST, ENVS_TO_GET};
@@ -34,7 +34,7 @@ impl FileSetCounter {
     }
 }
 
-#[derive(Debug, Serialize, Clone)]
+#[derive(Debug, Serialize, Deserialize, Clone)]
 pub struct FileSet {
     pub file_set_type: FileSetType,
     pub files: Vec<BundledFile>,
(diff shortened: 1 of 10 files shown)


cli/src/scanner.rs
```diff
@@ -296,227 +296,3 @@ impl EnvScanner {
         envs
     }
 }
-
-#[cfg(test)]
-mod tests {
```
dfrankland, line 301: moved to the [`test` module](https://doc.rust-lang.org/book/ch11-03-test-organization.html#the-tests-module-and-cfgtest), mostly so `setup_repo_with_commit` can be reused with other tests

cli/tests/scanner.rs
dfrankland, line 1: moved from `src`

cli/tests/test_utils/mock_git_repo.rs
dfrankland, line 1: moved from `src`

cli/tests/test_utils/mock_server.rs
```diff
@@ -0,0 +1,161 @@
...
+    pub requests: Mutex<Vec<RequestPayload>>,
+    pub host: String,
+}
+
+pub type SharedMockServerState = Arc<MockServerState>;
+
+/// Mock server spawned in a new thread.
+///
+/// NOTE: must use a multithreaded executor to have the server run while running tests
+#[allow(dead_code)] // TODO: move this to its own crate to get rid of the need for this
```
dfrankland, line 40: Regarding this `dead_code` stuff, Rust is setup so that everything within the `tests` module works as integration tests rather than reusable code which is why it complains about these functions. The correct way is to make "`test-utils`" crate and reuse that, but figured there's not really enough here to warrant that. I'm up for splitting it out into a new crate if anyone asks though.

cli/tests/upload.rs
```diff
@@ -0,0 +1,161 @@
...
+            "--org-url-slug",
+            "test-org",
+            "--token",
+            "test-token",
+        ])
+        .assert()
+        // should fail due to quarantine and succeed without quarantining
+        .failure();
+
+    let requests = state.requests.lock().unwrap().clone();
```
dfrankland, line 59: integration tests, wooooo!

riya-n (approved):
🥳 🎉 wooo thank you!!

PR #102, opened 2024-09-23
modularize github codeowners parsing code

dfrankland:
This sets us up to create a common interface to each CODEOWNERS provider that we implement


codeowners/src/github.rs
dfrankland, line 1: Pretty much exactly the same code from `lib.rs`. Moved here to better split up code for each CODEOWNERS provider e.g. GitHub vs GitLab vs Bitbucket

codeowners/src/lib.rs
```diff
@@ -1,455 +1,28 @@
...
-    fs::File,
-    io::{BufRead, BufReader, Read},
-    path::{Path, PathBuf},
-    str::FromStr,
-};
+pub use github::{GitHubOwner, GitHubOwners};
 
-const CODEOWNERS: &str = "CODEOWNERS";
+pub trait OwnersOfPath {
+    type Owner;
```
dfrankland, line 8: split primary public interface of this crate into traits to allow implementing a similar interface for each CODEOWNERS provider

codeowners/src/lib.rs
```diff
@@ -1,455 +1,28 @@
...
-/// Mappings of owners to path patterns
-#[derive(Debug, PartialEq, Clone)]
-pub struct Owners {
-    paths: Vec<(Pattern, Vec<Owner>)>,
-}
-
-impl Owners {
     /// Resolve a list of owners matching a given path
-    pub fn of<P>(&self, path: P) -> Option<&Vec<Owner>>
+    fn of<P>(&self, path: P) -> Option<Vec<Self::Owner>>
```
dfrankland, line 11: removed this reference in favor of an owned `Vec` since it makes the interface more flexible to implement for each CODEOWNERS provider

codeowners/test_fixtures/github/codeowners_example
dfrankland, line 1: Moved from inside of tests since this added a lot of noise

riya-n (approved):
🎉

PR #103, opened 2024-09-23
add gitlab codeowners parsing

dfrankland:
~depends on #102~

adds code to parse GitLab CODEOWNERS exactly the same way that GitLab does in their services:
https://gitlab.com/gitlab-org/gitlab/-/tree/master/ee/lib/gitlab/code_owners

### DIFF
--- a/codeowners/src/gitlab/mod.rs
+++ b/codeowners/src/gitlab/mod.rs
@@ -1,0 +1,228 @@
+mod entry;
+mod error;
+mod file;
+mod reference_extractor;
+mod section;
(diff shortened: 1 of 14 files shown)


codeowners/src/gitlab/mod.rs
dfrankland, line 1: This is what exposes all of the reference implementation via our own interface

riya-n (approved):
Amazing 💯 Still learning rust but this seems fine to me. Thanks so much for adding all the tests also 🥇

PR #104, opened 2024-09-23
incorporate gitlab codeowners parsing to cli

dfrankland:
depends on #103

### DIFF
--- a/cli/src/codeowners.rs
+++ b/cli/src/codeowners.rs
@@ -1,6 +1,9 @@
-use std::path::{Path, PathBuf};
+use std::{
+    fs::File,
+    path::{Path, PathBuf},
+};
 
-use codeowners::{FromPath, GitHubOwners};
+use codeowners::{FromReader, GitHubOwners, GitLabOwners};
 use serde::{Deserialize, Serialize};
 
 use crate::constants::CODEOWNERS_LOCATIONS;
@@ -9,7 +12,7 @@ use crate::constants::CODEOWNERS_LOCATIONS;
 pub struct CodeOwners {
     pub path: PathBuf,
     #[serde(skip_serializing, skip_deserializing)]
-    pub owners: Option<GitHubOwners>,
+    pub owners: Option<Owners>,
 }
 
 impl CodeOwners {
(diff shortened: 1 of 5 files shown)


cli/tests/upload.rs
```diff
@@ -27,12 +28,21 @@ fn generate_mock_junit_xmls<T: AsRef<Path>>(directory: T) {
     JunitMock::write_reports_to_file(directory.as_ref(), reports).unwrap();
 }
 
+fn generate_mock_codeowners<T: AsRef<Path>>(directory: T) {
+    const CODEOWNERS: &str = r#"
+        [Owners of Everything]
```
dfrankland, line 33: GitHub chokes on these headings that GitLab supports

cli/tests/upload.rs
```diff
@@ -138,7 +154,7 @@ async fn upload_bundle() {
     let time_since_junit_modified = chrono::Utc::now()
         - chrono::DateTime::from_timestamp_nanos(bundled_file.last_modified_epoch_ns as i64);
     assert_eq!(time_since_junit_modified.num_minutes(), 0);
-    assert!(bundled_file.owners.is_empty());
+    assert_eq!(bundled_file.owners, ["@user"]);
```
dfrankland, line 157: This is the CODEOWNER for the JUnit.XML file... not any single test referenced by the JUnit.XML file.

cli/src/codeowners.rs
```diff
@@ -30,20 +33,35 @@ impl CodeOwners {
             all_locations.find_map(|location| locate_codeowners(&repo_root, location));
 
         codeowners_path.map(|path| {
-            let owners_result = GitHubOwners::from_path(&path);
+            let owners_result = File::open(&path)
+                .map_err(anyhow::Error::from)
+                .and_then(|file| GitHubOwners::from_reader(&file).map(Owners::GitHubOwners))
+                .or_else(|_| {
```
riya-n, line 39: This makes sense but should we be doing this off of env vars as we add support for other providers?

riya-n (approved):

gnalh (approved):

cli/src/codeowners.rs
dfrankland: Totally, I agree. We've got to redo how we use codeowners anyways since as-is it's not really useful. At that point, I think we should start considering finding and parsing codeowners files based on env.

PR #105, opened 2024-09-27
[TRUNK-12896] Include bundle_upload_id in meta.json on test bundle upload

max-trunk:
Includes `bundle_upload_id` in the `meta.json` metadata file to help facilitate making that value accessible by the ETL job. Also renames a couple function / type names

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -8,7 +8,7 @@ use tokio_retry::strategy::ExponentialBackoff;
 use tokio_retry::Retry;
 use trunk_analytics_cli::bundler::BundlerUtil;
 use trunk_analytics_cli::clients::{
-    create_trunk_repo, create_bundle_upload_intent, put_bundle_to_s3,
+    create_bundle_upload_intent, create_trunk_repo, put_bundle_to_s3,
 };
 use trunk_analytics_cli::codeowners::CodeOwners;
 use trunk_analytics_cli::constants::{
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -2,7 +2,6 @@ use std::env;
 use std::io::Write;
 use std::time::{SystemTime, UNIX_EPOCH};
 
-use anyhow::Context;
 use clap::{Args, Parser, Subcommand};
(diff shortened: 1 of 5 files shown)


cli/src/main.rs
```diff
@@ -240,6 +241,14 @@ async fn run_upload(
 
     let envs = EnvScanner::scan_env();
     let os_info: String = env::consts::OS.to_string();
+
+    let upload_op = Retry::spawn(default_delay(), || {
+        create_bundle_upload_intent(&api_address, &token, &org_url_slug, &repo.repo)
+    })
+    .await?;
+
+    let upload = upload_op.context("Failed to create bundle upload.")?;
```
max-trunk, line 250: Was looking for a way to set `upload` here to a `CreateBundleUploadResponse` from an `Option<CreateBundleUploadResponse>`, failing otherwise, since I don't think we'd want to continue if `upload_op` were `None`. LMK if that's an incorrect assumption. Based on this SO thread :) https://stackoverflow.com/questions/69738600/simplest-way-to-unwrap-an-option-and-return-error-if-none-anyhow

dfrankland: This is good, but I think we can do one better :)

In `create_bundle_upload_intent` I think we can and should change `Option<CreateBundleUploadResponse>` to just `CreateBundleUploadResponse` and then you won't need to do that here. That will also require changing from checking for `is_success` to checking if the API call is not successful (look at `get_quarantining_config` for example). This also makes `create_bundle_upload_intent` more consistent with our other API calls like `get_quarantining_config`. Eventually, I'll get around to cleaning this up so there's a better, more consistent pattern to all of the API calls with retries we make.

Removing the `Option<T>` type is a little auto-magical because it's how `serde` implicitly figures out how to deserialize the response using auto-implemented traits:
https://docs.rs/serde/latest/serde/de/index.html#implementations-of-deserialize-provided-by-serde

max-trunk: Updated! lmk how it looks

dfrankland (approved):
I think this looks great! Definitely makes our code much nicer

PR #106, opened 2024-09-30
Add support for xcresult 

gnalh:


cli/src/main.rs
```diff
@@ -192,9 +196,28 @@ async fn run_upload(
...
+    // NOTE: This temp directory must be in a higher scope, otherwise it will be dropped before we can use it.
+    let junit_temp_dir = tempfile::tempdir()?;
+    if xcresult_path.is_some() && cfg!(target_os = "macos") {
+        let xcresult = XCResultFile::new(xcresult_path.unwrap());
+        let junit = xcresult?.junit();
+        let junit_temp_path = junit_temp_dir.path().join("xcresult_junit.xml");
+        let mut junit_temp = std::fs::File::create(&junit_temp_path)?;
+        junit_temp.write_all(&junit)?;
+        junit_temp.seek(std::io::SeekFrom::Start(0))?;
+        temp_paths.push(junit_temp_path.to_str().unwrap().to_string());
```
gnalh, line 209: Is there a cleaner way to handle this?

cli/tests/xcresult.rs
```diff
@@ -0,0 +1,64 @@
...
+#[cfg(target_os = "linux")]
+#[test]
+fn test_xcresult_with_valid_path_invalid_os() {
+    let path_str = "tests/data/test1.xcresult";
+    let xcresult = XCResultFile::new(path_str.to_string());
+    assert_eq!(
+        xcresult.err().unwrap().to_string(),
+        "xcrun is only available on macOS"
+    );
+}
```
gnalh, line 64: The actual tests. Commenting since the test data files are obfuscating this at the bottom.

cli/src/xcresult.rs
```diff
@@ -0,0 +1,405 @@
...
+            return Err(anyhow::anyhow!(
+                "failed to run xcrun -- please make sure you have xcode installed"
+            ))
+        }
+    };
+    let results_obj_raw = match String::from_utf8(output.stdout) {
+        Ok(val) => val,
+        Err(_) => return Err(anyhow::anyhow!("got non UTF-8 data from xcrun output")),
+    };
+    Ok(results_obj_raw)
```
dfrankland, line 40:
```suggestion
   String::from_utf8(output.stdout)
     .map_err(|_| Err(anyhow::anyhow!("got non UTF-8 data from xcrun output"))
```

gnalh:
Need to investigate why one of the tests fail on x86 darwin, but not aarch. Other than that, this should be good to review.
```
        PASS [   2.080s] trunk-analytics-cli::xcresult test_xcresult_with_invalid_path
        PASS [   2.557s] trunk-analytics-cli::xcresult test_xcresult_with_invalid_xcresult
        PASS [   2.943s] trunk-analytics-cli::xcresult test_complex_xcresult_with_valid_path
        PASS [   1.969s] trunk-analytics-cli::xcresult test_xcresult_with_valid_path
```
vs
```
        PASS [   2.255s] trunk-analytics-cli::xcresult test_xcresult_with_invalid_path
        FAIL [   2.806s] trunk-analytics-cli::xcresult test_complex_xcresult_with_valid_path

--- STDOUT:              trunk-analytics-cli::xcresult test_complex_xcresult_with_valid_path ---

running 1 test
test test_complex_xcresult_with_valid_path ... FAILED

failures:

failures:
    test_complex_xcresult_with_valid_path

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 3 filtered out; finished in 0.67s


--- STDERR:              trunk-analytics-cli::xcresult test_complex_xcresult_with_valid_path ---
thread 'test_complex_xcresult_with_valid_path' panicked at cli/tests/xcresult.rs:64:5:
assertion failed: xcresult.is_ok()
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

        PASS [   2.609s] trunk-analytics-cli::xcresult test_xcresult_with_invalid_xcresult
        PASS [   1.957s] trunk-analytics-cli::xcresult test_xcresult_with_valid_path
```

cli/src/xcresult.rs
```diff
@@ -0,0 +1,346 @@
...
+                .get("startedTime")
+                .and_then(|r| r.get("_value"))
+                .and_then(|r| r.as_str()),
+        ) {
+            let ended_time_parsed = chrono::DateTime::parse_from_str(ended_time, DATE_FORMAT)?;
+            let started_time_parsed = chrono::DateTime::parse_from_str(started_time, DATE_FORMAT)?;
+            let duration = (ended_time_parsed.timestamp_millis()
+                - started_time_parsed.timestamp_millis()) as f32
+                / 1000.0;
+            testsuites_junit.set_time(std::time::Duration::from_secs_f32(duration));
```
riya-n, line 276: why not keep this in ms granularity?

riya-n (approved):
this is really great!! spent some time going through this and it looks good to me but also not a rust/xcresult expert

cli/src/xcresult.rs
gnalh: Good catch!

cli/src/xcresult.rs
```diff
@@ -0,0 +1,345 @@
...
+        })
+    }
+
+    fn find_tests(&self, id: &str) -> Result<serde_json::Value, anyhow::Error> {
+        xcresulttool(&self.path, Some(vec!["--id", id]))
+    }
+
+    fn generate_id(&self, raw_id: &str) -> String {
+        uuid::Uuid::new_v5(&uuid::Uuid::NAMESPACE_URL, raw_id.as_bytes()).to_string()
+    }
```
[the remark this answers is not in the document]
gnalh: Without it there's a warning:
```
the candidate is defined in an impl for the type `XCResultFile` (rustc E0599)
```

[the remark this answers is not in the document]
dfrankland: Weird, that's not the error message I see. Either way, the problem is likely because if the method doesn't accept `self` in some shape or form, you much call the method statically like so: `XCResultFile::generate_id(raw_id)` instead of ~`self.generate_id(raw_id)`~

cli/src/xcresult.rs
```diff
@@ -0,0 +1,345 @@
...
+                            testsuites_junit.add_test_suite(testsuite_junit);
+                        }
+                    }
+                }
+            }
+        }
+        Ok(testsuites_junit)
+    }
+
+    pub fn generate_junits(&self) -> Result<Vec<Report>, anyhow::Error> {
```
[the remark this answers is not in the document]
gnalh:
```
let report_junit: Report = self.junit_report(action)?;
```
This line can error.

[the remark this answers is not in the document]
dfrankland: Ope my bad eyes 🧓

cli/src/main.rs
```diff
@@ -167,10 +219,6 @@ async fn run_upload(
         repo_head_commit_epoch,
     )?;
 
-    if junit_paths.is_empty() {
-        return Err(anyhow::anyhow!("No junit paths provided."));
-    }
-
```
gnalh, line 173: This is redundant since it's checked lower below as well.

xcresult/src/xcresult.rs
```diff
@@ -0,0 +1,348 @@
...
+}
+
+fn xcrun_version() -> anyhow::Result<i32> {
+    let version_raw = xcrun(&["--version"])?;
+    // regex to match version where the output looks like xcrun version 70.
+    let re = regex::Regex::new(r"xcrun version (\d+)")?;
+    lazy_static! {
+        static ref RE: regex::Regex = regex::Regex::new(r"xcrun version (\d+)").unwrap();
+    }
+    let res = re
```
dfrankland, line 36: I think there might have been a small typo here?

```suggestion
    lazy_static! {
        /// regex to match version where the output looks like xcrun version 70.
        static ref RE: regex::Regex = regex::Regex::new(r"xcrun version (\d+)").unwrap();
    }
    let res = RE
```

xcresult/src/xcresult.rs
```diff
@@ -0,0 +1,348 @@
...
+        .captures(&version_raw.to_string())
```
dfrankland, line 37: nit: `version_raw` is of type `String`, so I think we can just pass a reference here

```suggestion
        .captures(&version_raw)
```

xcresult/src/xcresult.rs
```diff
@@ -0,0 +1,348 @@
...
+        .and_then(|capture_group| capture_group.get(1))
+        .map(|version| Ok(version.as_str().parse::<i32>().ok().unwrap_or(0)))
```
dfrankland, line 39: nit: you can use `.unwrap_or_default()` to do the same thing as `.unwrap_or(0)` here—the same goes for elsewhere where `.unwrap_or("")` is used

xcresult/src/xcresult.rs
```diff
@@ -0,0 +1,348 @@
...
+        .unwrap_or_else(|| Err(anyhow::anyhow!("failed to parse xcrun version")));
+    if let Ok(version) = res {
+        return Ok(version);
+    }
+    Err(anyhow::anyhow!("failed to parse xcrun version"))
```
dfrankland, line 44: nit: I think you've got some redundant unwrapping of `Result`s here

```suggestion
    RE
        .captures(&version_raw)
        .and_then(|capture_group| capture_group.get(1))
        .map(|version| Ok(version.as_str().parse::<i32>().ok().unwrap_or(0)))
        .unwrap_or_else(|| Err(anyhow::anyhow!("failed to parse xcrun version")))
```

xcresult/src/xcresult.rs
```diff
@@ -0,0 +1,348 @@
...
+}
+
+fn xcresulttool(
+    path: &str,
+    options: Option<&Vec<&str>>,
```
dfrankland, line 49: optional: Same comments about `&[T]` and `AsRef<str>` vs `&Vec<T>` and `&str`

xcresult/src/xcresult.rs
```diff
@@ -0,0 +1,348 @@
...
+) -> Result<serde_json::Value, anyhow::Error> {
```
dfrankland, line 50: nit: keep consistent with the use of `anyhow::Result<serde_json::Value>`—here and below

xcresult/src/xcresult.rs
```diff
@@ -0,0 +1,348 @@
...
+        base_args.extend(val);
+    }
+    let output = xcrun(&base_args)?;
+    serde_json::from_str(&output)
+        .map_err(|_| anyhow::anyhow!("failed to parse json from xcrun output"))
+}
+
+impl XCResult {
+    pub fn new(path: String) -> Result<XCResult, anyhow::Error> {
+        let binding = fs::canonicalize(path.clone())
```
dfrankland, line 66: nit: we don't really need ownership of `path`, nor do we need to clone it here—I recommend to double check the other function signatures to see if it's possible to accept `AsRef<str>`

```suggestion
    pub fn new<T: AsRef<str>>(path: T) -> Result<XCResult, anyhow::Error> {
        let binding = fs::canonicalize(path.as_ref())
```

xcresult/src/xcresult.rs
```diff
@@ -0,0 +1,348 @@
...
+        if let Some(identifier) = testsuite
+            .get("identifierURL")
+            .and_then(|r| r.get("_value"))
+            .and_then(|r| r.as_str())
+        {
+            index_map.append(&mut indexmap! {
+                XmlString::new("id") => XmlString::new(XCResult::generate_id(identifier)),
+            });
+        }
+        testsuite_junit.extra = index_map;
```
dfrankland, line 219: Similar [comment to before](https://github.com/trunk-io/analytics-cli/pull/106#discussion_r1783343394): no need to create a new `IndexMap` when we can just mutate `.extra` directly

xcresult/src/xcresult.rs
```diff
@@ -0,0 +1,348 @@
...
+        }
+        Ok(testsuites_junit)
+    }
+
+    pub fn generate_junits(&self) -> Result<Vec<Report>, anyhow::Error> {
+        let mut report_junits: Vec<Report> = Vec::new();
+        if let Some(actions) = self
+            .results_obj
+            .get("actions")
+            .and_then(|a| a.get("_values"))
```
dfrankland, line 334: very optional, might be a future improvement rather than being included in this PR: there seems to be a lot of `&'static str`—like `"_values"`—which we repeat throughout the parsing code. We can make it a bit easier to understand that these are all the same values by making these values constants which we reuse.

dfrankland (approved):

xcresult/src/xcresult.rs
gnalh: Maybe I'm misunderstanding, but I don't believe we are grabbing the same values. The structure of the json is such that it's "key" -> "_values"/"_value", which correspond to the underlying value for the particular field. The key is different each time in the above call.

dfrankland: `"_value"` might be a better example since it's repeated more often. What I'm proposing is something like

```rs
const RESULTS_FIELD_VALUE: &str: "_value";

// ...
x.get(RESULTS_FIELD_VALUE)
// ...
y.get(RESULTS_FIELD_VALUE)
// ...
z.get(RESULTS_FIELD_VALUE)
```

PR #107, opened 2024-10-01
1. improve junit-mock for testing

dfrankland:
improves junit-mocking for better tests, especially ones concerning validation

PR #108, opened 2024-10-01
2. add context crate for junit, env parsing and validation

dfrankland:
depends on #107

Adds a `context` crate which we can reuse in the CLI (and elsewhere with bindings) for parsing and validating the two primary sources of context we collect: environmental variables and JUnit XMLs.

### DIFF
--- a/context/src/env/validator.rs
+++ b/context/src/env/validator.rs
@@ -1,0 +1,274 @@
+use thiserror::Error;
+
+use crate::{validate_field_len, FieldLen};
+
+use super::parser::{BranchClass, CIInfo};
+
+pub const MAX_BRANCH_NAME_LEN: usize = 36;
+pub const MAX_EMAIL_LEN: usize = 254;
+pub const MAX_FIELD_LEN: usize = 1000;
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub enum EnvValidationLevel {
+    Valid = 0,
+    SubOptimal = 1,
+    Invalid = 2,
+}
+
+impl Default for EnvValidationLevel {
+    fn default() -> Self {
+        Self::Valid
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum EnvValidationIssue {
+    SubOptimal(EnvValidationIssueSubOptimal),
+    Invalid(EnvValidationIssueInvalid),
+}
+
+#[derive(Error, Debug, Clone, PartialEq, Eq)]
+pub enum EnvValidationIssueSubOptimal {
+    #[error("CI info actor too short")]
+    CIInfoActorTooShort(String),
+    #[error("CI info actor too long, truncated to {}", MAX_FIELD_LEN)]
+    CIInfoActorTooLong(String),
+    #[error("CI info author email too short")]
+    CIInfoAuthorEmailTooShort(String),
+    #[error("CI info author email too long, truncated to {}", MAX_EMAIL_LEN)]
+    CIInfoAuthorEmailTooLong(String),
+    #[error("CI info author name too short")]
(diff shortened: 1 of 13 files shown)


context/src/env/validator.rs
dfrankland, line 1: This builds upon https://github.com/trunk-io/trunk/blob/33c8b6d4810478c505dd1a4df7e7a857468aab4a/trunk/services/metrics-data-pipeline/trunk_etl_validator/validator.py#L149

context/src/junit/date_parser.rs
dfrankland, line 1: This builds upon https://github.com/trunk-io/trunk/blob/main/trunk/services/metrics-data-pipeline/trunk_etl_junit_parser/date_parser.py

context/src/junit/date_parser.rs
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    fn parse_naive_date<T: AsRef<str>>(date_str: T) -> Option<TimestampAndOffset> {
+        SpeedateDate::parse_str(date_str.as_ref())
+            .ok()
+            .map(|d| TimestampAndOffset {
+                timestamp_secs_micros: Some((d.timestamp(), 0)),
+                offset_secs: None,
+            })
+    }
```
dfrankland, line 55: I'm using [`speedate`](https://docs.rs/speedate/latest/speedate/index.html) here because it's faster at parsing dates than [`chrono`](https://docs.rs/chrono/latest/chrono/index.html) (the de-facto standard date-time crate) and it is also less strict about the date format—this makes it ideal for trying to parse several different ISO 8601 formats without sacrificing performance.

context/src/junit/parser.rs
dfrankland, line 1: This builds upon https://github.com/trunk-io/trunk/blob/main/trunk/services/metrics-data-pipeline/trunk_etl_junit_parser/xml_parser.py

context/src/junit/validator.rs
dfrankland, line 1: This builds upon https://github.com/trunk-io/trunk/blob/33c8b6d4810478c505dd1a4df7e7a857468aab4a/trunk/services/metrics-data-pipeline/trunk_etl_validator/validator.py#L44

context/src/lib.rs
```diff
@@ -0,0 +1,36 @@
+// NOTE: This lint isn't applicable since we compile with nightly
+/* trunk-ignore(clippy/E0554) */
+#![feature(round_char_boundary)]
```
dfrankland, line 3: This feature is for [`floor_char_boundary`](https://doc.rust-lang.org/std/primitive.str.html#method.floor_char_boundary) which is used to safely split strings on UTF-8 character boundaries

context/tests/env.rs
dfrankland, line 1: Built upon https://github.com/trunk-io/trunk/blob/main/trunk/services/metrics-data-pipeline/trunk_etl_env_parser/test.py

context/src/junit/parser.rs
```diff
@@ -0,0 +1,588 @@
...
+        Default::default()
+    }
+
+    pub fn errors(&self) -> &Vec<JunitParseError> {
+        &self.errors
+    }
+
+    pub fn reports(&self) -> &Vec<Report> {
+        &self.reports
+    }
```
dfrankland, line 98: I'm using [`quick-junit`](https://docs.rs/quick-junit/latest/quick_junit/) since it is what [`cargo-nextest`](https://nexte.st/) uses and is full-featured including test reruns. Most JUnit parsers do not include all the data that `quick-junit` does making it ideal for our use-cases. Unfortunately, there's not a parser / deserializer built into `quick-junit` (although we could contribute this back), so I've created this one here.

gnalh (approved):

PR #109, opened 2024-10-01
3. split out api, test_utils, and bundlerepo from cli

dfrankland:
depends on #108

Primary purpose of this change is to split out `BundleRepo` from `cli` to `context` to add validation for the context we gather from git repos.

`test_utils` got split out from `cli` because `BundleRepo` tests rely on some of the utils there.

`api` got split out from `cli` because `test_utils` relies on structs used in mocking API calls.


context/src/repo/mod.rs
dfrankland, line 1: Moved from CLI to this crate since we will want to use validate and use git details as a part of context we use for uploads.

context/src/repo/mod.rs
```diff
@@ -0,0 +1,198 @@
...
+        {
+            if let Some(target_id) = remote_branch.target().try_id() {
+                if target_id.as_bytes() == remote_branch.id().as_bytes() {
+                    return Ok(remote_branch.name().to_path().to_str().map(String::from));
+                }
+            }
+        }
+        Ok(None)
+    }
+}
```
dfrankland, line 135: I re-wrote this to be clearer, have less variables getting defined and mutated, and easier to make bindings to for other languages

context/src/repo/mod.rs
```diff
@@ -0,0 +1,198 @@
...
+
+#[derive(Debug, Clone, Default, Serialize, Deserialize, PartialEq, Eq)]
+pub struct RepoUrlParts {
+    pub host: String,
+    pub owner: String,
+    pub name: String,
+}
+
+impl RepoUrlParts {
```
dfrankland, line 144: I renamed this from `Repo` since it was confusing seeing `repo.repo` everywhere. This is really only the results of parsing a repo URL.

context/src/repo/mod.rs
```diff
@@ -0,0 +1,198 @@
...
+
+        if host.is_empty() || owner.is_empty() || name.is_empty() {
+            return Err(anyhow::anyhow!(
+                "Invalid repo url format. Expected non-empty parts: {:?} (url: {})",
+                parts,
+                url
+            ));
+        }
+
+        Ok(Self { host, owner, name })
```
dfrankland, line 196: I only copied this logic, but we should probably follow some best practices here, like:
1. Use the [`url`](https://docs.rs/url/latest/url/) crate
2. [Use `lazy_static` for regexes](https://docs.rs/regex/1.11.0/regex/#avoid-re-compiling-regexes-especially-in-a-loop)

context/tests/repo.rs
dfrankland, line 1: Pretty much completely copied without logic changes (just updated to use `RepoUrlParts`

test_utils/Cargo.toml
dfrankland, line 1: Finally split this out from `cli` to make it easier to reuse some utils in test for the `context` crate

api/Cargo.toml
dfrankland, line 1: This crate got split out from `cli` to be reused in `test_utils`

cli/src/clients.rs
```diff
@@ -1,12 +1,12 @@
...
 
 use anyhow::Context;
-
-use crate::types::{
+use api::{
     CreateBundleUploadRequest, CreateBundleUploadResponse, CreateRepoRequest,
-    GetQuarantineBulkTestStatusRequest, QuarantineConfig, Repo,
+    GetQuarantineBulkTestStatusRequest, QuarantineConfig,
 };
+use context::repo::RepoUrlParts;
```
gnalh, line 8: Nit: This no longer aligns with how we name it in our other repos. We should be consistent.

cli/src/clients.rs
```diff
@@ -16,7 +16,7 @@ pub async fn create_trunk_repo(
     origin: &str,
     api_token: &str,
     org_slug: &str,
-    repo: &Repo,
+    repo: &RepoUrlParts,
```
gnalh, line 19: Nit, update the parameter to match if we stick with this new name type.
```suggestion
    repo_url_parts: &RepoUrlParts,
```

context/src/repo/mod.rs
gnalh: See my nit above, but this does conflict with how we've named this elsewhere.

gnalh (approved):
Overall LGTM, just a few nits on names.

riya-n (approved):

PR #110, opened 2024-10-01
4. add bundle repo validator

dfrankland:
depends on #109 

Adds validation for git repo details we collect independently from other context we collect

### DIFF
--- a/context/src/lib.rs
+++ b/context/src/lib.rs
@@ -5,33 +5,4 @@
 pub mod env;
 pub mod junit;
 pub mod repo;
-
-fn safe_truncate_string<'a, const MAX_LEN: usize, T: AsRef<str>>(value: &'a T) -> &'a str {
-    safe_truncate_str::<MAX_LEN>(value.as_ref())
-}
-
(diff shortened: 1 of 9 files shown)


context/src/lib.rs
```diff
@@ -5,33 +5,4 @@
...
-        FieldLen::TooShort(String::from(trimmed_field))
-    } else if (1..=MAX_LEN).contains(&trimmed_field_len) {
-        FieldLen::Valid
-    } else {
-        FieldLen::TooLong(String::from(safe_truncate_string::<MAX_LEN, _>(
-            &trimmed_field,
-        )))
-    }
-}
+mod string_safety;
```
dfrankland, line 8: Moved this to its own module for better organization

context/src/repo/validator.rs
dfrankland, line 1: Currently we don't validate this data we collect independently from other pieces of context like env. This helps to check independently whether we have this information before we mash everything together.

context/src/string_safety.rs
dfrankland, line 1: moved from `lib.rs` for organization

riya-n (approved):
nice! lgtm

gnalh (approved):

PR #111, opened 2024-10-01
5. add python + js bindings for context crate

dfrankland:
depends on #110 

Adds bindings for Python and JS that can be used elsewhere in our apps and services. As a follow up, we need to figure out how to test, build, and publish these bindings.

[`maturin`](https://www.maturin.rs/) is used for Python bindings
[`wasm-pack`](https://rustwasm.github.io/docs/wasm-pack/) is used for JS bindings

### DIFF
--- a/.github/actions/build_target/action.yaml
+++ b/.github/actions/build_target/action.yaml
@@ -23,7 +23,7 @@ runs:
 
     - name: Compile binary
       shell: bash
-      run: cargo build -q --all --profile=${{ inputs.profile }} --target ${{ inputs.target }}
+      run: cargo build -q --workspace ${{ contains(inputs.target, 'musl') && '--exclude context-js --exclude context-py' || '' }} --profile=${{ inputs.profile }} --target ${{ inputs.target }}
 
     - name: Create binary with debug info
       shell: bash
--- a/context/Cargo.toml
+++ b/context/Cargo.toml
@@ -12,7 +12,9 @@ tempfile = "3.2.0"
 [dependencies]
 anyhow = "1.0.44"
 chrono = "0.4.33"
-gix = { version = "0.63.0", default-features = false, features = [], optional = true }
+gix = { version = "0.63.0", default-features = false, features = [
+], optional = true }
+js-sys = { version = "0.3.70", optional = true }
 openssl = { version = "0.10.66", features = ["vendored"], optional = true }
 pyo3 = { version = "0.20.0", optional = true }
 quick-junit = "0.5.0"
@@ -22,8 +24,11 @@ serde = { version = "1.0.130", default-features = false, features = ["derive"] }
 serde_json = "1.0.68"
 speedate = "0.14.4"
 thiserror = "1.0.63"
+wasm-bindgen = { version = "0.2.84", optional = true }
 
 [features]
 default = ["git-access"]
 git-access = ["dep:gix", "dep:openssl"]
-pyo3 = ["dep:pyo3"]
+bindings = []
+pyo3 = ["bindings", "dep:pyo3"]
+wasm = ["bindings", "dep:wasm-bindgen", "dep:js-sys"]
--- a/context/Cargo.toml
+++ b/context/Cargo.toml
@@ -12,9 +12,9 @@ tempfile = "3.2.0"
 [dependencies]
 anyhow = "1.0.44"
 chrono = "0.4.33"
-gix = { version = "0.63.0", default-features = false, features = [] }
-openssl = { version = "0.10.66", features = ["vendored"] }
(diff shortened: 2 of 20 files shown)


.github/actions/build_target/action.yaml
```diff
@@ -23,7 +23,7 @@ runs:
 
     - name: Compile binary
       shell: bash
-      run: cargo build -q --all --profile=${{ inputs.profile }} --target ${{ inputs.target }}
+      run: cargo build -q --workspace ${{ contains(inputs.target, 'musl') && '--exclude context-js --exclude context-py' || '' }} --profile=${{ inputs.profile }} --target ${{ inputs.target }}
```
dfrankland, line 26: `musl` targets cannot build bindings meant for other targets

context/Cargo.toml
```diff
@@ -12,12 +12,23 @@ tempfile = "3.2.0"
...
 speedate = "0.14.4"
 thiserror = "1.0.63"
+wasm-bindgen = { version = "0.2.84", optional = true }
+
+[features]
+default = ["git-access"]
+git-access = ["dep:gix", "dep:openssl"]
+bindings = []
+pyo3 = ["bindings", "dep:pyo3"]
+wasm = ["bindings", "dep:wasm-bindgen", "dep:js-sys"]
```
dfrankland, line 34: `gix` cannot be compiled to wasm since it depends on a unix OS and additionally, we don't need to include bindings implementations for the CLI, so features have been added to conditionally include bindings and compatibility.

context/src/env/validator.rs
```diff
@@ -244,24 +259,48 @@ pub fn validate(ci_info: &CIInfo) -> EnvValidation {
...
-    pub fn issues(&self) -> &[EnvValidationIssue] {
-        &self.issues
+    pub fn issues_flat(&self) -> Vec<EnvValidationFlatIssue> {
+        self.issues
+            .iter()
+            .map(|i| EnvValidationFlatIssue {
+                level: EnvValidationLevel::from(i),
+                error_message: i.to_string(),
+            })
+            .collect()
```
dfrankland, line 292: Python and wasm don't work with enums that have nested (tuple or struct variants), so flattening can be done to essentially give the same data. Also, references cannot be passed back and forth, so owned data is returned instead.

context/src/junit/bindings.rs
dfrankland, line 1: Since `quick-junit`'s `Report` and nested data is not suitable for being passed back and forth via bindings, we must map the data to and fro using our own interop like this

context/src/junit/validator.rs
```diff
@@ -160,14 +166,26 @@ pub fn validate(report: &Report) -> JunitReportValidation {
...
+}
+
+#[cfg_attr(feature = "pyo3", pymethods)]
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
 impl JunitReportValidation {
-    pub fn test_suites(&self) -> &[JunitTestSuiteValidation] {
-        &self.test_suites
+    pub fn test_suites_owned(&self) -> Vec<JunitTestSuiteValidation> {
+        self.test_suites.clone()
     }
```
dfrankland, line 189: As mentioned above, we cannot pass references via bindings

context/src/repo/mod.rs
```diff
@@ -53,39 +58,47 @@ impl BundleRepo {
...
-                    .map(|s| s.to_string())
-            });
-
-            if let Ok(mut git_head) = git_repo.head() {
-                bundle_repo_options.repo_head_branch = bundle_repo_options
-                    .repo_head_branch
-                    .or_else(|| git_head.referent_name().map(|s| s.as_bstr().to_string()))
-                    .or_else(|| {
-                        Self::git_head_branch_from_remote_branches(&git_repo)
+            // If repo root found, try to get repo details from git.
```
dfrankland, line 63: This looks like a lot of code change, but it's just indentation (turn white-space diffing off). Using a feature we can prevent including any code for reading git repo details which is important for wasm compilation.

context/src/repo/mod.rs
```diff
@@ -136,6 +148,37 @@ impl BundleRepo {
     }
 }
 
+#[cfg(feature = "pyo3")]
+#[pymethods]
+impl BundleRepo {
+    #[new]
+    fn py_new(
```
dfrankland, line 155: Python explicitly needs a `new` constructor otherwise this cannot be instantiated in Python code. This struct needs to be create in Python since it can read from bundle uploads and check for validation.

riya-n (approved):

gnalh (approved):
LGTM

PR #115, opened 2024-10-04
add CI build and test for context-py

dfrankland:
Adds a build and test process for `context-py` Python bindings


.github/workflows/pyo3.yml
```diff
@@ -0,0 +1,125 @@
...
+          container: "2_28"
+          before-script-linux: |
+            # Needed for `openssl-sys`:
+            # https://docs.rs/openssl/latest/openssl/
+            if [ -f /usr/bin/dnf ]; then
+              # quay.io/pypa/manylinux_2_28_x86_64:latest
+              /usr/bin/dnf upgrade
+              /usr/bin/dnf install -y pkgconf perl-IPC-Cmd openssl-devel python3-pip
+              /usr/bin/ln -s /usr/bin/pip3 /usr/bin/pip
+              /usr/bin/ln -sf /usr/bin/python3.11 /usr/bin/python3
```
gnalh, line 51: What is this?

gnalh (approved):
Overall looks good. Now that we have Python code, can we also add python linters into this repo?

.github/workflows/pyo3.yml
dfrankland:
```
            # Needed for `openssl-sys`:
            # https://docs.rs/openssl/latest/openssl/
```

rustfmt.toml
dfrankland, line 1: This file is not used by `trunk` CLI (and IMO is _very_ divergent from normal Rust code styling)

.trunk/configs/.bandit
```diff
@@ -0,0 +1,2 @@
+[bandit]
+exclude = tests
```
dfrankland, line 2: `assert`s are used in tests and bandit says to exclude `tests` 🤷 
https://bandit.readthedocs.io/en/latest/config.html#exclusions

.trunk/trunk.yaml
dfrankland, line 1: upgraded to support python linters and formatters

.trunk/trunk.yaml
```diff
@@ -7,34 +7,38 @@ cli:
...
+    - trufflehog@3.82.6
     - yamllint@1.35.1
 actions:
-  disabled:
+  enabled:
     - trunk-announce
     - trunk-check-pre-push
     - trunk-fmt-pre-commit
-  enabled:
     - trunk-upgrade-available
```
dfrankland, line 42: I think we want to use these

PR #118, opened 2024-10-05
create api_client for unified way of calling apis

dfrankland:
We previously separated the concerns of API calls in a very disparate way. This brings them together in one place and reduces some of the boiler plate which we copied each time we added an API call.

I added some tests for `CallApi` in here, but the rest of the testing required more test setup, so I moved it out to #127


cli/src/main.rs
```diff
@@ -254,7 +235,7 @@ async fn run_upload(
         return Err(anyhow::anyhow!("No JUnit files found to upload."));
     }
 
-    let failures = extract_failed_tests(&repo, &org_url_slug, &file_sets).await?;
+    let failures = extract_failed_tests(&repo, &org_url_slug, &file_sets).await;
```
dfrankland, line 238: This function never fails so I removed `Result` from the return type

cli/src/main.rs
```diff
@@ -265,15 +246,15 @@ async fn run_upload(
...
                 failures,
-                &api_address,
-                &token,
-                &org_url_slug,
-                &repo,
-                default_delay(),
+                exit_code,
             )
-            .await?,
+            .await,
```
dfrankland, line 257: same here, there's no failures, so no need for `Result` in the return type

cli/src/main.rs
```diff
@@ -489,12 +462,6 @@ async fn run(cli: Cli) -> anyhow::Result<i32> {
     }
 }
 
-fn default_delay() -> std::iter::Take<ExponentialBackoff> {
-    ExponentialBackoff::from_millis(RETRY_BASE_MS)
-        .factor(RETRY_FACTOR)
-        .take(RETRY_COUNT)
-}
-
```
dfrankland, line 497: moved to `call_api`

cli/src/main.rs
```diff
@@ -514,10 +481,3 @@ fn setup_logger() -> anyhow::Result<()> {
     builder.init();
     Ok(())
 }
-
-fn get_api_address() -> String {
-    std::env::var(TRUNK_PUBLIC_API_ADDRESS_ENV)
-        .ok()
-        .and_then(|s| if s.is_empty() { None } else { Some(s) })
-        .unwrap_or_else(|| DEFAULT_ORIGIN.to_string())
-}
```
dfrankland, line 523: moved to `ApiClient`

cli/src/runner.rs
```diff
@@ -28,7 +25,7 @@ pub async fn run_test_command(
     log::info!("Command exit code: {}", exit_code);
     let (file_sets, ..) = build_filesets(repo, output_paths, team, codeowners, Some(start))?;
     let failures = if exit_code != EXIT_SUCCESS {
-        extract_failed_tests(repo, org_slug, &file_sets).await?
+        extract_failed_tests(repo, org_slug, &file_sets).await
```
dfrankland, line 28: this never fails, so no need for `Result`

cli/src/runner.rs
```diff
@@ -162,32 +159,27 @@ pub async fn extract_failed_tests(
             }
         }
     }
-    Ok(failures)
+    failures
```
dfrankland, line 162: not needed

cli/src/runner.rs
```diff
@@ -232,8 +224,8 @@ pub async fn run_quarantine(
         exit_code
     };
 
-    Ok(QuarantineRunResult {
+    QuarantineRunResult {
         exit_code,
         quarantine_status: quarantine_results,
-    })
```
dfrankland, line 238: same here, not needed

cli/src/utils.rs
```diff
@@ -3,26 +3,6 @@ use crate::types::CustomTag;
...
-             slug may be incorrect - find \
-             it on the Trunk app (Settings \
-             -> Manage Organization -> \
-             Organization Slug)."
-        }
-        _ => "For more help, contact us at https://slack.trunk.io/",
-    }
-    .to_string()
-}
-
```
dfrankland, line 25: Moved to `api_client/mod.rs`

gnalh (approved):

cli/src/api_client/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+                status_code_help(
+                    &response,
+                    CheckUnauthorized::Check,
+                    CheckNotFound::DoNotCheck,
+                    |_| format!("Failed to create repo."),
+                )
+            },
+            log_progress_message: |time_elapsed, _| {
+                format!("Communicating with Trunk services is taking longer than expected. It has taken {} seconds so far.", time_elapsed.as_secs())
+            },
```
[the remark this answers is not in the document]
dfrankland: Yeah, that hasn't changed. We should definitely go through and prevent retry for things like "not found" or "unauthorized" statuses

cli/src/api_client/call_api.rs
```diff
@@ -0,0 +1,220 @@
+use std::time::Duration;
+
+use tokio::time::{self, Instant};
+use tokio_retry::{strategy::ExponentialBackoff, Action, Retry};
+
+// Tokio-retry uses base ^ retry * factor formula.
+// This will give us 8ms, 64ms, 512ms, 4096ms, 32768ms
```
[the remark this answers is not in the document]
dfrankland: No change here. We can adjust as needed

cli/src/api_client/call_api.rs
```diff
@@ -0,0 +1,220 @@
...
+        },
+        time::Duration,
+    };
+
+    use lazy_static::lazy_static;
+    use tokio::time;
+
+    use super::{
+        CallApi, CHECK_PROGRESS_INTERVAL_SECS, REPORT_SLOW_PROGRESS_TIMEOUT_SECS, RETRY_COUNT,
+    };
```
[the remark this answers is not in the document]
dfrankland: It's because these are within a `mod`—`mod` has it's own scope, just like a new file, so all imports must be defined for all code within a `mod`

PR #121, opened 2024-10-08
Override id field when specified in junit

gnalh:
* Override id field when specified in junit
* Migrate to quick-unit 
* migrate to our junit parser

### DIFF
--- a/cli/src/runner.rs
+++ b/cli/src/runner.rs
@@ -135,40 +135,28 @@ pub async fn extract_failed_tests(
                 }
             };
             for report in junitxml.reports() {
-                let test_suites = report.test_suites.clone();
-                for suite in test_suites {
-                    let parent_name = suite.name.into_string();
-                    for case in suite.test_cases {
-                        let is_failure = match case.status {
-                            TestCaseStatus::NonSuccess { .. } => true,
-                            _ => false,
-                        };
-                        if is_failure {
-                            let name = case.name.into_string();
-                            let classname = match case.classname {
-                                Some(classname) => Some(classname.into_string()),
-                                None => None,
-                            };
-                            let file = match case.extra.get("file") {
-                                Some(file) => Some(file.clone().into_string()),
-                                None => None,
-                            };
-                            let id = match case.extra.get("id") {
-                                Some(id) => Some(id.clone().into_string()),
-                                None => None,
-                            };
-                            let test = Test::new(
-                                name,
-                                parent_name.clone(),
-                                classname,
-                                file,
-                                id,
-                                org_slug,
-                                repo,
-                            );
-
-                            failures.push(test);
+                for suite in &report.test_suites {
+                    let parent_name = String::from(suite.name.as_str());
+                    for case in &suite.test_cases {
+                        if !matches!(case.status, TestCaseStatus::NonSuccess { .. }) {
+                            continue;
                         }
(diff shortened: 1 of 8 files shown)


cli/src/runner.rs
```diff
@@ -125,29 +126,49 @@ pub async fn extract_failed_tests(
...
+                                id,
+                                org_slug,
+                                repo,
+                            );
+
+                            failures.push(test);
+                        }
                     }
                 }
             }
```
dfrankland, line 174: There's a little more copying than necessary here. We can iterate by reference and clone only what's needed. Also, I sprinkled in some small changes to make control flow less verbose.

```suggestion
            for report in junitxml.reports() {
                for suite in &report.test_suites {
                    let parent_name = String::from(suite.name.as_str());
                    for case in &suite.test_cases {
                        if !matches!(case.status, TestCaseStatus::NonSuccess { .. }) {
                            continue;
                        }

                        let name = String::from(case.name.as_str());
                        let classname = case.classname.as_ref().map(|s| String::from(s.as_str()));
                        let file = case.extra.get("file").map(|s| String::from(s.as_str()));
                        let id = case.extra.get("id").map(|s| String::from(s.as_str()));
                        let test = Test::new(
                            name,
                            parent_name.clone(),
                            classname,
                            file,
                            id,
                            org_slug,
                            repo,
                        );

                        failures.push(test);
                    }
                }
            }
```

Also, if wanted, we can be fancy and do something like this too:
```rs
let xml_string_to_string = |s: &quick_junit::XmlString| String::from(s.as_str());
let classname = case.classname.as_ref().map(xml_string_to_string);
let file = case.extra.get("file").map(xml_string_to_string);
let id = case.extra.get("id").map(xml_string_to_string);
```

dfrankland (approved):

context/src/junit/parser.rs
```diff
@@ -326,6 +331,10 @@ impl JunitParser {
             test_case.extra.insert("filepath".into(), filepath.into());
         }
 
+        if let Some(id) = parse_attr::id(e) {
+            test_case.extra.insert("id".into(), id.into());
```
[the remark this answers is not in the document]
gnalh: Yeah I can change that. We aren't doing it above for the other extra attrs. I'll update all of them to match.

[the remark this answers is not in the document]
dfrankland: That was probably my mistake. Thanks for cleaning things up!

cli/src/runner.rs
gnalh: I like your style.

context/src/junit/validator.rs
```diff
@@ -65,6 +65,17 @@ pub fn validate(report: &Report) -> JunitReportValidation {
...
+            let test_case_id = uuid::Uuid::parse_str(raw_test_suite_id).unwrap_or_default();
+            if test_case_id.get_version() != Some(uuid::Version::Sha1) {
+                test_suite_validation.add_issue(JunitValidationIssue::Invalid(
+                    JunitTestSuiteValidationIssueInvalid::TestSuiteInvalidId(
+                        raw_test_suite_id.to_string().clone(),
+                    ),
+                ));
+            }
+        }
+
```
gnalh, line 78: @dfrankland I added some validation + tests. Let me know if you want to review this before I merge.

PR #122, opened 2024-10-09
force sentry to collect events with env dev for tests

dfrankland:
I'm adding more Sentry events and I'm seeing these sometimes get sent using the prod env in tests 😬 

This can happen when `cargo test --release` is used and Sentry defaults its environment based on the [`debug_assertions`](https://doc.rust-lang.org/reference/conditional-compilation.html#debug_assertions) feature flag which is turned off/on by `--release`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,7 @@
 members = [
   "api",
   "cli",
+  "cli-tests",
   "codeowners",
   "context",
   "context-js",
--- a/cli-tests/src/upload.rs
+++ b/cli-tests/src/upload.rs
@@ -144,7 +144,7 @@ async fn upload_bundle() {
     assert_eq!(bundle_meta.envs.get("CI"), Some(&String::from("1")));
     let time_since_upload = chrono::Utc::now()
         - chrono::DateTime::from_timestamp(bundle_meta.upload_time_epoch as i64, 0).unwrap();
-    assert!(time_since_upload.num_minutes() <= 5);
+    more_asserts::assert_lt!(time_since_upload.num_minutes(), 5);
     assert_eq!(bundle_meta.test_command, None);
     assert!(bundle_meta.os_info.is_some());
     assert!(bundle_meta.quarantined_tests.is_empty());
@@ -172,7 +172,7 @@ async fn upload_bundle() {
     );
     let time_since_junit_modified = chrono::Utc::now()
         - chrono::DateTime::from_timestamp_nanos(bundled_file.last_modified_epoch_ns as i64);
-    assert!(time_since_junit_modified.num_minutes() <= 5);
+    more_asserts::assert_lt!(time_since_junit_modified.num_minutes(), 5);
     assert_eq!(bundled_file.owners, ["@user"]);
     assert_eq!(bundled_file.team, None);
 
--- a/cli-tests/src/upload.rs
+++ b/cli-tests/src/upload.rs
@@ -144,7 +144,7 @@ async fn upload_bundle() {
     assert_eq!(bundle_meta.envs.get("CI"), Some(&String::from("1")));
     let time_since_upload = chrono::Utc::now()
         - chrono::DateTime::from_timestamp(bundle_meta.upload_time_epoch as i64, 0).unwrap();
-    assert_eq!(time_since_upload.num_minutes(), 0);
+    assert!(time_since_upload.num_minutes() <= 5);
     assert_eq!(bundle_meta.test_command, None);
     assert!(bundle_meta.os_info.is_some());
     assert!(bundle_meta.quarantined_tests.is_empty());
@@ -172,7 +172,7 @@ async fn upload_bundle() {
     );
     let time_since_junit_modified = chrono::Utc::now()
         - chrono::DateTime::from_timestamp_nanos(bundled_file.last_modified_epoch_ns as i64);
-    assert_eq!(time_since_junit_modified.num_minutes(), 0);
+    assert!(time_since_junit_modified.num_minutes() <= 5);
     assert_eq!(bundled_file.owners, ["@user"]);
     assert_eq!(bundled_file.team, None);
 
--- a/cli-tests/src/upload.rs
+++ b/cli-tests/src/upload.rs
@@ -1,19 +1,38 @@
-use std::fs;
-use std::io::BufReader;
-use std::path::Path;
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -2,6 +2,7 @@
 members = [
   "api",
   "cli",
+  "cli-tests",
```
dfrankland, line 5: It's a fairly common pattern to split out a crate that's just for tests. For example, see [`tokio`](https://github.com/tokio-rs/tokio).

cli-tests/src/upload.rs
```diff
@@ -1,19 +1,38 @@
...
+    static ref CARGO_RUN: CargoRun = CargoBuild::new()
+        .bin("trunk-analytics-cli")
+        .target_dir(CARGO_MANIFEST_DIR.join("../target"))
+        .manifest_path(CARGO_MANIFEST_DIR.join("../cli/Cargo.toml"))
+        .features("force-sentry-env-dev")
+        .current_release()
+        .current_target()
+        .run()
+        .unwrap();
+}
```
dfrankland, line 35: [`escargot`](https://docs.rs/escargot/latest/escargot/) is a nice interface to build crates with features enabled or other things configured. [It's recommended by `assert_cmd`](https://docs.rs/assert_cmd/latest/assert_cmd/cargo/index.html#limitations).

cli/src/main.rs
```diff
@@ -124,13 +124,15 @@ const RETRY_COUNT: usize = 5;
...
-    ));
+    let mut options = sentry::ClientOptions::default();
+    options.release = sentry::release_name!();
+
+    #[cfg(feature = "force-sentry-env-dev")]
+    {
+        options.environment = Some("development".into())
+    }
+
+    let _guard = sentry::init((SENTRY_DSN, options));
```
dfrankland, line 135: `environment` defaults to `"production"` or `"development"` based on the [`debug_assertions`](https://doc.rust-lang.org/reference/conditional-compilation.html#debug_assertions) feature flag. Our new feature `force-sentry-env-dev` forces it to `"development"` for tests.

Cargo.toml
gnalh: That's interesting. How does this typically work with code owners? I assume that both directories will be marked as owned by the same team / group of individuals?

cli-tests/src/upload.rs
```diff
@@ -127,7 +144,7 @@ async fn upload_bundle() {
     assert_eq!(bundle_meta.envs.get("CI"), Some(&String::from("1")));
     let time_since_upload = chrono::Utc::now()
         - chrono::DateTime::from_timestamp(bundle_meta.upload_time_epoch as i64, 0).unwrap();
-    assert_eq!(time_since_upload.num_minutes(), 0);
+    assert!(time_since_upload.num_minutes() <= 5);
```
gnalh, line 147: Nit: This seems to be a fairly popular crate that has an explicit `assert_lt`
https://docs.rs/more-asserts/latest/more_asserts/#

I personally prefer being explicit about expectations in tests since it helps with debugging.

cli-tests/src/upload.rs
```diff
@@ -155,7 +172,7 @@ async fn upload_bundle() {
     );
     let time_since_junit_modified = chrono::Utc::now()
         - chrono::DateTime::from_timestamp_nanos(bundled_file.last_modified_epoch_ns as i64);
-    assert_eq!(time_since_junit_modified.num_minutes(), 0);
+    assert!(time_since_junit_modified.num_minutes() <= 5);
```
gnalh, line 175: Similar comment to the above.

gnalh (approved):

cli-tests/src/upload.rs
```diff
@@ -1,19 +1,38 @@
...
+    types::{BundleMeta, FileSetType},
+};
+
+lazy_static! {
+    static ref CARGO_MANIFEST_DIR: PathBuf = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+    static ref CARGO_RUN: CargoRun = CargoBuild::new()
+        .bin("trunk-analytics-cli")
+        .target_dir(CARGO_MANIFEST_DIR.join("../target"))
+        .manifest_path(CARGO_MANIFEST_DIR.join("../cli/Cargo.toml"))
+        .features("force-sentry-env-dev")
```
max-trunk, line 30: My guess is that every new test file in this crate that wants to invoke the CLI will need to do this? Not that it's a huge lift to copy it into new test files - but is there a Rust-y way to do this once and leverage it across modules?

Also - does enabling the `force-sentry-env-dev` feature in the dev dependencies not take care of this https://github.com/trunk-io/analytics-cli/pull/122/files#diff-48118bb96b547e6c79e43924f9db25f71065d526d458312f2c673458370dd4e3R19?

max-trunk (approved):
LG! couple questions for my Rust learning

Cargo.toml
dfrankland: Forgot to mention that splitting out test crates is necessary when enabling different features for tests.

I don't see people using code owners that often, so I'm not exactly sure. Assigning code owners can be as simple as
```codeowners
cli*/ @cli-team
```
or

```codeowners
cli/ @cli-team
cli-*/ @cli-team
```

cli-tests/src/upload.rs
dfrankland: No problem. I've been trying to keep things minimal until needed, but I have no issue with adding that. Thanks for doing the research 👍

cli-tests/src/upload.rs
dfrankland: If by modules, you mean modules within `cli-tests/src` then we just need to add a `pub` to this `static ref` and reuse it in within the crate.

If you are thinking about how one might reuse this across crates in the workspace, then we could move this to `test_utils` with `pub` and reuse that here and elsewhere.

Enabling `force-sentry-env-dev` for the dependency only enables it for the `lib.rs`, not for the `main.rs`. Here we are using `escargot` to build the `main.rs` binary with the `force-sentry-env-dev` feature enabled.

max-trunk: > If by modules, you mean modules within cli-tests/src then we just need to add a pub to this static ref and reuse it in within the crate.

Yep! this is what I meant. Thanks for the info 🫡

PR #127, opened 2024-10-10
 add test for api_client slow API call logging

dfrankland:
depends on #118

Adds more testing for the `ApiClient`, plus the logging and reporting it performs


test_utils/src/mock_server.rs
```diff
@@ -35,54 +36,106 @@ pub struct MockServerState {
...
+
+impl MockServerBuilder {
+    pub fn new() -> Self {
+        Self {
+            repo_create_handler: post(repo_create_handler),
+            create_bundle_handler: post(create_bundle_handler),
+            get_quarantining_config_handler: post(get_quarantining_config_handler),
+            s3_upload_handler: put(s3_upload_handler),
+        }
+    }
```
dfrankland, line 55: allows setting the route handlers for doing things like failed API calls, slow API calls, etc.

cli/src/api_client/mod.rs
```diff
@@ -238,3 +238,61 @@ fn status_code_help<T: FnMut(&Response) -> String>(
...
+    async fn logs_and_reports_for_slow_api_calls() {
+        let mut mock_server_builder = MockServerBuilder::new();
+        let logs = mock_logger(None);
+        let (events, guard) = mock_sentry();
+
+        async fn slow_s3_upload_handler() -> Response<String> {
+            time::sleep(Duration::from_secs(11)).await;
+            Response::new(String::from("OK"))
+        }
+        mock_server_builder.set_s3_upload_handler(slow_s3_upload_handler);
```
dfrankland, line 263: Now we have some really awesome ability to capture logs, Sentry events, and modify APIs for tests

gnalh (approved):

PR #129, opened 2024-10-15
[TRUNK-12978] v1 validate command

max-trunk:
[TRUNK-12978](https://linear.app/trunk/issue/TRUNK-12978/flaky-test-cli-validate-command)

Adds a first iteration of a `validate` command, intended to allow users onboarding to flaky tests to verify they're producing valid JUnit xml files that we can accurately process, _before_ they configure their CI jobs to start uploading to us.

Users pass a list of `--junit-paths` to validate. For each file found, `validate` will parse and validate using the parser and validator in this repo's `context` crate.

If any fatal parsing errors are encountered, `validate` will exit with a non-zero exit code.

If a file contains `INVALID`-level validation errors, `validate` considers this file to be 'invalid.' If a file contains no `INVALID`-level validation errors, and zero or more `SUBOPTIMAL`-level validation errors, `validate` considers this file to be 'valid.' If any 'invalid' files are encountered, `validate` exits with a non-zero exit code. If all files found are 'valid,' `validate` exits successfully and prints (eventually) a link to return to the onboarding flow of flaky tests.

Example invocation where fatal parsing error is encountered:

Screenshot 2024-10-17 at 10 30 15 AM

Example invocation where some files are invalid:

Screenshot 2024-10-17 at 10 29 23 AM

Example invocation where all files are valid, some with `SUBOPTIMAL` validation errors:

Screenshot 2024-10-17 at 10 29 47 AM

Example invocation where all files are valid, all with no `SUBOPTIMAL` validation errors:

Screenshot 2024-10-17 at 10 29 10 AM

Follow-ups:
 - show line numbers of where validation errors occur
 - add some more context to validation errors to better describe impact on flaky tests processing to users
 - use actual URL for flaky tests onboarding when that is known


cli/src/scanner.rs
```diff
@@ -69,16 +69,21 @@ impl FileSet {
...
+            let original_path_abs = path
+                .to_str()
+                .expect("failed to convert path to string")
+                .to_string();
+            let original_path_rel = path
+                .strip_prefix(repo_root)
+                .unwrap_or(&path)
                 .to_str()
                 .expect("failed to convert path to string")
                 .to_string();
```
dfrankland, line 81: I know you're following the already existing pattern here, but I think we need to update this so it doesn't panic here. Instead of using `expect` we can use `map_err` + `?` (try sigil)

cli/src/types.rs
```diff
@@ -124,6 +125,12 @@ pub struct BundleMeta {
     pub codeowners: Option<CodeOwners>,
 }
 
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct WithFilePath<T> {
+    pub file_path: String,
+    pub wrapped: T,
+}
```
dfrankland, line 132: I think you probably don't need a named struct, but rather just a simple tuple:
https://doc.rust-lang.org/rust-by-example/primitives/tuples.html

cli/src/main.rs
```diff
@@ -511,10 +537,222 @@ async fn run_test(test_args: TestArgs) -> anyhow::Result<i32> {
...
+            reports.extend(junit_parser.into_reports().iter().map(
+                |report| WithFilePath::<Report> {
+                    file_path: bundled_file.original_path_rel.clone(),
+                    wrapped: report.clone(),
+                },
+            ));
+            Ok::<(), anyhow::Error>(())
+        })?;
+        Ok::<(), anyhow::Error>(())
+    })?;
```
dfrankland, line 606: nit: this is more idiomatically a `fold` operation. `for_each` is typically for side-effects

cli/src/main.rs
```diff
@@ -511,10 +537,222 @@ async fn run_test(test_args: TestArgs) -> anyhow::Result<i32> {
     Ok(exit_code)
 }
 
+async fn run_validate(validate_args: ValidateArgs) -> anyhow::Result<i32> {
```
dfrankland, line 540: This file is getting unwieldy in length. Can we split out a file for this command and we'll split out the other commands into files later?

Also, I'd love for you to split out functions for each logical part of the command that you run so that it's easy to understand the flow

max-trunk: Totally

cli/src/scanner.rs
max-trunk: Thanks! went with `ok_or_else` since I was working with an `Option` rather than a `Result`

cli/src/main.rs
max-trunk: Alrighty, lmk if what I've got now is a bit easier on the eyes

cli/src/validate.rs
```diff
@@ -0,0 +1,261 @@
...
+        "Navigate to <URL for next onboarding step> to continue getting started with Flaky Tests"
+    );
+}
+
+fn print_validation_errors(
+    report_validations: &[(JunitReportValidation, String)],
+) -> (usize, usize) {
+    log::info!("");
+    let mut num_invalid_reports: usize = 0;
+    let mut num_optionally_invalid_reports: usize = 0;
```
TylerJang27, line 186: nit: rename to suboptimal for consistency with the underlying types/enums
(and rename elsewhere)

cli-tests/src/validate.rs
```diff
@@ -0,0 +1,44 @@
...
+
+    let assert = Command::new(CARGO_RUN.path())
+        .current_dir(&temp_dir)
+        .args(&["validate", "--junit-paths", "./*"])
+        .assert()
+        .failure();
+
+    println!("{assert}");
+}
+
```
TylerJang27, line 31: nit: would also be nice to have a test for suboptimal junits

TylerJang27 (approved):
LGTM!

cli/src/main.rs
```diff
@@ -511,10 +526,20 @@ async fn run_test(test_args: TestArgs) -> anyhow::Result<i32> {
...
+    } = validate_args;
+    print_cli_start_info();
+    validate(junit_paths, show_warnings)
+}
+
 async fn run(cli: Cli) -> anyhow::Result<i32> {
     match cli.command {
         Commands::Upload(upload_args) => run_upload(upload_args, None, None, None, None).await,
         Commands::Test(test_args) => run_test(test_args).await,
+        Commands::Validate(validate_args) => run_validate(validate_args).await,
```
dfrankland, line 542: Totally optional, but we can get rid of the `run_validate` indirection here:

```suggestion
        Commands::Validate(validate_args) => {
            let ValidateArgs {
                junit_paths,
                show_warnings,
            } = validate_args;
            print_cli_start_info();
            validate(junit_paths, show_warnings).await
        },
```

IMO, this gets us closer to splitting concerns in a way that prevents repeated code and inconsistencies

cli/src/validate.rs
```diff
@@ -0,0 +1,261 @@
...
+    }
+}
+
+fn parse_file_sets(
+    file_sets: Vec<FileSet>,
+) -> anyhow::Result<(Vec<(Report, String)>, Vec<(JunitParseError, String)>)> {
+    file_sets.iter().try_fold(
+        (
+            Vec::<(Report, String)>::new(),          // Vec<(Report, file path)>
+            Vec::<(JunitParseError, String)>::new(), // Vec<(JunitParseError, file path)>
```
dfrankland, line 58: Accidental comments?

cli/src/validate.rs
```diff
@@ -0,0 +1,261 @@
...
+            num_invalid_reports,
+            num_suboptimal_reports,
+        );
+        Ok(EXIT_FAILURE)
+    }
+}
+
+fn parse_file_sets(
+    file_sets: Vec<FileSet>,
+) -> anyhow::Result<(Vec<(Report, String)>, Vec<(JunitParseError, String)>)> {
```
dfrankland, line 54: You can create a type alias for these more complex types and compose them together nicely for readability
https://doc.rust-lang.org/reference/items/type-aliases.html

For example:
```rs
type ParsedReportAndFilePaths = Vec<(Report, String)>;
type JunitParseErrorsAndFilePaths = Vec<(JunitParseError, String)>;
```

which would update your return type to:
```rs
anyhow::Result<(ParsedReportAndFilePaths, JunitParseErrorsAndFilePaths)>
```

Btw, it occurs to me that we could use a `BTreeMap` here though, no? Something like
```rs
type JunitFileToReportAndErrors = BTreeMap<String, (Vec<Report>, Vec<JunitParseError>)>;
```

cli/src/validate.rs
```diff
@@ -0,0 +1,261 @@
...
+                    Ok::<(Vec<(Report, String)>, Vec<(JunitParseError, String)>), anyhow::Error>(
+                        file_set_parse_results,
+                    )
+                },
+            )?;
+            file_sets_parse_results.0.extend(file_set_parse_results.0);
+            file_sets_parse_results.1.extend(file_set_parse_results.1);
+            Ok::<(Vec<(Report, String)>, Vec<(JunitParseError, String)>), anyhow::Error>(
+                file_sets_parse_results,
+            )
```
dfrankland, line 96: If you set a return type of the closure you won't have to annotate the type of `Result` here:

e.g.

```rs
        |mut file_sets_parse_results, file_set| -> anyhow::Result<(Vec<(Report, String)>, Vec<(JunitParseError, String)>)> {
            // ...
            Ok(file_sets_parse_results)
        }
```

cli/src/validate.rs
```diff
@@ -0,0 +1,261 @@
...
+
+    for error in parse_errors {
+        if error.1 != current_file_original_path {
+            current_file_original_path = error.1;
+            log::warn!("  File: {}", current_file_original_path);
+        }
+
+        log::warn!("\t{}", error.0);
+    }
+}
```
dfrankland, line 133: This is so much nicer! Thank you 🙏

cli/src/validate.rs
```diff
@@ -0,0 +1,261 @@
...
+            Ok::<(Vec<(Report, String)>, Vec<(JunitParseError, String)>), anyhow::Error>(
+                file_sets_parse_results,
+            )
+        },
+    )
+}
+
+fn print_matched_files(file_sets: &[FileSet], file_counter: FileSetCounter) {
+    log::info!("");
+    log::info!(
```
dfrankland, line 103: Consider using `\n` for making newlines

dfrankland (approved):

PR #133, opened 2024-10-24
Build wasm package

dfrankland:
depends on #135

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -103,17 +103,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - uses: actions/setup-node@v4
-        with:
-          node-version: 20.x
-
-      - name: Install pnpm
-        uses: trunk-io/trunk-action/install@v1
-        with:
-          tools: pnpm
-
-      - name: Install dependencies
-        run: pnpm install --dir ./context-js
+      - name: Setup and build wasm
+        uses: ./.github/actions/setup_build_wasm
 
       - name: Trunk Check
         uses: trunk-io/trunk-action@v1
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -103,6 +103,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - uses: actions/setup-node@v4
+        with:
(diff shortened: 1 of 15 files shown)


context-js/src/lib.rs
```diff
@@ -37,8 +37,8 @@ pub fn env_parse(env_vars: js_sys::Object) -> Result<Option<env::parser::CIInfo>
 }
 
 #[wasm_bindgen]
-pub fn env_validate(ci_info: env::parser::CIInfo) -> env::validator::EnvValidation {
-    env::validator::validate(&ci_info)
+pub fn env_validate(ci_info: &env::parser::CIInfo) -> env::validator::EnvValidation {
+    env::validator::validate(ci_info)
 }
```
dfrankland, line 42: Turns out that only references work as arguments for wasm-exported functions

context/src/repo/mod.rs
```diff
@@ -186,6 +186,37 @@ impl BundleRepo {
     }
 }
 
+#[cfg(feature = "wasm")]
+#[wasm_bindgen]
+impl BundleRepo {
+    #[wasm_bindgen(constructor)]
+    pub fn js_new(
```
dfrankland, line 193: These constructors are needed to use `new` in JS

.github/workflows/pull_request.yml
```diff
@@ -103,6 +103,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - name: Setup and build wasm
+        uses: ./.github/actions/setup_build_wasm
+
```
TylerJang27, line 108: Sanity check: is the intention here to build the wasm targets and then run eslint on them in CI? Is there a specific pattern/issue that this is designed to catch (other than just making sure they build as part of a PR workflow)?

context-js/tests/parse_and_validate.test.ts
TylerJang27, line 1: 🏆

TylerJang27 (approved):
💯

.github/workflows/pull_request.yml
dfrankland: I'm running lints that take into account types on the test code. The test code relies on types from the wasm build, so I have to build the wasm package prior to linting.

PR #137, opened 2024-10-29
(Feat): Upload bindings to S3

TylerJang27:
Paired with https://github.com/trunk-io/trunk/pull/18229.


.github/workflows/pyo3.yml
```diff
@@ -82,6 +82,14 @@ jobs:
             set -e
             pip3 install context-py --find-links dist --force-reinstall
             cd ./context-py && pytest
+      - name: Upload wasm package to S3
```
dfrankland, line 85: copy-pasta: We need to update the titles of each upload step 😅 

```suggestion
      - name: Upload wheel package to S3
```

dfrankland (approved):

PR #138, opened 2024-10-29
[TRUNK-13061] validate - output report-level issues only once per junit

max-trunk:
[TRUNK-13061](https://linear.app/trunk/issue/TRUNK-13061/validate-cli-command-output-file-level-errors-only-once-per-file)

Extends the junit validator to classify certain validation issues as _report-level_ issues, and modifies the `validate` command to print these only once per junit file

Screenshot 2024-10-29 at 10 56 02 AM

Note that the list of issues that we _want_ to classify as report-level are still TBD and will be declared in this sheet https://docs.google.com/spreadsheets/d/18gFzo88H4FzBKaAWpjXYv3hJhlNHwHW8lKV0j2SfOiA/edit?gid=0#gid=0. I've migrated two issues to be report-level to start:

  1. `report has test cases with missing file or filepath`
  2. `report has stale (> 1 hour(s)) timestamps`

as these were discussed in the original [slack thread](https://trunk-io.slack.com/archives/C07L4B34WUE/p1729094430525819) that spurred this initiative


TylerJang27 (approved):
Changes LGTM but would appreciate another set of more rust-adept eyes

context/src/junit/validator.rs
```diff
@@ -114,9 +115,9 @@ pub fn validate(report: &Report) -> JunitReportValidation {
                     .unwrap_or_default(),
             ) {
                 FieldLen::Valid => (),
-                FieldLen::TooShort(s) => {
-                    test_case_validation.add_issue(JunitValidationIssue::SubOptimal(
-                        JunitTestCaseValidationIssueSubOptimal::TestCaseFileOrFilepathTooShort(s),
+                FieldLen::TooShort(_s) => {
```
dfrankland, line 118: optional: we can just use `..` to denote that we don't care what we're matching on—or just use `_`

```suggestion
                FieldLen::TooShort(..) => {
```

context/src/junit/validator.rs
```diff
@@ -169,8 +170,8 @@ pub fn validate(report: &Report) -> JunitReportValidation {
                         JunitTestCaseValidationIssueSubOptimal::TestCaseOldTimestamp(timestamp),
                     ));
                 } else if time_since_timestamp.num_hours() > i64::from(TIMESTAMP_STALE_HOURS) {
-                    test_case_validation.add_issue(JunitValidationIssue::SubOptimal(
-                        JunitTestCaseValidationIssueSubOptimal::TestCaseStaleTimestamp(timestamp),
+                    report_validation.add_issue(JunitValidationIssue::SubOptimal(
+                        JunitReportValidationIssueSubOptimal::StaleTimestamps,
                     ));
```
dfrankland, line 175: I'm not sure it makes sense to change collecting `test_case_validation` to `report_validation`. This will make it confusing if later on we want to inspect issues with test cases. Perhaps we can either add logic to derive report validation issues from all child issues? Or we can add to both the test case and report validation, then add logic that dedupes the total issues?

context/src/junit/validator.rs
```diff
@@ -348,7 +427,7 @@ impl JunitTestSuiteValidation {
     pub fn max_level(&self) -> JunitValidationLevel {
         self.test_cases
             .iter()
-            .map(|test_suite| test_suite.level)
+            .map(|test_case| test_case.level)
```
dfrankland, line 430: Thank you! 🙏

cli/src/validate.rs
```diff
@@ -203,27 +203,41 @@ fn print_validation_errors(
...
+            + num_test_case_validation_warnings;
 
-        let num_validation_errors_str = if num_invalid_validation_errors > 0 {
-            num_invalid_validation_errors.to_string().red()
+        let num_validation_errors_str = if num_validation_errors > 0 {
+            num_validation_errors.to_string().red()
         } else {
-            num_invalid_validation_errors.to_string().green()
+            num_validation_errors.to_string().green()
         };
```
dfrankland, line 241: Considering this and the additional logic in the follow up PR, it might be a good idea to move some of this logic into the `context` crate since it can be used elsewhere to get quick totals for validation issues.

context/src/junit/validator.rs
max-trunk: > This will make it confusing if later on we want to inspect issues with test cases

Fair point!

> Perhaps we can either add logic to derive report validation issues from all child issues?

Going to head down this path

max-trunk: Alrighty - reverted to collecting all individual test case / test suite issues as before, then added a `derive_all_issues()` step afterwards to derive a list of 'all issues to print,' if you will, that includes the report-level issues and handles the deduping. Doing this also eliminated the need for lots of redundant helper functions, subsequently making things cleaner in `cli/src/validate.rs` as well. Lemme know how that looks to you!

cli/src/validate.rs
max-trunk: Done 👍 should look a lot nicer now

dfrankland (approved):
very cool, brother 😎

PR #139, opened 2024-10-29
[TRUNK-13055] Treat unparsable XML as a validation error

max-trunk:
Part of [TRUNK-13055](https://linear.app/trunk/issue/TRUNK-13055/validate-cli-command-refine-error-messages)

Previously, `validate` would short-circuit and exit unsuccessfully upon encountering an unreadable / unparsable xml file

Screenshot 2024-10-29 at 11 12 27 AM

Now, `validate` treats this scenario as a validation error and continues parsing / validating the remaining files, as per https://docs.google.com/spreadsheets/d/18gFzo88H4FzBKaAWpjXYv3hJhlNHwHW8lKV0j2SfOiA/edit?gid=0#gid=0

Screenshot 2024-10-29 at 10 57 25 AM

Also refactors some of the code in `/cli/src/validate.rs` to be a bit cleaner (imo)


cli-tests/src/utils.rs
```diff
@@ -63,3 +65,10 @@ pub fn generate_mock_codeowners<T: AsRef<Path>>(directory: T) {
     "#;
     fs::write(directory.as_ref().join("CODEOWNERS"), CODEOWNERS).unwrap();
 }
+
+pub fn write_junit_xml_to_dir<T: AsRef<Path>>(xml: &str, directory: T) -> Result<()> {
+    let path = directory.as_ref().join("junit-0.xml");
+    let mut file = fs::File::create(&path)?;
+    file.write_all(xml.as_bytes())?;
+    Ok(())
+}
```
dfrankland, line 74: No need to use `anyhow` or return a result here. `.unwrap()` is fine in tests since we want hard failures anyways

cli-tests/src/utils.rs
```diff
@@ -63,3 +65,10 @@ pub fn generate_mock_codeowners<T: AsRef<Path>>(directory: T) {
     "#;
     fs::write(directory.as_ref().join("CODEOWNERS"), CODEOWNERS).unwrap();
 }
+
+pub fn write_junit_xml_to_dir<T: AsRef<Path>>(xml: &str, directory: T) -> Result<()> {
+    let path = directory.as_ref().join("junit-0.xml");
+    let mut file = fs::File::create(&path)?;
+    file.write_all(xml.as_bytes())?;
```
dfrankland, line 72: nit, optional: You can make it easier to pass strings by using a generic

```suggestion
pub fn write_junit_xml_to_dir<T: AsRef<str>, U: AsRef<Path>>(xml: T, directory: U) -> Result<()> {
    let path = directory.as_ref().join("junit-0.xml");
    let mut file = fs::File::create(&path)?;
    file.write_all(xml.as_ref().as_bytes())?;
```

cli/src/validate.rs
```diff
@@ -62,44 +61,44 @@ pub async fn validate(junit_paths: Vec<String>, show_warnings: bool) -> anyhow::
...
+            match junit_parser.parse(file_buf_reader) {
+                Err(e) => {
+                    parse_results.insert(
+                        bundled_file.original_path_rel.clone(),
+                        (Err(anyhow::anyhow!(e)), Vec::new()),
+                    );
+                    return parse_results;
+                }
+                _ => (),
+            };
```
dfrankland, line 91: nit: it's better to use `if let` for cases like this where we don't care about matching other variants

```suggestion
            if let Err(e) = junit_parser.parse(file_buf_reader) {
                parse_results.insert(
                    bundled_file.original_path_rel.clone(),
                    (Err(anyhow::anyhow!(e)), Vec::new()),
                );
                return parse_results;
            }
```

cli/src/validate.rs
```diff
@@ -249,30 +260,41 @@ fn print_validation_errors(
...
+        match report_parse_error {
+            Some(parse_error) => {
+                log::info!(
+                    "  {} - {}",
+                    print_validation_level(JunitValidationLevel::Invalid),
+                    parse_error,
+                );
+            }
+            _ => (),
+        }
```
dfrankland, line 279: same here, use `if let`

PR #142, opened 2024-10-30
(Feat): Generate python stubs in uploaded wheels

TylerJang27:
Use `py3o_stub_gen` to generate stubs/types for python wheels. The stubs right now are not as comprehensive as I would like, but I'll add more as a follow-up/as needed.

Tested that I was successfully able to retrieve these locally on Mac and Linux (and will be testing in CI shortly).


.trunk/trunk.yaml
```diff
@@ -40,7 +40,15 @@ lint:
         - name: lint
           run_when: [cli, monitor, ci]
 actions:
+  definitions:
+    - id: generate-pyi
+      description: Generate Python stubs
+      # Don't use hermetic runtime so we can use nightly cargo
+      run: |
+        cargo run --bin stub_gen
+        mv context-py/context{-,_}py.pyi
```
TylerJang27, line 49: Waffled on just calling these directly in CI or requiring folks to autogen these on `pre-commit`, but I landed on the current approach with a centralized Trunk Action to call

.github/workflows/pyo3.yml
```diff
@@ -87,9 +92,11 @@ jobs:
         if: "!cancelled() && github.event_name != 'pull_request'"
         with:
           bucket: ${{ secrets.S3_ASSET_BUCKET }}
-          name: wheels-linux-${{ matrix.platform.target }}.tar.gz
+          name: linux-${{ matrix.platform.target }}
           path: dist
+          s3-flags: --recursive
```
TylerJang27, line 97: We need to upload all the wheels in order to easily install them

context-py/py.typed
TylerJang27, line 1: Required for [maturing](https://www.maturin.rs/project_layout#adding-python-type-information) to pick up the typings

dfrankland (approved):
THIS IS SO COOL. NICE WORK PARDNER 🤠

PR #146, opened 2024-10-31
fix parsing junits without testsuites element

dfrankland:
depends on #144

Some junit reporters don't include a top-level `testsuites` element. This ensures we support parsing that.


max-trunk (approved):
lgtm!

PR #147, opened 2024-10-31
(Fix): Improve compatibility of python bindings

TylerJang27:
Adds features [abi3-py39](https://pyo3.rs/v0.14.5/features.html#abi3-py36--abi3-py37--abi3-py38--abi3-py39), to support more python versions, and [extension-module](https://pyo3.rs/v0.14.5/features.html#extension-module), to skip linking against `libpython.so`. `extension-module` can only be applied on Linux. Also fixes the version to be compatible on Glue (see [openssl note](https://github.com/openssl/openssl/issues/25366)).

Works in our CI setup


dfrankland (approved):
send it

PR #150, opened 2024-11-06
Append org_url_slug and repo_full_name to id generation

gnalh:
The IDs should be globally unique and there's a small chance of collision with the current approach if the generated id url from xcresult matches across repos. Next steps are to move these pieces server side to prevent these types of issues in the future.


xcresult/src/xcresult.rs
```diff
@@ -66,13 +69,19 @@ fn xcresulttool<T: AsRef<str>>(
 }
 
 impl XCResult {
-    pub fn new<T: AsRef<str>>(path: T) -> anyhow::Result<XCResult> {
+    pub fn new<T: AsRef<str>>(
+        path: T,
+        repo_url_parts: &RepoUrlParts,
+        org_url_slug: T,
```
dfrankland, line 75: nit: reusing `T` here will necessitate that both `path` and `org_url_slug`. Conventionally, those will likely always be the same in this repo, but it's recommended to allow them to both be independent

```suggestion
    pub fn new<T: AsRef<str>, U: AsRef<str>>(
        path: T,
        repo_url_parts: &RepoUrlParts,
        org_url_slug: U,
```

dfrankland (approved):
rad 🤙

PR #151, opened 2024-11-06
Add clear log message for empty xcresult

gnalh:
Add a clearer error message for when we are unable to locate tests in the xcresult path.


dfrankland (approved):
Awesome 👍

PR #153, opened 2024-11-07
[TRUNK-13277] Attempt to resolve correct commit for GH checkout action

max-trunk:
[TRUNK-13277](https://linear.app/trunk/issue/TRUNK-13277/qa-bug-unclear-what-commit-corresponds-to-the-pr-test-summary)

Attempt to resolve the correct commit for GH actions.

The GH checkout action checks out a new merge commit (e.g. https://github.com/trunk-io/analytics-cli/actions/runs/11714618861/job/32629607945#step:2:53) that doesn't end up in the actual git history. We're using this merge commit as the commit to store for sha, epoch, message and author. This ends up being confusing for users when they go to trace back the commit from the FT web app to their git repo only to find that the commit doesn't exist.

With changes in this PR, commit info is accurate in the bundle metadata:

```
cat meta.json | jq | grep repo_head_sha
    "repo_head_sha": "312a7b10dce98f82e8f4eba78b59a671842342b9",
    "repo_head_sha_short": "312a7b1",
cat meta.json | jq | grep repo_head_commit_message
    "repo_head_commit_message": "attempt to resolve correct commit for gh actions\n",
cat meta.json | jq | grep repo_head_author_name
    "repo_head_author_name": "Max Cruz",
```

And correct commit sha is shown in uploads view:

Screenshot 2024-11-07 at 9 38 36 AM

### DIFF
--- a/context/src/repo/mod.rs
+++ b/context/src/repo/mod.rs
@@ -1,5 +1,3 @@
-use std::{path::PathBuf, process::Command};
-
 use anyhow::Context;
 #[cfg(feature = "pyo3")]
 use pyo3::prelude::*;
@@ -7,6 +5,7 @@ use pyo3::prelude::*;
 use pyo3_stub_gen::derive::gen_stub_pyclass;
 use regex::Regex;
 use serde::{Deserialize, Serialize};
+use std::{path::PathBuf, process::Command};
 #[cfg(feature = "wasm")]
 use wasm_bindgen::prelude::*;
 
--- a/context/src/repo/mod.rs
+++ b/context/src/repo/mod.rs
@@ -1,4 +1,4 @@
-use std::path::PathBuf;
+use std::{path::PathBuf, process::Command};
 
 use anyhow::Context;
 #[cfg(feature = "pyo3")]
@@ -74,7 +74,7 @@ impl BundleRepo {
                 bundle_repo_options.repo_url = bundle_repo_options.repo_url.or_else(|| {
                     git_repo
                         .config_snapshot()
-                        .string_by_key(GIT_REMOTE_ORIGIN_URL_CONFIG)
+                        .string(GIT_REMOTE_ORIGIN_URL_CONFIG)
                         .map(|s| s.to_string())
                 });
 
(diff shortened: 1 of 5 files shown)


context/src/repo/mod.rs
```diff
@@ -155,6 +163,53 @@ impl BundleRepo {
...
+        current_commit: gix::Commit<'a>,
+        repo_head_branch: String,
+    ) -> gix::Commit<'a> {
+        // for GH actions, grab sha of PR head, not the PR merge commit
+        let re = Regex::new(r"refs\/remotes\/pull\/[0-9]+\/merge").unwrap();
+        if re.is_match(&repo_head_branch) && current_commit.parent_ids().count() == 2 {
+            log::debug!("attempting to resolve PR head commit");
+
+            let branch_to_fetch = repo_head_branch.replace("remotes/", "");
+            match Command::new("git")
```
pv72895, line 179: I believe `gix` has a fetch command we can use instead of running a command - https://docs.rs/gix/latest/gix/clone/fetch/index.html (this is probably linking the wrong thing lol)

max-trunk: I tried extensively to implement this with `gix` but was unsuccessful every time. I think it's implementation of `fetch` right now is half-baked, or only used internally for it's `clone` module

context/src/repo/mod.rs
```diff
@@ -1,12 +1,12 @@
...
-
 use anyhow::Context;
 #[cfg(feature = "pyo3")]
 use pyo3::prelude::*;
 #[cfg(feature = "pyo3")]
 use pyo3_stub_gen::derive::gen_stub_pyclass;
 use regex::Regex;
 use serde::{Deserialize, Serialize};
+#[allow(unused_imports)]
+use std::{path::PathBuf, process::Command};
```
dfrankland, line 9: nit: you need to use `#[cfg(feature = "git-access")]` on these

context/src/repo/mod.rs
```diff
@@ -157,6 +165,76 @@ impl BundleRepo {
...
     }
+
+    #[cfg(feature = "git-access")]
+    fn resolve_repo_head_commit<'a>(
+        git_repo: &'a gix::Repository,
+        current_commit: gix::Commit<'a>,
+        repo_head_branch: String,
+    ) -> gix::Commit<'a> {
+        // for GH actions, grab PR branch HEAD commit, not the PR merge commit
+        let re = Regex::new(r"refs\/remotes\/pull\/[0-9]+\/merge").unwrap();
```
dfrankland, line 176: nit: regexs should be created in a `lazy_static` so we don't try to compile them on every invocation:
https://docs.rs/regex/latest/regex/#avoid-re-compiling-regexes-especially-in-a-loop

context/src/repo/mod.rs
dfrankland: unrelated to this PR: we should probably switch to [`git2`](https://crates.io/crates/git2) if `gix` isn't cutting the mustard

dfrankland (approved):

PR #156, opened 2024-11-14
fix: require non empty string value in value parser

tenesttang-trunk:
Require non empty string value for parsing --junit-paths

### DIFF
--- a/cli-tests/src/upload.rs
+++ b/cli-tests/src/upload.rs
@@ -10,6 +10,7 @@ use api::{
 use assert_cmd::Command;
 use assert_matches::assert_matches;
 use context::repo::RepoUrlParts as Repo;
+use predicates::prelude::*;
 use tempfile::tempdir;
 use test_utils::mock_server::{MockServerBuilder, RequestPayload};
 use trunk_analytics_cli::{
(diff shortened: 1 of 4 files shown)


pv72895 (changes requested):
We have tests on the CLI invocations - can you update those as well? Holding off on approving until then

https://github.com/trunk-io/analytics-cli/blob/d94c2ee9c2ea22a208cb131d157fff9b98070f92/cli-tests/src/upload.rs

https://github.com/trunk-io/analytics-cli/blob/a0ff6550727e91f91396196590265d398e658df8/cli-tests/src/validate.rs

And this will teach you how to write tests!

cli-tests/src/upload.rs
```diff
@@ -205,6 +206,37 @@ async fn upload_bundle_no_files() {
     println!("{assert}");
 }
 
+#[tokio::test(flavor = "multi_thread")]
```
pv72895, line 209: Do you know why this is necessary / used, besides the fact that the other tests do it here? I don't either, but I wanted to ask

pv72895 (approved):

cli-tests/src/upload.rs
tenesttang-trunk: I don't know, I am cargo culting 😱

PR #166, opened 2024-11-19
(Fix): Parse GH branch and PR number correctly

TylerJang27:
Updates to match our other parsers. I will be manually auditing the rest of the env parsing as well to verify parity

Also adds a fix for parsing [Semaphore branch](https://docs.semaphoreci.com/reference/env-vars)


TylerJang27:
There's more than I thought to sync up. Will be pushing more to this branch

tenesttang-trunk (approved):
Great catch, thank you!

PR #174, opened 2024-11-22
[TRUNK-13579] patch junit parser

max-trunk:
Part of [TRUNK-13579](https://linear.app/trunk/issue/TRUNK-13579/move-junit-parsing-out-of-glue-and-into-dagster)

Reconciles some things with respect to junit parsing encountered when using the bindings for python in the ETL.

Summary of changes:
 - Reports either the error returned by `JunitParser::parse()` or any of the errors collected in `JunitParser::errors()` as an error to the caller in the `junit_parse()` pyfunction. Previously would only report the error on `parse()` and ignore any errors collected in `errors()`
 - Exports some additional classes in the `context_py` module
 - Lots more `context_py` junit parsing tests
 - Adds a `timestamp_micros` field to the `BindingsX` classes for timestamps w/ microseconds
 - Fixes an issue in `JunitDateParser::parse_date()` where the cached `self.date_type` was set incorrectly, causing a panic when alternating between parsing datetimes and parsing naive dates. Probably an unlikely scenario? but encountered it in unit testing
 - Modifies the junit parser to ignore nested `<testsuite>`s. Verified that this is what the current [python junit parser](https://github.com/weiwei/junitparser) does
 - Modifies the junit parser to only set `<testcase>` status once. There's a comment in the parser code stating we expect a test case to only have one [status](https://github.com/trunk-io/analytics-cli/blob/25ce229d1e2d8247a6fe93b48740b3e12cf8f93b/context/src/junit/parser.rs#L178), but there's an example junit in the monorepo that has a test case with [two](https://github.com/trunk-io/trunk/blob/c57c36a91c23656328856bee350ce6812b803d73/trunk/py/metrics_data_pipeline/trunk_etl_junit_parser/test_data/basic_junit.xml#L5-L13) 
 - Modifies the junit parser to not treat missing report name as a parsing error. This is parity with the python parser. Doesn't look like we're using report name anywhere downstream in the ETL? And we'd be throwing away a lot of junits if we required it

### DIFF
--- a/context/src/junit/bindings.rs
+++ b/context/src/junit/bindings.rs
@@ -11,6 +11,8 @@ use quick_junit::{
 #[cfg(feature = "wasm")]
 use wasm_bindgen::prelude::*;
 
+const MICROSECONDS_PER_SECOND: i64 = 1_000_000;
+
 #[cfg_attr(feature = "pyo3", gen_stub_pyclass, pyclass(get_all))]
 #[cfg_attr(feature = "wasm", wasm_bindgen(getter_with_clone))]
 #[derive(Clone, Debug)]
@@ -18,6 +20,7 @@ pub struct BindingsReport {
     pub name: String,
     pub uuid: Option<String>,
     pub timestamp: Option<i64>,
+    pub timestamp_micros: Option<i64>,
     pub time: Option<f64>,
     pub tests: usize,
     pub failures: usize,
@@ -42,6 +45,7 @@ impl From<Report> for BindingsReport {
             name: name.into_string(),
             uuid: uuid.map(|u| u.to_string()),
             timestamp: timestamp.map(|t| t.timestamp()),
+            timestamp_micros: timestamp.map(|t| t.timestamp_micros()),
             time: time.map(|t| t.as_secs_f64()),
             tests,
             failures,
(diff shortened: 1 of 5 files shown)


context/src/junit/bindings.rs
```diff
@@ -11,13 +11,16 @@ use quick_junit::{
...
+const MICROSECONDS_PER_SECOND: i64 = 1_000_000;
+
 #[cfg_attr(feature = "pyo3", gen_stub_pyclass, pyclass(get_all))]
 #[cfg_attr(feature = "wasm", wasm_bindgen(getter_with_clone))]
 #[derive(Clone, Debug)]
 pub struct BindingsReport {
     pub name: String,
     pub uuid: Option<String>,
     pub timestamp: Option<i64>,
+    pub timestamp_micros: Option<i64>,
```
dfrankland, line 23: Do we really need micro second precision?

dfrankland (approved):

context/src/junit/bindings.rs
max-trunk: Really need, not sure. But it's what our python date parser was giving us and was just going for parity to start

TylerJang27 (approved):

PR #176, opened 2024-11-25
parse_meta_from_tarball python bindings

dfrankland:
Depends on https://github.com/trunk-io/analytics-cli/pull/181

Will be used to filter bundles without completely downloading them


context-py/tests/test_parse_and_validate.py
```diff
@@ -276,3 +276,113 @@ def test_repo_validate():
...
+            tar.add(meta_file_path, "meta.json")
+            for i in range(1000):
+                tar.add(meta_file_path, f"junk{str(i)}")
+        with open(meta_tarball_path, "rb") as f:
+            encoder = zstd.ZstdCompressor(level=6)
+            meta_tarball_compressed = encoder.compress(f.read())
+
+    raw_stream = StreamingBody(
+        io.BytesIO(meta_tarball_compressed), len(meta_tarball_compressed)
+    )
```
dfrankland, line 376: `StreamingBody` is the type returned by S3 when fetching an object

context-py/src/py_bytes_read.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    pub fn new(py_bytes_reader: Bound<'py, PyAny>) -> PyResult<Self> {
+        let content_length = py_bytes_reader
+            .getattr("_content_length")?
+            .extract::<usize>()?;
+        Ok(Self {
+            inner: py_bytes_reader,
+            content_length,
+            content_length_read: 0,
+            inner_buffer: Vec::with_capacity(0),
```
dfrankland, line 28: fun fact: there's no allocation when a `Vec` is created with `0` capacity 👍

context-py/src/py_bytes_read.rs
```diff
@@ -0,0 +1,76 @@
...
+
+pub struct PyBytesReader<'py> {
+    inner: Bound<'py, PyAny>,
+    content_length: usize,
+    content_length_read: usize,
+    inner_buffer: Vec<u8>,
+}
+
+impl<'py> PyBytesReader<'py> {
+    const DEFAULT_CHUNK_SIZE: usize = 1024;
```
dfrankland, line 18: This matches `StreamingBody`'s default chunk size

context-py/src/lib.rs
```diff
@@ -75,6 +80,22 @@ fn repo_validate(bundle_repo: repo::BundleRepo) -> repo::validator::RepoValidati
...
+    py: Python<'_>,
+    reader: PyObject,
+) -> PyResult<BindingsVersionedBundle> {
+    let py_bytes_reader = PyBytesReader::new(reader.into_bound(py))?;
+    let rt = tokio::runtime::Builder::new_current_thread()
+        .enable_all()
+        .build()?;
+    let versioned_bundle = rt
+        .block_on(parse_tarball(py_bytes_reader))
+        .map_err(|err| PyTypeError::new_err(err.to_string()))?;
```
dfrankland, line 95: Even though this is not async, reading bytes still happens as needed

bundle/src/bundle_meta.rs
```diff
@@ -83,5 +93,35 @@ pub enum VersionedBundle {
...
+            _ => None,
+        }
+    }
+    pub fn get_v0_6_2(&self) -> Option<BundleMetaV0_6_2> {
+        match &self.0 {
+            VersionedBundle::V0_6_2(bundle_meta) => Some(bundle_meta.clone()),
+            _ => None,
+        }
+    }
+}
```
dfrankland, line 124: Using `gen_stub_pymethods`doesn't seem to work on enums yet:
https://github.com/Jij-Inc/pyo3-stub-gen/issues/69

All that's needed is a new type here to get around it

Also, the Python bindings aren't using `serde`, so `VersionedBundle` doesn't end up creating a nice `dict`—we can just use methods to get the `meta.json` values

context-py/tests/test_parse_and_validate.py
```diff
@@ -276,3 +276,113 @@ def test_repo_validate():
     assert repo_validation.max_level() == RepoValidationLevel.Valid, "\n" + "\n".join(
         [issue.error_message for issue in repo_validation.issues_flat()]
     )
+
+
+def test_parse_meta_from_tarball():
```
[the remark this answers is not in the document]
dfrankland: Can do no problemlo

PR #182, opened 2024-11-26
chore: check and warn if codeowners present but junit missing file paths

tenesttang-trunk:
If CODEOWNERS is present but jUnit is missing file paths, print a suboptimal message.

### DIFF
--- a/cli/src/validate.rs
+++ b/cli/src/validate.rs
@@ -10,7 +10,7 @@ use context::junit::{
     parser::{JunitParseError, JunitParser},
     validator::{
         validate as validate_report, JunitReportValidation, JunitReportValidationFlatIssue,
-        JunitValidationLevel,
+        JunitReportValidationIssueSubOptimal, JunitValidationLevel,
     },
 };
 
@@ -70,7 +70,7 @@ pub async fn validate(
 
     let codeowners = CodeOwners::find_file(&current_dir, &codeowners_path);
 
-    print_codeowners(codeowners);
+    print_codeowners_validation(codeowners, &report_validations);
 
     Ok(exit)
 }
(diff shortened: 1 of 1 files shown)


cli/src/validate.rs
```diff
@@ -291,6 +294,23 @@ fn print_codeowners(codeowners: Option<CodeOwners>) {
                 print_validation_level(JunitValidationLevel::Valid)
             );
             println!("    Path: {:?}", owners.path);
+            for report_validation in report_validations {
+                match report_validation.1 {
+                    Ok(validation) => {
+                        let issues = validation.all_issues_owned();
+                        if issues.iter().any(
+                            |i|
+                            i.error_message == JunitReportValidationIssueSubOptimal::TestCasesFileOrFilepathMissing.to_string()
```
dfrankland, line 303: Converting enums to strings and doing a comparison is a smell—typically it means we aren't managing our types well. In this case it seems like `all_issues`/`all_issues_owned` is serializing information when it should only do that when needed. That's outside the scope of your change here, but I decided I'd help out with a follow-up PR 😃

cli/src/validate.rs
```diff
@@ -291,6 +294,23 @@ fn print_codeowners(codeowners: Option<CodeOwners>) {
...
+                        ) {
+                            println!(
+                                "    {} - CODEOWNERS found but test cases are missing filepaths. We will not be able to correlate flaky tests with owners.",
+                                print_validation_level(JunitValidationLevel::SubOptimal)
+                            );
```
dfrankland, line 308: nit: Although there should only ever be 1 report, this looks like it might get printed once per report—as mentioned above we can use iterators to flatten things a bit more and ensure we only print this once

dfrankland (commented):
There's a bit too much to go over in this PR, so I made a follow-up PR that we should go over together and get an understanding of my comments:
https://github.com/trunk-io/analytics-cli/pull/183

dfrankland:
Oh! Also, I forgot to ask for a test if we can as well 👍

dfrankland (approved):
Followed up offline. We'll merge a follow-up PR afterwards

PR #184, opened 2024-11-27
bundle version downgrading

dfrankland:
Allows for easily getting the lowest denominator of a bundle version

### DIFF
--- a/bundle/src/bundle_meta.rs
+++ b/bundle/src/bundle_meta.rs
@@ -46,6 +46,14 @@ pub struct BundleMetaV0_5_29 {
     pub base_props: BundleMetaBaseProps,
 }
 
+impl From<BundleMetaV0_5_34> for BundleMetaV0_5_29 {
+    fn from(bundle_meta: BundleMetaV0_5_34) -> Self {
+        BundleMetaV0_5_29 {
+            base_props: bundle_meta.base_props,
+        }
+    }
+}
+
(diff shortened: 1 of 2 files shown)


bundle/src/bundle_meta.rs
```diff
@@ -105,12 +122,20 @@ pub struct BindingsVersionedBundle(pub VersionedBundle);
...
     pub fn get_v0_5_29(&self) -> Option<BundleMetaV0_5_29> {
         match &self.0 {
+            VersionedBundle::V0_6_2(bundle_meta) => Some(BundleMetaV0_5_29::from(
+                BundleMetaV0_5_34::from(bundle_meta.clone()),
+            )),
+            VersionedBundle::V0_5_34(bundle_meta) => {
+                Some(BundleMetaV0_5_29::from(bundle_meta.clone()))
+            }
             VersionedBundle::V0_5_29(bundle_meta) => Some(bundle_meta.clone()),
-            _ => None,
```
tenesttang-trunk, line 109: I'm confused why this line was deleted

tenesttang-trunk (approved):

bundle/src/bundle_meta.rs
dfrankland: It's deleted because nothing will match it. `VersionedBundle` only has 3 variants and each of them are matched here 😃

PR #186, opened 2024-11-27
chore: add end to end test to exercise missing filepath + codeowners present

tenesttang-trunk:
Add end to end test to exercise the case when junit tests are missing filepaths but there is a codeowners file present.


dfrankland (approved):
Speedy! Looks great :)

PR #187, opened 2024-11-28
fix streaming body types

dfrankland:


context-py/requirements-dev.txt
```diff
@@ -17,6 +17,6 @@ pytest==8.3.3
 # TEST DEPENDENCIES
 # -------
 
-botocore==1.35.69
-botocore-stubs==1.35.69
+botocore==1.34.162
+botocore-stubs==1.34.162
```
dfrankland, line 21: matches the version in the monorepo

dfrankland: Totally, I was just considering this is a public repo that we shouldn't document too much of our monorepo

PR #188, opened 2024-12-02
[TRUNK-13580] JUnit validator changes for ETL usage

max-trunk:
Part of [TRUNK-13580](https://linear.app/trunk/issue/TRUNK-13580/move-junit-validation-out-of-glue-and-into-dagster)

Tweaks to the junit validator to prepare for usage in the ETL

Summary of changes:
 - Extends `validate()` to collect all of the 'valid' (suboptimal issues are ok) test suites & test cases in its resulting `JunitReportValidation`. This allows the ETL to get a quick list of the successfully-parsed test suites & test cases it's allowed to process
 - Introduces a new `BindingsJunitReportValidation` type. Since `JunitReportValidation` now includes a vector of `quick_junit::TestSuite` as the 'valid' test suites, pyo3 can no longer directly create a class from it
 - Changes the `junit_validate` pyfunction to return the new `BindingsJunitReportValidation` type
 - Fixes an issue where converting from `BindingsJunitReport` to `quick_junit::Report` would result in test cases with an incorrect number of microseconds in their timestamps

### DIFF
--- a/context-py/src/lib.rs
+++ b/context-py/src/lib.rs
@@ -70,8 +70,32 @@ fn junit_parse(xml: Vec<u8>) -> PyResult<Vec<junit::bindings::BindingsReport>> {
 #[pyfunction]
 fn junit_validate(
     report: junit::bindings::BindingsReport,
-) -> junit::validator::JunitReportValidation {
-    junit::validator::validate(&report.into())
+) -> junit::bindings::BindingsJunitReportValidation {
+    junit::bindings::BindingsJunitReportValidation::from(junit::validator::validate(&report.into()))
+}
+
+#[gen_stub_pyfunction]
+#[pyfunction]
+fn junit_validation_level_to_string(
+    junit_validation_level: junit::validator::JunitValidationLevel,
+) -> String {
+    match junit_validation_level {
+        junit::validator::JunitValidationLevel::Valid => "VALID".to_string(),
+        junit::validator::JunitValidationLevel::SubOptimal => "SUBOPTIMAL".to_string(),
+        junit::validator::JunitValidationLevel::Invalid => "INVALID".to_string(),
+    }
+}
+
+#[gen_stub_pyfunction]
+#[pyfunction]
+fn junit_validation_type_to_string(
+    junit_validation_type: junit::validator::JunitValidationType,
+) -> String {
+    match junit_validation_type {
+        junit::validator::JunitValidationType::Report => "Report".to_string(),
+        junit::validator::JunitValidationType::TestSuite => "TestSuite".to_string(),
+        junit::validator::JunitValidationType::TestCase => "TestCase".to_string(),
+    }
 }
 
 #[gen_stub_pyfunction]
@@ -109,10 +133,14 @@ fn context_py(m: &Bound<'_, PyModule>) -> PyResult<()> {
     m.add_class::<junit::bindings::BindingsTestCase>()?;
     m.add_class::<junit::bindings::BindingsTestCaseStatusStatus>()?;
     m.add_class::<junit::bindings::BindingsNonSuccessKind>()?;
+    m.add_class::<junit::bindings::BindingsJunitReportValidation>()?;
+    m.add_class::<junit::validator::JunitReportValidationFlatIssue>()?;
     m.add_class::<junit::validator::JunitValidationLevel>()?;
     m.add_class::<junit::validator::JunitValidationType>()?;
     m.add_function(wrap_pyfunction!(junit_parse, m)?)?;
     m.add_function(wrap_pyfunction!(junit_validate, m)?)?;
+    m.add_function(wrap_pyfunction!(junit_validation_level_to_string, m)?)?;
+    m.add_function(wrap_pyfunction!(junit_validation_type_to_string, m)?)?;
 
     m.add_class::<repo::BundleRepo>()?;
     m.add_class::<repo::RepoUrlParts>()?;
--- a/context/src/junit/bindings.rs
+++ b/context/src/junit/bindings.rs
@@ -84,7 +84,7 @@ impl Into<Report> for BindingsReport {
                     let micros_delta = TimeDelta::microseconds(micro_secs);
                     DateTime::from_timestamp(
                         micros_delta.num_seconds(),
-                        micros_delta.num_nanoseconds().unwrap_or_default() as u32,
+                        micros_delta.subsec_nanos() as u32,
                     )
                 })
                 .map(|dt| dt.fixed_offset()),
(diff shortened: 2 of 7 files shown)


context-py/src/lib.rs
```diff
@@ -70,8 +70,32 @@ fn junit_parse(xml: Vec<u8>) -> PyResult<Vec<junit::bindings::BindingsReport>> {
...
+#[gen_stub_pyfunction]
+#[pyfunction]
+fn junit_validation_type_to_string(
+    junit_validation_type: junit::validator::JunitValidationType,
+) -> String {
+    match junit_validation_type {
+        junit::validator::JunitValidationType::Report => "Report".to_string(),
+        junit::validator::JunitValidationType::TestSuite => "TestSuite".to_string(),
+        junit::validator::JunitValidationType::TestCase => "TestCase".to_string(),
+    }
```
max-trunk, line 98: Can't get stubs for pymethods on enums yet https://github.com/Jij-Inc/pyo3-stub-gen/issues/69, figured these functions would be the path of least resistance

context/src/junit/validator.rs
```diff
@@ -86,14 +86,15 @@ pub fn validate(report: &Report) -> JunitReportValidation {
         if let Some(raw_test_suite_id) = test_suite.extra.get("id") {
             let test_case_id = uuid::Uuid::parse_str(raw_test_suite_id).unwrap_or_default();
             if test_case_id.get_version() != Some(uuid::Version::Sha1) {
-                test_suite_validation.add_issue(JunitValidationIssue::Invalid(
-                    JunitTestSuiteValidationIssueInvalid::TestSuiteInvalidId(
+                test_suite_validation.add_issue(JunitValidationIssue::SubOptimal(
+                    JunitTestSuiteValidationIssueSubOptimal::TestSuiteInvalidId(
```
max-trunk, line 90: Ran this by Gabe - the current ETL validation logic ignores the `id` attribute on test suites, so calling non-uuid IDs invalid here would cause us to start throwing away some test suites (there are some junits have something like `<testsuite id="0" ... >`). De-escalating to suboptimal for now so we at least continue to process them, even though it looks like we're not using test suite ID anywhere downstream in the ETL

context-py/src/lib.rs
dfrankland: I recommend to use a new type pattern to wrap the enum and expose the functions that way. I actually ran into the same issue here:
https://github.com/trunk-io/analytics-cli/pull/176/files#r1857903898

You can see this new type pattern here:
https://github.com/trunk-io/analytics-cli/blob/c21215e6210d2a33cf15c7ed0ec74c5e0f03fcc9/bundle/src/bundle_meta.rs#L113-L147

context/src/junit/bindings.rs
```diff
@@ -80,7 +86,8 @@ impl Into<Report> for BindingsReport {
                 .and_then(|micro_secs| {
                     DateTime::from_timestamp(
                         micro_secs / MICROSECONDS_PER_SECOND,
-                        (micro_secs % MICROSECONDS_PER_SECOND) as u32,
+                        ((micro_secs % MICROSECONDS_PER_SECOND) * NANOSECONDS_PER_MICROSECOND)
+                            as u32,
```
dfrankland, line 90: Rather than doing the math ourselves, how about using `chrono` to do the work for us?
https://docs.rs/chrono/latest/chrono/struct.TimeDelta.html

That way we don't run into human error, and we don't have to write tests for our own time conversions

context/src/junit/validator.rs
```diff
@@ -197,9 +198,19 @@ pub fn validate(report: &Report) -> JunitReportValidation {
...
+            }
+
             test_suite_validation.test_cases.push(test_case_validation);
         }
 
+        if test_suite_validation.level != JunitValidationLevel::Invalid {
+            let mut valid_test_suite = test_suite.clone();
+            valid_test_suite.test_cases = valid_test_cases;
+            report_validation.valid_test_suites.push(valid_test_suite);
+        }
```
dfrankland, line 212: Do we need to clone these or can we derive them via a method? I think it might be better to do this as a method so as to not take more memory

context/src/junit/bindings.rs
```diff
@@ -655,3 +663,71 @@ impl Into<NonSuccessKind> for BindingsNonSuccessKind {
...
+            .count()
+    }
+
+    pub fn num_suboptimal_issues(&self) -> usize {
+        self.all_issues
+            .iter()
+            .filter(|issue| issue.level == JunitValidationLevel::SubOptimal)
+            .count()
+    }
+}
```
dfrankland, line 733: This seems to be related to the comment from the PR description:
> Introduces a new BindingsJunitReportValidation type. Since JunitReportValidation now includes a vector of quick_junit::TestSuite as the 'valid' test suites, pyo3 can no longer directly create a class from it

I'm not sure that's true? That's only if you are implementing `pyclass(get_all)`, not `pyclass`. Do we need to implement `pyclass(get_all)` or can we just implement `pymethods` for the things that Python needs access to?

context/src/junit/bindings.rs
max-trunk: Totally - thanks for the tip

context/src/junit/validator.rs
max-trunk: I think that since `validate()` here is working with a reference to a `Report` and not storing any of the test suites / cases nor any identifiers in its iteration, I'm not seeing how we'd be able to derive the valid suites / cases later in a method. Is there a different viable path that you're thinking of?

context/src/junit/bindings.rs
max-trunk: The root of the issue is that I'd like to return the actual `quick_junit::TestSuite` structs as the 'valid' test suites, and I don't see a way to do that without first converting them to our `BindingsTestSuite` structs first. Is there a way to do that conversion as a `pymethod` on `JunitReportValidation`?

context-py/src/lib.rs
max-trunk: I took a look at this, but I'd need to convert all of the structs that use the current `JunitValidationType` and `JunitValidationLevel` enums to use the new wrapped types. I figured that'd be more work than adding these functions here as a stopgap. Def open to pushback tho!

context/src/junit/bindings.rs
dfrankland: I think I see your point.

The nicest way we can try to do that is by newtype wrapping `quick_junit::TestSuite` and having it `Deref` to its inner `quick_junit::TestSuite`. That way we can use Python methods on it and also use it within our Rust code. I now that would take more work, but it's a pattern that is more scalable, so we should consider if we find ourselves needing more of this same pattern.

Another possibility would be to have multiple `impl`s with some methods specifcally for Python to use. An issue with that, is that currently `pymethod` cannot be used multiple times, so you could not use some of the Rust `impl` with the Python `impl`.

context/src/junit/validator.rs
dfrankland: Gotcha. The other way I'd think about it is storing the indexes of these rather than the values themselves. That way we can reference them from the `Report` later on.

dfrankland (approved):
Thanks for answering some of my questions!

context-py/src/lib.rs
dfrankland: This is fine as-is. I think if we have to support a bunch, probably using newtype pattern will be more scalable

PR #189, opened 2024-12-02
[ruby] Set up initial bindings

gnalh:
Boilerplate for the bindings for ruby + `parse_env` as an example.


.github/workflows/ruby.yml
```diff
@@ -0,0 +1,30 @@
...
+    tags:
+      - "*"
+  pull_request: {}
+  workflow_dispatch: {}
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
```
gnalh, line 16: Only building for right now. Need to figure out publishing still.

Cargo.toml
```diff
@@ -7,6 +7,7 @@ members = [
   "context",
   "context-js",
   "context-py",
+  "context-ruby/ext/context_ruby",
```
gnalh, line 10: This is how magnus recommends structuring the bindings.

.github/workflows/pull_request.yml
```diff
@@ -106,6 +106,11 @@ jobs:
       - name: Setup Rust & Cargo
         uses: ./.github/actions/setup_rust_cargo
 
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: "3.1"
+          working-directory: context-ruby
+
```
gnalh, line 113: Waiting for custom image from our self-hosted runners for ruby to work on self-hosted. This means that the trunk check step will fail until updated.

.github/workflows/ruby.yml
```diff
@@ -0,0 +1,32 @@
...
+  build:
+    runs-on: ubuntu-latest
```
gnalh, line 18: Will move this to the custom image once available.

context/src/env/parser.rs
```diff
@@ -434,3 +533,29 @@ impl<'a> EnvParser<'a> {
...
+
+#[cfg(feature = "ruby")]
+pub fn ruby_init(ruby: &magnus::Ruby) -> Result<(), magnus::Error> {
+    let ci_platform = ruby.define_class("CIPlatform", ruby.class_object())?;
+    ci_platform.define_method("to_s", magnus::method!(CIPlatform::to_string, 0))?;
+    let ci_info = ruby.define_class("CIInfo", ruby.class_object())?;
+    ci_info.define_singleton_method("new", magnus::function!(CIInfo::new, 1))?;
+    ci_info.define_method("platform", magnus::method!(CIInfo::platform, 0))?;
+    ci_info.define_method("job_url", magnus::method!(CIInfo::job_url, 0))?;
+    ci_info.define_method("branch", magnus::method!(CIInfo::branch, 0))?;
```
gnalh, line 545: There is a `define_attr` method of the defined class, but it doesn't work like you'd expect. It makes it so that ruby can read / write to an object constructed in run time, but does not allow you to access the field if it was constructed in rust. Explicit getters and setters seem to be needed.

context-ruby/ext/context_ruby/src/lib.rs
```diff
@@ -0,0 +1,41 @@
...
+
+pub fn env_validate(ci_info: &env::parser::CIInfo) -> env::validator::EnvValidation {
+    env::validator::validate(ci_info)
+}
+
+pub fn repo_validate(bundle_repo: repo::BundleRepo) -> repo::validator::RepoValidation {
+    repo::validator::validate(&bundle_repo)
+}
+
+#[magnus::init]
```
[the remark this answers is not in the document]
dfrankland: Ohhh, woah—I see exactly what you mean. Yeah, maybe we can make this a bit easier to use a global mutable static variable. I think that's what `wasm-bindgen` and `pyo3` are doing under the hood

context/src/env/parser.rs
```diff
@@ -64,6 +67,49 @@ pub enum CIPlatform {
...
+            CIPlatform::TravisCI => "TRAVIS_CI",
+            CIPlatform::Webappio => "WEBAPPIO",
+            CIPlatform::AWSCodeBuild => "AWS_CODEBUILD",
+            CIPlatform::BitbucketPipelines => "BITBUCKET",
+            CIPlatform::AzurePipelines => "AZURE_PIPELINES",
+            CIPlatform::GitLabCI => "GITLAB_CI",
+            CIPlatform::Drone => "DRONE",
+        }
+    }
+}
```
dfrankland, line 88: nit: We should use `ci_platform_env_key` here.

Optional: We could use this outside of Ruby too and there's a nice couple of traits that we can implement for it as well: `Into<&str>` and `ToString`

For example:
```suggestion
impl Into<&str> for CIPlatform {
    fn into(self) -> &'static str {
        match self {
            CIPlatform::GitHubActions => ci_platform_env_key::GITHUB_ACTIONS,
            CIPlatform::JenkinsPipeline => ci_platform_env_key::JENKINS_PIPELINE,
            CIPlatform::CircleCI => ci_platform_env_key::CIRCLECI,
            CIPlatform::Buildkite => ci_platform_env_key::BUILDKITE,
            CIPlatform::Semaphore => ci_platform_env_key::SEMAPHORE,
            CIPlatform::TravisCI => ci_platform_env_key::TRAVIS_CI,
            CIPlatform::Webappio => ci_platform_env_key::WEBAPPIO,
            CIPlatform::AWSCodeBuild => ci_platform_env_key::AWS_CODEBUILD,
            CIPlatform::BitbucketPipelines => ci_platform_env_key::BITBUCKET,
            CIPlatform::AzurePipelines => ci_platform_env_key::AZURE_PIPELINES,
            CIPlatform::GitLabCI => ci_platform_env_key::GITLAB_CI,
            CIPlatform::Drone => ci_platform_env_key::DRONE,
        }
    }
}

impl ToString for CIPlatform {
    fn to_string(&self) -> String {
        String::from(Into::<&str>::into(*self))
    }
}

#[cfg(feature = "ruby")]
impl CIPlatform {
    pub fn to_string(&self) -> &str {
        (*self).into()
    }
}
```

context/src/env/parser.rs
```diff
@@ -64,6 +67,49 @@ pub enum CIPlatform {
...
+            4 => Ok(CIPlatform::Semaphore),
+            5 => Ok(CIPlatform::TravisCI),
+            6 => Ok(CIPlatform::Webappio),
+            7 => Ok(CIPlatform::AWSCodeBuild),
+            8 => Ok(CIPlatform::BitbucketPipelines),
+            9 => Ok(CIPlatform::AzurePipelines),
+            10 => Ok(CIPlatform::GitLabCI),
+            11 => Ok(CIPlatform::Drone),
+            // TODO
+            _ => Ok(CIPlatform::GitHubActions),
```
dfrankland, line 108: Shouldn't this return `Err(magnus::Error)`?

context/src/env/parser.rs
```diff
@@ -392,6 +443,54 @@ impl CIInfo {
     }
 }
 
+// unfortunately magnus requires explicit getters and setters for the object
+// to be accessible from Ruby
```
dfrankland, line 447: Probably doesn't need to be a code comment since I'm assuming we'll be doing this all over the place. I agree that this is a lot. There must be a nice proc macro someone has written to help do this the same way that `pyo3` or `wasm-bindgen` have done. At worse, we can try to implement our own in the future.

dfrankland (approved):

context/src/env/parser.rs
gnalh: Nope, you can't construct Magnus errors without a ruby object. It should return an error, but it doesn't work with anyhow out of the box. The TODO here was to return an error, but I forget to add that in. Will add it.

context/src/env/parser.rs
gnalh: wasn't sure if we'd want this to be general available. Will update.

gnalh: Just checked and these aren't quire the same. `ci_platform_env_key::JENKINS_PIPELINE` is `BUILD_ID` and not `JENKINS_PIPELINE`. I'm good making this change since it's just an example so far, but it won't necessarily match expectations upfront.

context/src/env/parser.rs
gnalh: Nevermind, figured out how to get the Magnus error populated. You use `magnus::Ruby` directly and not the one populated in `Magnus::init`

PR #191, opened 2024-12-03
Tenest/trunk 13755 codeowners parse from bytestream

tenesttang-trunk:
Add function to parse codeowners from a byte vector. Add a basic py test. Will expand on test after more work on py bindings.

### DIFF
--- a/codeowners/src/codeowners.rs
+++ b/codeowners/src/codeowners.rs
@@ -1,7 +1,6 @@
 use std::{
     ffi::OsStr,
     fs::File,
-    io::BufRead,
     path::{Path, PathBuf},
 };
 
@@ -71,13 +70,12 @@ impl CodeOwners {
         })
     }
 
-    pub fn parse<R: BufRead>(&mut self, codeowners: R) -> Self
-    where
-        R: Read,
-    {
-        let owners_result = GitHubOwners::from_reader(&codeowners)
+    pub fn parse(codeowners: Vec<u8>) -> Self {
+        let owners_result = GitHubOwners::from_reader(codeowners.as_slice())
             .map(Owners::GitHubOwners)
-            .or_else(|_| GitLabOwners::from_reader(&codeowners).map(Owners::GitLabOwners));
+            .or_else(|_| {
+                GitLabOwners::from_reader(codeowners.as_slice()).map(Owners::GitLabOwners)
+            });
 
         if let Err(ref err) = owners_result {
             log::error!(
--- a/codeowners/src/codeowners.rs
+++ b/codeowners/src/codeowners.rs
@@ -1,5 +1,7 @@
 use std::{
+    ffi::OsStr,
     fs::File,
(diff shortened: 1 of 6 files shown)


codeowners/src/codeowners.rs
```diff
@@ -68,6 +69,27 @@ impl CodeOwners {
...
+                "Found CODEOWNERS file `{}` in bundle, but couldn't parse it",
+                err
+            );
+        }
+
+        let owners = Result::ok(owners_result);
+        Self {
+            path: PathBuf::from(OsStr::new("")),
+            owners,
+        }
```
dfrankland, line 91: No need to use `PathBuf::from` here. We don't need to bother with `OsStr`, also we can simplify this a little bit

```suggestion
        Self {
            path: PathBuf::new(),
            owners: owners_result.ok(),
        }
```

codeowners/src/codeowners.rs
```diff
@@ -68,6 +69,27 @@ impl CodeOwners {
...
+            .or_else(|_| {
+                GitLabOwners::from_reader(codeowners.as_slice()).map(Owners::GitLabOwners)
+            });
+
+        if let Err(ref err) = owners_result {
+            log::error!(
+                "Found CODEOWNERS file `{}` in bundle, but couldn't parse it",
+                err
+            );
+        }
```
dfrankland, line 85: This isn't part of the CLI, so I'd remove this. This will just log into the void.

context-py/src/lib.rs
```diff
@@ -96,6 +97,14 @@ pub fn parse_meta_from_tarball(
     Ok(BindingsVersionedBundle(versioned_bundle))
 }
 
+#[gen_stub_pyfunction]
+#[pyfunction]
+fn codeowners_parse(codeowners_bytes: Vec<u8>) -> PyResult<CodeOwners> {
+    let codeowners = CodeOwners::parse(codeowners_bytes);
+
+    Ok(codeowners)
+}
```
dfrankland, line 106: No need for `PyResult` since nothing can fail here

```suggestion
fn codeowners_parse(codeowners_bytes: Vec<u8>) -> CodeOwners {
    CodeOwners::parse(codeowners_bytes)
}
```

context-py/tests/test_parse_compressed_codeowners.py
```diff
@@ -0,0 +1,7 @@
+def test_parse_meta_from_tarball():
+    from context_py import codeowners_parse
+
+    codeowners_text = """* @trunk/test"""
+    codeowners = codeowners_parse(str.encode(codeowners_text))
```
dfrankland, line 5: No need to encode the string here, Python has a byte-string notation

```suggestion
    codeowners_text = b"""* @trunk/test"""
    codeowners = codeowners_parse(codeowners_text)
```

dfrankland (approved):

PR #192, opened 2024-12-03
[TRUNK-13708] Split quarantine from upload

cmillar-trunk:
Separates the quarantine command out from the upload command. Leaves use_quarantining as an option for the test command.

### DIFF
--- a/cli/src/upload.rs
+++ b/cli/src/upload.rs
@@ -6,9 +6,10 @@ use std::{
     io::BufReader,
     time::{SystemTime, UNIX_EPOCH},
 };
-#[cfg(target_os = "macos")]
-use xcresult::XCResult;
 
+use crate::junit_utils::junit_require;
+#[cfg(target_os = "macos")]
+use crate::junit_utils::junitify_xcresult;
 use api::BundleUploadStatus;
 use bundle::{
     parse_custom_tags, BundleMeta, BundleMetaBaseProps, BundleMetaDebugProps, BundleMetaJunitProps,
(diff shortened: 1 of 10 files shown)


cli/src/upload.rs
```diff
@@ -245,7 +208,6 @@ pub async fn run_upload(
             envs,
             upload_time_epoch: SystemTime::now().duration_since(UNIX_EPOCH)?.as_secs(),
             test_command,
-            quarantined_tests: resolved_quarantine_results.quarantine_results.to_vec(),
```
max-trunk, line 248: If we remove the `quarantined_tests` from the upload metadata here, I don't know if the ETL will continue to work properly with respect to quarantining - at least in its current state. Right now, the ETL expects `quarantined_tests` in the metadata [(1)](https://github.com/trunk-io/trunk/blob/c6bf42c9115ce1a818cc64d1a640513fc2302bf3/trunk/py/metrics_data_pipeline/trunk_etl_junit_parser/xml_parser.py#L63-L66) and uses it to set each test's `quarantine_result` [(2)](https://github.com/trunk-io/trunk/blob/c6bf42c9115ce1a818cc64d1a640513fc2302bf3/trunk/py/metrics_data_pipeline/trunk_etl_junit_parser/xml_parser.py#L391-L397). Unless I'm misunderstanding (which is very possible)

cmillar-trunk: Ah, so it'd be better for us to still support the `use_quarantining` flag in the upload command? I'll switch over to that.

dfrankland: tl;dr: We plan to eventually remove this, but we cannot remove this yet.

I was just thinking about this yesterday. I spoke with @federicoweber and I think we're going to move to a new endpoint that's used on the `quarantine` command only. @federicoweber is working on that endpoint now, and it should be done today, but that won't be ready to store and retrieve data for the ETL. We'll have to coordinate a bit to solve this appropriately.

cmillar-trunk:
Cancelled based on conversations that we don't want to support this due to issues around incompatible states (ie you do an upload, then break stuff, then quarantine - quarantining is attached to an upload, so where does that quarantine go?).

PR #194, opened 2024-12-04
add sha check to repo validator

katchao:


context/tests/repo.rs
```diff
@@ -1,9 +1,10 @@
 use chrono::DateTime;
 use context::repo::{
     self,
-    validator::{RepoValidationIssue, RepoValidationLevel},
+    validator::{RepoValidationIssue, RepoValidationLevel, MAX_SHA_FIELD_LEN},
     BundleRepo, RepoUrlParts,
 };
+use gix::odb::pack::bundle;
```
katchao, line 7: This diff somehow got added in by itself at some point? Can someone to sanity check that this should be removed before I remove it, ty!

katchao: K nvm I just removed it and it works fine

context/src/repo/validator.rs
```diff
@@ -85,6 +86,10 @@ pub enum RepoValidationIssueSubOptimal {
 pub enum RepoValidationIssueInvalid {
     #[error("repo branch name too short")]
     RepoBranchNameTooShort(String),
+    #[error("repo sha too short")]
+    RepoShaTooShort(String),
+    #[error("repo sha too long, truncated to {}", MAX_SHA_FIELD_LEN)]
+    RepoShaTooLong(String),
```
max-trunk, line 92: Note that by saying these errors are 'invalid,' we're saying that, if they're present, we cannot process the bundle for flaky tests and will skip the bundle in processing. Is that the case? Or can we still process the bundle? If we can still process the bundle I'd suggest making these errors 'suboptimal-level' (`RepoValidationIssueSubOptimal`) meaning we'll still track them in the defects table and flag them to the end user, but we can still process the data

max-trunk (commented):
Code lgtm! just one question

context/src/repo/validator.rs
pv72895: Right now, we are planning on dropping uploads without a git state, so that makes not having a git state an `INVALID` error - https://trunk-io.slack.com/archives/C044VBASZ0D/p1733183476781069?thread_ts=1733182342.660259&cid=C044VBASZ0D

max-trunk: @katchao - Chatted with @pv72895; sounds like we want to keep `RepoShaTooShort` as an `INVALID` error due to our db model's dependency on its presence, but switch `RepoShaTooLong` to a `SUBOPTIMAL` error

pv72895, line 92: I think `RepoShaTooLong` should probably be a suboptimal error, especially if we're doing something like truncating it - things like `RepoAuthorNameTooLong` also are suboptimal errors. SHAs are mainly used as keys (meaning test results belonging to the same SHA have a high impact on whether or not a test is called flaky), so we care more about if they're there and less about their length (we care about them being too long I guess for data constraints??)

pv72895 (approved):

PR #197, opened 2024-12-05
(Feat): Support passing a bazel BuildEventProtocol path

TylerJang27:
Support testing, validating, and uploading from a bazel [BuildEventProtocol](https://bazel.build/remote/bep) JSON file

Screenshot 2024-12-05 at 1 32 08 PM

I've added unit and integration tests, and also validated this manually with a local bazel run and a successful upload.

As a follow-up I'll be vendoring the proto crate in this repo


cli/src/main.rs
```diff
@@ -42,12 +42,19 @@ struct TestArgs {
 struct ValidateArgs {
     #[arg(
         long,
-        required = true,
+        required_unless_present = "bazel_bep_path",
+        conflicts_with = "bazel_bep_path",
```
TylerJang27, line 46: This allows it to be mutually exclusive with the `--junit-paths` as well as `--xc-result-path` when applicable

cli/src/main.rs
```diff
@@ -42,12 +42,19 @@ struct TestArgs {
...
         value_parser = clap::builder::NonEmptyStringValueParser::new(),
         help = "Comma-separated list of glob paths to junit files."
     )]
     junit_paths: Vec<String>,
+    #[arg(
+        long,
+        required_unless_present = "junit_paths",
+        help = "Path to bazel build event protocol JSON file."
+    )]
+    bazel_bep_path: Option<String>,
```
TylerJang27, line 57: If we decide we want to support the binary format too at some point, we can just name that `bazel_bep_binary_path`. JSON seems the sanest default here

cli/src/main.rs
```diff
@@ -121,21 +129,26 @@ async fn run_test(test_args: TestArgs) -> anyhow::Result<i32> {
...
+    };
 
     log::info!("running command: {:?}", command);
     let run_result = run_test_command(
         &repo,
         &org_url_slug,
         command.first().unwrap(),
         command.iter().skip(1).collect(),
-        junit_paths,
+        &junit_spec,
```
TylerJang27, line 151: We can't early parse here like we do in other cases, since we haven't run the test yet 🙃

cli/src/main.rs
```diff
@@ -200,11 +213,23 @@ async fn run(cli: Cli) -> anyhow::Result<i32> {
...
+                bazel_bep_path,
                 show_warnings,
                 codeowners_path,
             } = validate_args;
+
+            let junit_file_paths = match bazel_bep_path {
+                Some(bazel_bep_path) => {
+                    let mut parser = BazelBepParser::new(bazel_bep_path.clone());
+                    parser.parse()?;
+                    parser.xml_files()
```
TylerJang27, line 225: Not sure if there's a more idiomatic way to do this in rust. Maybe chaining?

cli/src/upload.rs
```diff
@@ -258,6 +273,7 @@ pub async fn run_upload(
 
     log::info!("Total files pack and upload: {}", file_counter.get_count());
     if file_counter.get_count() == 0 {
+        // DONOTLAND FIX LOG MESSAGE HERE
```
TylerJang27, line 276: Will be auditing this to make sure the message still makes sense/works

context/src/bazel_bep/parser.rs
```diff
@@ -0,0 +1,99 @@
...
+                self.errors
+                    .push(format!("Error parsing build event: {}", err));
+                return;
+            }
+            if let Some(build_event) = parse_event.ok() {
+                if let Some(Payload::TestResult(test_result)) = build_event.payload {
+                    let xml_files = test_result.test_action_output.into_iter().fold(
+                        Vec::new(),
+                        |mut xml_files, action_output| {
+                            if action_output.name.ends_with(".xml") {
```
TylerJang27, line 64: `test.xml` by [default](https://docs.trunk.io/flaky-tests/get-started/frameworks/bazel) afaik, but my understanding is that this may vary for different runners

cli/src/main.rs
```diff
@@ -121,21 +129,26 @@ async fn run_test(test_args: TestArgs) -> anyhow::Result<i32> {
         repo_head_commit_epoch.clone(),
     )?;
 
-    if junit_paths.is_empty() {
-        return Err(anyhow::anyhow!("No junit paths provided."));
+    if junit_paths.is_empty() && bazel_bep_path.is_none() {
+        return Err(anyhow::anyhow!("No junit or bazel BEP paths provided."));
```
pv72895, line 133: I think I'd rather us keep the old error message until we document this feature (even if it'll appear in `--help`)

context/src/bazel_bep/parser.rs
```diff
@@ -0,0 +1,99 @@
+use anyhow::Ok;
+use bazel_bep::types::build_event_stream::{build_event::Payload, file::File::Uri, BuildEvent};
+use serde_json::Deserializer;
+
+#[derive(Debug, Clone)]
+pub struct TestResult {
+    pub cached: bool,
+    pub xml_files: Vec<String>,
+}
```
pv72895, line 9: I was under the impression that one test result = one file - is that not the case? Or just being proactive here?

pv72895 (commented):
My pre-review - all looks good to me at a first passthroug. Since we're ok with this feature being hacky, I'm ok with believing the external types you're using to be correct / up-to-date enough

context/src/bazel_bep/parser.rs
TylerJang27: From what I've seen it's just one xml file. But there are multiple files (test.log), and conceivably if you were using bazel to wrap some other invocation, you could get multiple junits, so being proactive

context/src/bazel_bep/parser.rs
```diff
@@ -0,0 +1,155 @@
...
+        });
+
+        if !&self.errors.is_empty() {
+            log::warn!("Errors parsing BEP file: {:?}", &self.errors);
+        }
+
+        log::info!(
+            "Parsed {} ({} cached) test results from BEP file",
+            self.test_results.len(),
+            self.test_results.iter().filter(|r| r.cached).count()
```
TylerJang27, line 95: Open to input on precomputing this, storing other info, etc. I wanted to strike a balance of only storing info that we found valuable here, while also reporting useful information for debug purposes...

context/test_fixtures/bep_example
TylerJang27, line 1: These are sanitized BEP files from our repro of the original issue. I have:
- Stripped env vars
- Simplified paths
- Removed extra log output
- Cleaned up CLI options

context/src/bazel_bep/parser.rs
```diff
@@ -0,0 +1,155 @@
...
+                                    xml_files.push(
+                                        file.strip_prefix(FILE_URI_PREFIX)
+                                            .unwrap_or(&file)
+                                            .to_string(),
+                                    );
+                                }
+                            }
+                            xml_files
+                        },
+                    );
```
dfrankland, line 75: nit: this is better done with `filter_map`

for example:

```rust
let xml_files = test_result
    .test_action_output
    .into_iter()
    .filter_map(|action_output| {
        if action_output.name.ends_with(".xml") {
            action_output.file.and_then(|f| {
                if let Uri(uri) = f {
                    Some(
                        uri.strip_prefix(FILE_URI_PREFIX)
                            .unwrap_or(&uri)
                            .to_string(),
                    )
                } else {
                    None
                }
            })
        } else {
            None
        }
    })
    .collect();
```

context/src/bazel_bep/parser.rs
```diff
@@ -0,0 +1,155 @@
...
+            .flatten()
+            .collect()
+    }
+
+    pub fn parse(&mut self) -> anyhow::Result<()> {
+        let file = std::fs::File::open(&self.bazel_bep_path)?;
+        let reader = std::io::BufReader::new(file);
+
+        let build_events = Deserializer::from_reader(reader).into_iter::<BuildEvent>();
+        build_events.for_each(|parse_event| {
```
dfrankland, line 53: `for_each` is for side-effects, I'd recommend using a normal `for` loop or a `fold` here.

For example:

```rust
let (errors, test_results) = Deserializer::from_reader(reader)
    .into_iter::<BuildEvent>()
    .fold(
        (Vec::<String>::new(), Vec::<TestResult>::new()),
        |(mut errors, mut test_results), parse_event| {
            match parse_event {
                Result::Err(ref err) => {
                    errors.push(format!("Error parsing build event: {}", err));
                }
                Result::Ok(build_event) => {
                    if let Some(Payload::TestResult(test_result)) = build_event.payload {
                        let xml_files = test_result
                            .test_action_output
                            .into_iter()
                            .filter_map(|action_output| {
                                if action_output.name.ends_with(".xml") {
                                    action_output.file.and_then(|f| {
                                        if let Uri(uri) = f {
                                            Some(
                                                uri.strip_prefix(FILE_URI_PREFIX)
                                                    .unwrap_or(&uri)
                                                    .to_string(),
                                            )
                                        } else {
                                            None
                                        }
                                    })
                                } else {
                                    None
                                }
                            })
                            .collect();

                        let cached =
                            if let Some(execution_info) = test_result.execution_info {
                                execution_info.cached_remotely || test_result.cached_locally
                            } else {
                                test_result.cached_locally
                            };

                        test_results.push(TestResult { cached, xml_files });
                    }
                }
            }

            (errors, test_results)
        },
    );

self.errors = errors;
self.test_results = test_results;
```

context/src/bazel_bep/parser.rs
```diff
@@ -0,0 +1,155 @@
...
+                    } else {
+                        test_result.cached_locally
+                    };
+
+                    self.test_results.push(TestResult { cached, xml_files });
+                }
+            }
+        });
+
+        if !&self.errors.is_empty() {
```
dfrankland, line 88: No need for the reference here
```suggestion
        if !self.errors.is_empty() {
```

context/src/bazel_bep/parser.rs
```diff
@@ -0,0 +1,155 @@
...
+            log::warn!("Errors parsing BEP file: {:?}", &self.errors);
+        }
+
+        log::info!(
+            "Parsed {} ({} cached) test results from BEP file",
+            self.test_results.len(),
+            self.test_results.iter().filter(|r| r.cached).count()
+        );
```
dfrankland, line 96: We shouldn't log in this crate since it's reusable. If this gets used via bindings it will log into the void.

cli-tests/Cargo.toml
```diff
@@ -7,6 +7,8 @@ edition = "2021"
 api = { path = "../api" }
 assert_cmd = "2.0.16"
 assert_matches = "1.5.0"
+# Fork from 0.2.2 that adds serde serialize/deserialize via pbjson
+bazel-bep = { git = "https://github.com/TylerJang27/bazel-bep.git", rev = "e51c546960067b9fe98ae35ae00bc53302973a9e" }
```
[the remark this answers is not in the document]
TylerJang27: Good point, will do as a follow-up

.trunk/trunk.yaml
TylerJang27, line 1: These changes didn't end up helping with CI like I'd hoped, but it may prove helpful locally. If need be we can remove later

.github/workflows/pyo3.yml
```diff
@@ -52,12 +53,22 @@ jobs:
...
+            fi
+
+            if [ "$(uname -m)" == "aarch64" ]; then
+              curl -L https://github.com/protocolbuffers/protobuf/releases/download/v29.1/protoc-29.1-linux-aarch_64.zip -o protoc.zip
+              unzip protoc.zip
+              export PROTOC=$(pwd)/bin/protoc
+            else
+              curl -L https://github.com/protocolbuffers/protobuf/releases/download/v29.1/protoc-29.1-linux-x86_64.zip -o protoc.zip
+              unzip protoc.zip
+              export PROTOC=$(pwd)/bin/protoc
```
TylerJang27, line 71: The package registries were giving me issues. This ended up being a more reliable way to install and pin within the VMs

.github/actions/setup_rust_cargo/action.yaml
```diff
@@ -9,6 +9,16 @@ runs:
...
+    - name: Install protoc
+      shell: bash
+      run: |
+        if [ -f /opt/homebrew/bin/brew ]; then
+          /opt/homebrew/bin/brew install protobuf
+        elif [ -f /usr/bin/apt ]; then
+          sudo /usr/bin/apt update
+          sudo /usr/bin/apt install -y protobuf-compiler
+        fi
+
```
TylerJang27, line 21: The arduino setup action was relentless with rate limiting (not sure why they don't allow bypassing the API call), and I didn't want to pass our whole GH token rn https://github.com/marketplace/actions/setup-protoc

cli/src/runner.rs
```diff
@@ -8,24 +8,45 @@ use bundle::{
...
     let start = SystemTime::now();
     let exit_code = run_test_and_get_exit_code(command, args).await?;
     log::info!("Command exit code: {}", exit_code);
-    let (file_sets, ..) =
-        build_filesets(&repo.repo_root, output_paths, team, codeowners, Some(start))?;
+
+    let output_paths = match junit_spec {
+        JunitSpec::Paths(paths) => paths,
+        JunitSpec::BazelBep(bep_path) => {
+            let mut parser = BazelBepParser::new(bep_path.clone());
```
dfrankland, line 36: Now that `junit_spec` is passed by value, we don't need to clone here

context/src/bazel_bep/parser.rs
```diff
@@ -0,0 +1,186 @@
...
+                }
+                (test_count, cached_count)
+            },
+        );
+        log::info!(
+            "Parsed {} ({} cached) test results from BEP file",
+            test_count,
+            cached_count
+        );
+    }
```
dfrankland, line 68: Optional: I would leave the `log::info` for the CLI create. This method could just return the number of tests and number of cached tests to be used however the caller pleases

dfrankland (approved):

context/src/bazel_bep/parser.rs
TylerJang27: Ah, fair point. Will handle in follow-up

PR #199, opened 2024-12-05
Classify merge branches in environment parser

jc3m:


context/src/env/parser.rs
```diff
@@ -416,19 +420,22 @@ pub enum BranchClass {
     Merge,
 }
 
-impl TryFrom<(&str, Option<usize>)> for BranchClass {
+impl TryFrom<(&str, Option<usize>, Option<String>)> for BranchClass {
     type Error = CIInfoParseError;
 
-    fn try_from(value: (&str, Option<usize>)) -> Result<Self, Self::Error> {
-        let (branch_name, pr_number) = value;
+    fn try_from(value: (&str, Option<usize>, Option<String>)) -> Result<Self, Self::Error> {
```
dfrankland, line 426: nit: `Option<String>` feels a little funny since this should only be 1 of `detached`, `merged_result`, or `merge_train`. It might be a bit better for us to "parse" this into an enum and use that here instead.

context/src/env/parser.rs
```diff
@@ -416,19 +420,22 @@ pub enum BranchClass {
...
             return Ok(BranchClass::PullRequest);
         }
         if branch_name.starts_with("remotes/pull/") || branch_name.starts_with("pull/") {
             Ok(BranchClass::PullRequest)
         } else if matches!(branch_name, "master" | "main") {
             Ok(BranchClass::ProtectedBranch)
-        } else if branch_name.contains("/trunk-merge/") {
+        } else if branch_name.contains("/trunk-merge/")
+            || branch_name.contains("/gh-readonly-queue/")
+            || merge_request_event_type.filter(|t| t == "merge").is_some()
```
dfrankland, line 437: This needs to match `merge_train`, no?

dfrankland:
Also, can we add a test for this?

context/src/env/parser.rs
jc3m: Switched to using an enum and parsing the value

context/src/env/parser.rs
jc3m: Ah yep, using the new enum now

jc3m:
> Also, can we add a test for this?

Good call, added tests!

context/src/env/validator.rs
```diff
@@ -248,12 +248,12 @@ pub fn validate(ci_info: &CIInfo) -> EnvValidation {
...
                 ));
             }
-            (BranchClass::Merge | BranchClass::ProtectedBranch, Some(..)) => {
+            (BranchClass::ProtectedBranch, Some(..)) => {
                 env_validation.add_issue(EnvValidationIssue::Invalid(
                     EnvValidationIssueInvalid::CIInfoPRNumberConflictsWithBranchClass,
                 ));
             }
-            (BranchClass::PullRequest, Some(..))
+            (BranchClass::PullRequest | BranchClass::Merge, Some(..))
```
jc3m, line 256: It seems like there is some inconsistency with how merge branches are implemented. We don't set the PR for Trunk merge, but Gitlab does appear to set the PR number in the CI environment for merge trains

dfrankland: Yep, you are correct. I think we can ignore the PR number for GitLab merge trains 👍

dfrankland (approved):
Looks pretty fire!

PR #202, opened 2024-12-06
[TRUNK-13581] unified env+repo validation

max-trunk:
Part of [TRUNK-13581](https://linear.app/trunk/issue/TRUNK-13581/move-env-parsing-and-validation-out-of-glue-and-into-dagster)

(apologies in advance for the large diff)

Adds a new 'meta context' validator that validates the combination of an `env::parser::CIInfo` and a `repo::BundleRepo` to prepare for usage in the ETL

The motivation for this is that, there are some instances where we rely on and validate against a combination of values provided by environment variables and the repo details we collect. For example, for many values, if absent from the environment variables, we [fall back to what's in the repo details](https://github.com/trunk-io/trunk/blob/74dc8d323c7a372562ea5d0942d3a63981cf8c65/trunk/py/metrics_data_pipeline/trunk_etl_env_parser/env_parser.py#L114-L118). If we were to slot in the env & repo parser & validator as they are now into the ETL, we could start flagging invalid validation issues for missing env values that would otherwise have fallen back to repo values, and vice-versa. I think this validator also provides a decently-extensible path towards validating the metadata as a whole.

See https://github.com/trunk-io/trunk/pull/19954 for usage

Summary of changes:
 - new `MetaContext` struct & validator
 - changes env parsing to not fail. Previously, inability to detect CI platform or classify a branch name would result in an error. Now allows CI platform and branch class both to be `UNKNOWN`. This mirrors what the ETL is currently doing [(1)](https://github.com/trunk-io/trunk/blob/74dc8d323c7a372562ea5d0942d3a63981cf8c65/trunk/py/metrics_data_pipeline/trunk_etl_env_parser/env_parser.py#L84) [(2)](https://github.com/trunk-io/trunk/blob/74dc8d323c7a372562ea5d0942d3a63981cf8c65/trunk/py/metrics_data_pipeline/trunk_etl_env_parser/env_parser.py#L232) and will continue to allow things like local uploads to be processed. 
 - re-classifies some env & repo validation issues from `INVALID` to `SUBOPTIMAL` since the meta validator will now catch them as `INVALID` (e.g. missing branch name is suboptimal in the isolated contexts of env and repo validation but invalid in the context of unified meta validation)
 - breaks up the `context-py` `test_parse_and_validate.py` into several smaller files since it was getting a bit unwieldy


context/src/env/validator.rs
```diff
@@ -76,21 +88,18 @@ pub enum EnvValidationIssueSubOptimal {
     CIInfoJobURLTooShort(String),
     #[error("CI info job URL too long, truncated to {}", MAX_FIELD_LEN)]
     CIInfoJobURLTooLong(String),
+    #[error("CI info is classified as a PR, but has not PR number")]
```
TylerJang27, line 91: *has no

context/src/env/validator.rs
```diff
@@ -244,17 +253,17 @@ pub fn validate(ci_info: &CIInfo) -> EnvValidation {
...
+                env_validation.add_issue(EnvValidationIssue::SubOptimal(
+                    EnvValidationIssueSubOptimal::CIInfoPRNumberMissing,
                 ));
             }
-            (BranchClass::Merge | BranchClass::ProtectedBranch, Some(..)) => {
-                env_validation.add_issue(EnvValidationIssue::Invalid(
-                    EnvValidationIssueInvalid::CIInfoPRNumberConflictsWithBranchClass,
+            (BranchClass::Merge | BranchClass::ProtectedBranch | BranchClass::None, Some(..)) => {
+                env_validation.add_issue(EnvValidationIssue::SubOptimal(
+                    EnvValidationIssueSubOptimal::CIInfoPRNumberConflictsWithBranchClass,
```
TylerJang27, line 262: I know you just moved this over. But I would expect this to actually be incorrect. Merge branch classes may have PR numbers. Depending on what setting you have configured, Trunk Merge will make a PR that for things that you enqueue and will merge those

I'm assuming we aren't using/surfacing these yet or we would be getting a bunch of suboptimal issues for this

context/src/env/validator.rs
```diff
@@ -76,21 +88,18 @@ pub enum EnvValidationIssueSubOptimal {
...
 #[derive(Error, Debug, Clone, PartialEq, Eq)]
-pub enum EnvValidationIssueInvalid {
-    #[error("CI info branch name too short")]
-    CIInfoBranchNameTooShort(String),
-    #[error("CI info is classified as a PR, but has not PR number")]
-    CIInfoPRNumberMissing,
-    #[error("CI info has a PR number, but branch is not classified as a PR")]
-    CIInfoPRNumberConflictsWithBranchClass,
-}
+pub enum EnvValidationIssueInvalid {}
```
TylerJang27, line 102: Is this used anymore? Should we just delete it?

context/src/meta/validator.rs
```diff
@@ -0,0 +1,290 @@
...
+    #[error("CI info branch name too short")]
+    CIInfoBranchNameTooShort(String),
+    #[error("CI info is classified as a PR, but has not PR number")]
+    CIInfoPRNumberMissing,
+    #[error("CI info has a PR number, but branch is not classified as a PR")]
+    CIInfoPRNumberConflictsWithBranchClass,
+}
+
+#[derive(Error, Debug, Clone, PartialEq, Eq)]
+pub enum MetaValidationIssueSubOptimal {
```
TylerJang27, line 70: If I'm reading this right, this is a subset of `EnvValidationIssueSubOptimal`, as is `MetaValidationIssueInvalid` above. Can we refactor this a bit to avoid the duplication here?

context/src/meta/validator.rs
TylerJang27, line 1: I like having a unified interface for all this, but I'm not sure I'm bought in on all the duplication that this file contains. Can you explain the tradeoffs here a bit more, and/or try to refactor out some of the duplication?

TylerJang27 (commented):
Left some notes. Deferring stamp to further discussion and also my review of https://github.com/trunk-io/trunk/pull/19954

context/src/meta/validator.rs
max-trunk: Yeah, I get that. I like the separate `MetaContext` and interface for the 'unified' validation of env+repo because it is a distinct kind of validation. It just so happens that the validation logic is quite duplicative of that of the env validation. 

I refactored things to reduce that duplication. The meta validator now uses the env validation logic and its resulting issues to derive the final meta validation issues / levels. LMK what you think of that approach!

context/src/env/validator.rs
max-trunk: Not currently used, but I'm fine with leaving it so that if we want to add an invalid issue of this type in the near future, the code is already equipped to handle it

context/src/env/validator.rs
max-trunk: Good eye! looks like this was caught and resolved in https://github.com/trunk-io/analytics-cli/pull/199#discussion_r1876357400

context-js/src/lib.rs
```diff
@@ -31,7 +31,11 @@ pub fn env_parse(env_vars: js_sys::Object) -> Option<env::parser::CIInfo> {
 }
 
 #[wasm_bindgen]
-pub fn parse_branch_class(value: &str, pr_number: Option<usize>, gitlab_merge_request_event_type: Option<env::parser::GitLabMergeRequestEventType>) -> env::parser::BranchClass {
+pub fn parse_branch_class(
+    value: &str,
+    pr_number: Option<usize>,
+    gitlab_merge_request_event_type: Option<env::parser::GitLabMergeRequestEventType>,
```
TylerJang27, line 37: optional: I really feel like this input should be provider-agnostic or part of a larger struct

(I realize this isn't from your changes)

context/src/meta/validator.rs
```diff
@@ -108,134 +80,40 @@ impl From<&MetaValidationIssue> for MetaValidationLevel {
...
+                        env_validation_issue_suboptimal.clone(),
+                    ),
+                ))
             }
-            (BranchClass::PullRequest, Some(..))
-            | (BranchClass::Merge | BranchClass::ProtectedBranch | BranchClass::None, None) => (),
-        };
-    }
+            _ => None,
+        } {
```
TylerJang27, line 113: nit: make this a reduce/fold

TylerJang27 (approved):
Nice!

PR #203, opened 2024-12-06
Add proto generation skeleton

gnalh:
Boilerplate to get protos working. I split this off from another branch to make reviewing easier. Reviewers: Please validate that the directory structure makes sense.

### DIFF
--- a/.github/actions/setup_proto/action.yaml
+++ b/.github/actions/setup_proto/action.yaml
@@ -1,0 +1,11 @@
+name: Setup Proto
+description: Setup Proto
+
+runs:
+  using: composite
+  steps:
+    - name: Install cmake
+      if: runner.os == 'Linux'
+      run: |
+        sudo apt-get install cmake -y
+      shell: bash
--- a/proto/proto/test_context.proto
+++ b/proto/proto/test_context.proto
@@ -1,0 +1,8 @@
+syntax = "proto3";
+
+import "google/protobuf/timestamp.proto";
+
+package test_context.test_run;
+
+message TestCaseRun {
+}
--- a/proto/src/build.rs
+++ b/proto/src/build.rs
@@ -1,0 +1,7 @@
+use std::io::Result;
+
+fn main() -> Result<()> {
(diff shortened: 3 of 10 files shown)


.github/actions/setup_proto/action.yaml
```diff
@@ -0,0 +1,9 @@
+name: Setup Ruby
+description: Setup Ruby
+
+runs:
+  using: composite
+  steps:
+    - name: Install cmake
+      run: |
+        sudo apt-get install cmake
```
gnalh, line 9: Protoc requires `cmake`. These setup actions could be resolved by pulling in a build system.

proto/src/build.rs
```diff
@@ -0,0 +1,7 @@
+use std::io::Result;
+
+fn main() -> Result<()> {
+    std::env::set_var("PROTOC", protobuf_src::protoc());
+    prost_build::compile_protos(&["proto/test_context.proto"], &["proto/"])?;
+    Ok(())
+}
```
gnalh, line 7: wrap calling protoc so that you don't have to do it. Similarly, if we used a build system then we would have it handle this piece.

proto/src/test_context.rs
```diff
@@ -0,0 +1,15 @@
...
+    include!(concat!(env!("OUT_DIR"), "/test_context.test_run.rs"));
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::test_context::test_run::TestCaseRun;
+    #[test]
+    fn create_test_run() {
+        let test_case_run = TestCaseRun::default();
+        assert_eq!(test_case_run, TestCaseRun::default());
```
gnalh, line 13: Simple validation test. Real tests will come in followup.

proto/proto/test_context.proto
```diff
@@ -0,0 +1,8 @@
+syntax = "proto3";
```
gnalh, line 1: Actual definition will come in followup. Just wanted to have the skeleton in place.

dfrankland (approved):

PR #205, opened 2024-12-07
(Chore): Cleanup bazel-bep usage

TylerJang27:
Follow-up to #197. Also vendors bazel-bep within this repo rather than using my fork.


cli/src/print.rs
```diff
@@ -0,0 +1,18 @@
+use colored::Colorize;
+use context::bazel_bep::parser::BazelBepParser;
+
+pub fn print_bep_results(parser: &BazelBepParser) {
+    if !parser.errors().is_empty() {
+        println!(
```
TylerJang27, line 6: I remember there was some discussion about the verbose log messages vs concise printing... What do we prefer now?

dfrankland: Right now it's concise only during the `validate` command, but not outside of that. I think we should probably match what JUnit parsing does for now?
https://github.com/trunk-io/analytics-cli/blob/9e7443f9ed99f36d80afbd7bdfbfa48b157522a6/codeowners/src/codeowners.rs#L59

dfrankland (approved):

pv72895 (approved):

PR #206, opened 2024-12-09
Test failures are passed through

cmillar-trunk:
Adds tests to validate that the test command passes or fails as the test it ran passed or failed, rather than reporting the status of an upload attempt. This behaviour is already present and desirable, but was untested.


gnalh (approved):
LGTM

dfrankland:
This will fail if we merge right now since the way we parse some of the arguments has been updated

PR #207, opened 2024-12-09
fix setting default-true flags to false

dfrankland:
more thorough testing


TylerJang27 (approved):
🙏

cli/src/upload.rs
```diff
@@ -99,6 +100,7 @@ pub struct UploadArgs {
         help = "Do not fail if test results are not found.",
         action = ArgAction::Set,
         required = false,
+        require_equals = true,
```
dfrankland, line 103: without these, argument parsing fails, and consequently the above test fails

PR #208, opened 2024-12-09
(Chore): Build protoc from source

TylerJang27:
Rather than installing protoc in CI, install from source using cmake. This changes https://github.com/trunk-io/analytics-cli/pull/197 to behave more like https://github.com/trunk-io/analytics-cli/pull/203


dfrankland:
superseded by https://github.com/trunk-io/analytics-cli/pull/256

PR #209, opened 2024-12-10
(Fix): Calculate failed tests accounting for retries

TylerJang27:
Uses the last timestamp for a given test ID to determine whether or not it failed, for the sake of setting a local exit code.

This is primarily to support `--bazel-bep-path`, which if bazel retries are on for flaky tests, [will output multiple junit xmls for a single test](https://github.com/trunk-io/trunk/actions/runs/12245144578/job/34158530182?pr=19962). Adds unit and integration tests to cover the repro case.
- There were a couple different options for how to approach this (BEP introspection, passing junit rankings along, etc.), but using the timestamp seems to be the most generic/extensible, and we still defer to the services for any final parsing, since the only logic we're applying here is setting the exit code

### DIFF
--- a/bundle/src/types.rs
+++ b/bundle/src/types.rs
@@ -31,6 +31,7 @@ pub struct Test {
     pub class_name: Option<String>,
     pub file: Option<String>,
     pub id: String,
+    pub timestamp_millis: Option<i64>,
 }
 
 impl Test {
@@ -42,6 +43,7 @@ impl Test {
         id: Option<String>,
         org_slug: &str,
         repo: &BundleRepo,
+        timestamp_millis: Option<i64>,
     ) -> Self {
         if let Some(id) = id {
             return Test {
@@ -50,6 +52,7 @@ impl Test {
                 class_name,
                 file,
                 id,
+                timestamp_millis,
             };
         }
         // generate a unique id if not provided
@@ -72,6 +75,7 @@ impl Test {
             class_name,
             file,
             id,
+            timestamp_millis,
         }
     }
 }
@@ -267,6 +271,7 @@ mod tests {
             None,
             org_slug,
             &repo,
+            Some(0),
         );
         assert_eq!(result.name, name);
         assert_eq!(result.parent_name, parent_name);
@@ -281,6 +286,7 @@ mod tests {
             Some(String::from("da5b8893-d6ca-5c1c-9a9c-91f40a2a3649")),
             org_slug,
             &repo,
+            Some(0),
         );
         assert_eq!(result.name, name);
         assert_eq!(result.parent_name, parent_name);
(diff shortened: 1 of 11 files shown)


bundle/src/types.rs
```diff
@@ -31,6 +31,7 @@ pub struct Test {
     pub class_name: Option<String>,
     pub file: Option<String>,
     pub id: String,
+    pub timestamp_millis: Option<i64>,
```
TylerJang27, line 34: I'm not sure of the full implications of expanding this type, since there are bindings for it. I can remove it from here if desired, and just store pairs/"max timestamp" in the extraction map

cli-tests/src/upload.rs
```diff
@@ -232,6 +235,59 @@ async fn upload_bundle_using_bep() {
...
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn upload_bundle_success_status_code() {
+    let temp_dir = tempdir().unwrap();
+    generate_mock_git_repo(&temp_dir);
+    let test_bep_path = get_test_file_path("test_fixtures/bep_retries");
+    let uri_fail = format!(
+        "file://{}",
+        get_test_file_path("../cli/test_fixtures/junit0_fail.xml")
```
TylerJang27, line 245: I don't love looking out of the crate, but this is certainly useful test data that is nontrivial to generate. The next best thing would probably be to make a common test data dir directory, but I didn't want to undertake that unilaterally here

cli/src/runner.rs
```diff
@@ -264,3 +306,99 @@ pub async fn run_quarantine(
...
+    use test_utils::inputs::get_test_file_path;
+
+    /// Contains 1 failure at 1:00
+    const JUNIT0_FAIL: &str = "test_fixtures/junit0_fail.xml";
+    // Contains 1 pass at 2:00
+    const JUNIT0_PASS: &str = "test_fixtures/junit0_pass.xml";
+    // Contains 1 failure at 3:00 and 1 failure at 5:00
+    const JUNIT1_FAIL: &str = "test_fixtures/junit1_fail.xml";
+    // Contains 2 passes at 4:00
+    const JUNIT1_PASS: &str = "test_fixtures/junit1_pass.xml";
```
TylerJang27, line 323: provides a variety of cases with which to test timestamp interleaving. Mainly the only pairing we care about though is `[JUNIT0_FAIL, JUNIT0_PASS]`, and the rest are just to prevent regressions

TylerJang27:
Spoke offline, going to walk this back partially. If there are failures, we still want to have a failing exit code, since we want the quarantining service to have the primarily responsibility of detecting flakiness.

This may provide some initial friction, and we may want to have a flag that does exactly what the current state of this PR is, but that's not desired behavior for now

More details in Slack

pv72895 (approved):

PR #214, opened 2024-12-10
Update the CLI to work with the updated APIs format

federicoweber:
APIs changes in https://github.com/trunk-io/trunk/pull/20034

### DIFF
--- a/api/src/lib.rs
+++ b/api/src/lib.rs
@@ -58,8 +58,6 @@ pub struct GetQuarantineBulkTestStatusRequest {
 
 #[derive(Debug, Serialize, Clone, Deserialize, Default)]
 pub struct QuarantineConfig {
-    #[serde(rename = "isPreview")]
-    pub is_preview_mode: bool,
     #[serde(rename = "isDisabled")]
     pub is_disabled: bool,
     #[serde(rename = "testIds")]
--- a/api/src/lib.rs
+++ b/api/src/lib.rs
@@ -1,7 +1,7 @@
-use std::collections::HashSet;
-
+use bundle::types::Test;
 use context::repo::RepoUrlParts;
 use serde::{Deserialize, Serialize};
+use std::collections::HashSet; // Add this import
 
 #[derive(Debug, Serialize, Clone, Deserialize, PartialEq, Eq)]
 pub struct CreateBundleUploadRequest {
@@ -54,6 +54,8 @@ pub struct GetQuarantineBulkTestStatusRequest {
     pub repo: RepoUrlParts,
     #[serde(rename = "orgUrlSlug")]
     pub org_url_slug: String,
+    #[serde(rename = "testIdentifiers")]
+    pub test_identifiers: Vec<Test>,
 }
 
 #[derive(Debug, Serialize, Clone, Deserialize, Default)]
--- a/api/src/lib.rs
+++ b/api/src/lib.rs
@@ -1,7 +1,6 @@
-use bundle::types::Test;
+use bundle::Test;
 use context::repo::RepoUrlParts;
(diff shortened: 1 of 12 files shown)


bundle/src/types.rs
```diff
@@ -20,7 +21,7 @@ pub struct QuarantineRunResult {
     pub quarantine_status: QuarantineBulkTestStatus,
 }
 
-#[derive(Debug, Serialize, Clone, Deserialize, PartialEq, Eq)]
+#[derive(Debug, Serialize, Clone, Deserialize, PartialEq, Eq, Hash)] // Add `Hash` here
```
dfrankland, line 24: Is this needed? Seems like we're just sticking `Test` in `Vec`

bundle/src/lib.rs
```diff
@@ -2,7 +2,7 @@ mod bundle_meta;
 mod bundler;
 mod custom_tag;
 mod files;
-mod types;
+pub mod types;
```
dfrankland, line 5: Not needed, we already `pub use types::*;` down below

api/src/lib.rs
```diff
@@ -1,7 +1,7 @@
-use std::collections::HashSet;
-
+use bundle::types::Test;
 use context::repo::RepoUrlParts;
 use serde::{Deserialize, Serialize};
+use std::collections::HashSet; // Add this import
```
dfrankland, line 4: Can we undo the change to `std::collections::HashSet` here? It was more nicely organized before

cli-tests/src/upload.rs
```diff
@@ -51,17 +48,27 @@ async fn upload_bundle() {
...
-            repo: Repo {
-                host: String::from("github.com"),
-                owner: String::from("trunk-io"),
-                name: String::from("analytics-cli"),
-            },
-            org_url_slug: String::from("test-org"),
-        })
-    );
+    let quarantine_request = requests_iter.next().unwrap();
+    if let RequestPayload::GetQuarantineBulkTestStatus(req) = quarantine_request {
```
dfrankland, line 52: nit: use `assert_matches!()` for this. There's an example below on how to use it

cli/src/api_client/mod.rs
```diff
@@ -159,10 +160,14 @@ impl ApiClient {
...
-                    .await
-                    .context("Failed to get response body as json.")
+                let response_body = response.text().await?;
+                match serde_json::from_str::<api::QuarantineConfig>(&response_body) {
+                    Ok(config) => Ok(config),
+                    Err(e) => {
+                        log::error!("Error retrieving quarantine configuration");
+                        return Err(anyhow::anyhow!("Failed to parse response body as json: {}", e));
+                    }
+                }
```
dfrankland, line 170: No need for this change. This is effectively the same thing, but much more verbose

cli/src/api_client/mod.rs
```diff
@@ -172,7 +177,14 @@ impl ApiClient {
...
         .call_api()
-        .await
+        .await;
+
+        if let Err(ref e) = result {
+            log::error!("Error during API call: {}", e);
+            return Err(anyhow::anyhow!(e.to_string()));
+        }
+
+        result
```
dfrankland, line 187: nit: remove some redundancy

```suggestion
        if result.is_err() {
            log::error!("Error retrieving quarantine configuration");
        }

        result
```

cli/src/api_client/mod.rs
federicoweber: ah sorry forgot to clean this up, I broke it down for debugging to log the body

dfrankland (approved):

PR #223, opened 2024-12-12
[TRUNK-13813] Create smoke tests workflow

cmillar-trunk:
Creates a smoke tests workflow, which runs the analytics cli on a release tag.

### DIFF
--- a/.github/workflows/smoke_test.yml
+++ b/.github/workflows/smoke_test.yml
@@ -17,11 +17,6 @@ jobs:
       - name: Setup Rust & Cargo
         uses: ./.github/actions/setup_rust_cargo
 
-      - name: Run failing test
-        run: |
-          cd cli-smoke-test
-          ../trunk-analytics-cli test cargo nextest always_fails --profile ci
-
       - name: Run succeeding test
         run: |
           cd cli-smoke-test
--- a/.github/workflows/smoke_test.yml
+++ b/.github/workflows/smoke_test.yml
@@ -1,21 +1,19 @@
 name: Smoke Test
+permissions:
+  actions: read
 on:
-  workflow_dispatch:
-    inputs:
-      release:
-        type: string
-        description: Release number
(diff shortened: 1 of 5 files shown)


.github/workflows/smoke_test.yml
```diff
@@ -0,0 +1,38 @@
...
+        uses: robinraju/release-downloader@v1
+        with:
+          tag: ${{ github.event.inputs.release }}
+          fileName: trunk-analytics-cli-aarch64-unknown-linux.tar.gz
+          extract: true
+
+      - name: Checkout Test Repo
+        uses: actions/checkout@v4
+        with:
+          repository: trunk-io/cli-smoke-test
```
dfrankland, line 25: This should live inside the `analytics-cli` repo

.github/workflows/smoke_test.yml
```diff
@@ -0,0 +1,38 @@
...
+  build_cli:
+    name: Build cli from release ${{ github.event.inputs.release }}
+    runs-on: public-amd64-2xlarge
+    steps:
+      - name: Get Release
+        uses: robinraju/release-downloader@v1
+        with:
+          tag: ${{ github.event.inputs.release }}
+          fileName: trunk-analytics-cli-aarch64-unknown-linux.tar.gz
+          extract: true
```
dfrankland, line 20: This is fine, except we the runner is AMD64, so we should be using the `x86_64` release binary.

IMO, it'd be better to use [the setup script we have in our docs](https://docs.trunk.io/flaky-tests/get-started/ci-providers/otherci#example-upload-script) which is essentially just
```bash
curl -fsSLO --retry 3 https://github.com/trunk-io/analytics-cli/releases/download/${{ github.event.inputs.release }}/trunk-analytics-cli-x86_64-unknown-linux.tar.gz && chmod +x trunk
```

dfrankland (approved):

PR #226, opened 2024-12-13
(Feat): Upload streamlined BEP file

TylerJang27:
Include a subset of the BEP file in the bundle tar. This strips it to just `testResult` events, which removes sensitive information that may be in other events

This is mainly for our debugging purposes and for evaluating the success of this feature

### DIFF
--- a/bundle/src/bundler.rs
+++ b/bundle/src/bundler.rs
@@ -3,10 +3,10 @@ use async_std::{io::ReadExt, stream::StreamExt};
 use async_tar_wasm::Archive;
 use context::bazel_bep::parser::BepParseResult;
 use futures_io::AsyncBufRead;
-use std::path::PathBuf;
 use std::{
     fs::File,
     io::{Seek, Write},
+    path::PathBuf,
 };
 #[cfg(feature = "wasm")]
 use tsify_next::Tsify;
@@ -21,8 +21,8 @@ use crate::bundle_meta::{BundleMeta, VersionedBundle};
 ///
 #[cfg_attr(feature = "wasm", derive(Tsify))]
 pub struct BundlerUtil {
-    pub meta: BundleMeta,
-    pub bep_result: Option<BepParseResult>,
+    meta: BundleMeta,
+    bep_result: Option<BepParseResult>,
 }
 
 const META_FILENAME: &'static str = "meta.json";
@@ -76,12 +76,16 @@ impl BundlerUtil {
         }
 
         if let Some(bep_result) = self.bep_result.as_ref() {
-            let mut bep_events_file = tempfile::tempfile()?;
-            bep_result.bep_test_events.iter().for_each(|event| {
-                if let Err(e) = serde_json::to_writer(&bep_events_file, event) {
-                    log::error!("Failed to write BEP event: {}", e);
-                }
-            });
+            let mut bep_events_file =
+                bep_result
+                    .bep_test_events
+                    .iter()
+                    .fold(tempfile::tempfile()?, |f, event| {
+                        if let Err(e) = serde_json::to_writer(&f, event) {
+                            log::error!("Failed to write BEP event: {}", e);
+                        }
+                        f
+                    });
             bep_events_file.flush()?;
             bep_events_file.seek(std::io::SeekFrom::Start(0))?;
             tar.append_file("bazel_bep.json", &mut bep_events_file)?;
--- a/bundle/src/bundler.rs
+++ b/bundle/src/bundler.rs
@@ -1,9 +1,13 @@
 use async_compression::futures::bufread::ZstdDecoder;
 use async_std::{io::ReadExt, stream::StreamExt};
 use async_tar_wasm::Archive;
+use context::bazel_bep::parser::BazelBepParser;
 use futures_io::AsyncBufRead;
-use std::io::{Seek, Write};
 use std::path::PathBuf;
+use std::{
+    fs::File,
+    io::{Seek, Write},
+};
 #[cfg(feature = "wasm")]
 use tsify_next::Tsify;
 #[cfg(feature = "wasm")]
@@ -18,6 +22,7 @@ use crate::bundle_meta::{BundleMeta, VersionedBundle};
 #[cfg_attr(feature = "wasm", derive(Tsify))]
 pub struct BundlerUtil {
     pub meta: BundleMeta,
+    pub bep_parser: Option<BazelBepParser>,
 }
 
 const META_FILENAME: &'static str = "meta.json";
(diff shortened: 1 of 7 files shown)


bundle/src/bundler.rs
```diff
@@ -65,11 +74,24 @@ impl BundlerUtil {
...
 
+        if let Some(bep_parser) = self.bep_parser.as_ref() {
+            let mut bep_events_file = tempfile::tempfile()?;
+            bep_parser.bep_test_events().iter().for_each(|event| {
+                if let Err(e) = serde_json::to_writer(&bep_events_file, event) {
+                    log::error!("Failed to write BEP event: {}", e);
+                }
+            });
+            bep_events_file.flush()?;
+            bep_events_file.seek(std::io::SeekFrom::Start(0))?;
```
TylerJang27, line 90: This seemed to be necessary in order for the tests to work on Mac.

context/src/bazel_bep/parser.rs
```diff
@@ -63,23 +68,27 @@ impl BazelBepParser {
...
+                ),
+                |(mut errors, mut test_results, mut bep_test_events), parse_event| {
                     match parse_event {
                         Result::Err(ref err) => {
                             errors.push(format!("Error parsing build event: {}", err));
                         }
                         Result::Ok(build_event) => {
-                            if let Some(Payload::TestResult(test_result)) = build_event.payload {
-                                let xml_files = test_result
+                            if let Some(Payload::TestResult(test_result)) = &build_event.payload {
```
TylerJang27, line 85: I will also be uploading `TestSummary` events in a follow-up. I considered the `started` events too, but that can include some env vars and bazel config that we probably shouldn't collect

bundle/src/bundler.rs
```diff
@@ -65,11 +74,24 @@ impl BundlerUtil {
...
         if let Some(CodeOwners { ref path, .. }) = self.meta.base_props.codeowners {
-            let mut file = std::fs::File::open(path)?;
+            let mut file = File::open(path)?;
             tar.append_file("CODEOWNERS", &mut file)?;
             total_bytes_in += std::fs::metadata(path)?.len();
         }
 
+        if let Some(bep_parser) = self.bep_parser.as_ref() {
+            let mut bep_events_file = tempfile::tempfile()?;
+            bep_parser.bep_test_events().iter().for_each(|event| {
```
max-trunk, line 84: This assumes that `bep_parser.parse()` was called somewhere upstream (and I do see that it was called in the upload flow before being passed to the `BundlerUtil` here). Curious if you've considered having the `BazelBepParser` output a single struct with all of the relevant parsed data that we can pass around rather than the parser itself?

max-trunk (approved):
LGTM, just 1 Q!

bundle/src/bundler.rs
```diff
@@ -1,9 +1,13 @@
...
 use async_std::{io::ReadExt, stream::StreamExt};
 use async_tar_wasm::Archive;
+use context::bazel_bep::parser::BazelBepParser;
 use futures_io::AsyncBufRead;
-use std::io::{Seek, Write};
 use std::path::PathBuf;
+use std::{
+    fs::File,
+    io::{Seek, Write},
+};
```
dfrankland, line 10: would be awesome if we could combine these `std` imports and stick them in their own group above these normal crate imports

bundle/src/bundler.rs
```diff
@@ -18,23 +22,28 @@ use crate::bundle_meta::{BundleMeta, VersionedBundle};
 #[cfg_attr(feature = "wasm", derive(Tsify))]
 pub struct BundlerUtil {
     pub meta: BundleMeta,
+    pub bep_parser: Option<BazelBepParser>,
```
dfrankland, line 25: I don't think either of these need to be `pub`

bundle/src/bundler.rs
dfrankland: Agreed—might be better to have a struct which represents an already parsed state

bundle/src/bundler.rs
```diff
@@ -65,11 +74,24 @@ impl BundlerUtil {
...
             total_bytes_in += std::fs::metadata(path)?.len();
         }
 
+        if let Some(bep_parser) = self.bep_parser.as_ref() {
+            let mut bep_events_file = tempfile::tempfile()?;
+            bep_parser.bep_test_events().iter().for_each(|event| {
+                if let Err(e) = serde_json::to_writer(&bep_events_file, event) {
+                    log::error!("Failed to write BEP event: {}", e);
+                }
+            });
```
dfrankland, line 88: nit, optional: `for_each` is for side-effects. IMO, `fold` or even a `for` loop might be more appropriate for showing intention

```suggestion
            let mut bep_events_file =
                bep_parser
                    .bep_test_events()
                    .iter()
                    .fold(tempfile::tempfile()?, |f, event| {
                        if let Err(e) = serde_json::to_writer(&f, event) {
                            log::error!("Failed to write BEP event: {}", e);
                        }
                        f
                    });
```

bundle/src/bundler.rs
dfrankland: Yeah, I think it's necessary when you write then immediately read again within the same scope

cli-tests/src/upload.rs
```diff
@@ -223,14 +225,24 @@ async fn upload_bundle_using_bep() {
...
-    assert!(junit_parser.parse(reader).is_ok());
+    assert!(junit_parser.parse(junit_reader).is_ok());
     assert!(junit_parser.errors().is_empty());
 
+    let mut bazel_bep_parser = BazelBepParser::new(
+        tar_extract_directory
+            .join("bazel_bep.json")
+            .to_string_lossy()
+            .to_string(),
+    );
```
dfrankland, line 241: If we update `BazelBepParser` you won't have to do all of this `Path` <-> `String` conversion:

```rust
#[derive(Debug, Clone, Default)]
pub struct BazelBepParser {
    bazel_bep_path: PathBuf,
    test_results: Vec<TestResult>,
    bep_test_events: Vec<BuildEvent>,
    errors: Vec<String>,
}

impl BazelBepParser {
    pub fn new<T: Into<PathBuf>>(bazel_bep_path: T) -> Self {
        Self {
            bazel_bep_path: bazel_bep_path.into(),
            ..Default::default()
        }
    }
    // ... other code ...
}
```

dfrankland (approved):

bundle/src/bundler.rs
TylerJang27: Good point. I'm refactoring to handle that

bundle/src/bundler.rs
TylerJang27: 👍 Need a linter for this 🙄

cli-tests/src/upload.rs
TylerJang27: Everytime I'm sitting here thinking "there has to be a better way," Dylan's got me

PR #237, opened 2024-12-14
Check if cli exists before executing

pat-trunk-io:
Check if CLI exists before executing.

We have seen some sporadic exit code 127.

https://trunk-io.slack.com/archives/C044VBASZ0D/p1734148787307919?thread_ts=1734148787.307919&cid=C044VBASZ0D

### DIFF
--- a/.github/workflows/smoke_test.yml
+++ b/.github/workflows/smoke_test.yml
@@ -39,6 +39,7 @@ jobs:
           TRUNK_PUBLIC_API_ADDRESS: https://api.trunk-staging.io
         shell: bash
         run: |
+          ls -lh ./trunk-analytics-cli && \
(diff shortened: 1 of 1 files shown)


gnalh (commented):
I don't think this is what we should do here since it obfuscates the results. I worry we'll see green and assume we're good to release.

pat-trunk-io:
I mean then it's a bit of a Heisenberg bug. If we observe it we can't reproduce, but if we don't observe it will happen.

Fwiw, what I'm trying to figure is "what" is missing here. Exit code 127 means an executable was missing, but is it the analytic cli or something that it calls?

gnalh:
The failure happened when running nextest. It doesn't seem to be a missing cli.

PR #240, opened 2024-12-16
[TRUNK-13813] Smoke test PRs

cmillar-trunk:
Adds the smoke test to prs. Does not notify slack to avoid noise.

### DIFF
--- a/.github/workflows/smoke_test_main.yml
+++ b/.github/workflows/smoke_test_main.yml
@@ -19,18 +19,8 @@ jobs:
           target: x86_64-unknown-linux-musl
           profile: release
 
-      - name: Setup Rust & Cargo
-        uses: ./.github/actions/setup_rust_cargo
-
-      - name: Install Nextest
-        shell: bash
-        run: |
-          cargo install --version 0.9.85 cargo-nextest --locked
-
-      - name: Install cmake
-        shell: bash
-        run: |
-          sudo apt-get install cmake -y
+      - name: Setup Run Smoke Tests
+        uses: ./.github/actions/setup_run_smoke_tests
 
       - name: Run succeeding test
         env:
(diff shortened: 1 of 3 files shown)


.github/workflows/smoke_test_main.yml
```diff
@@ -0,0 +1,44 @@
...
+        shell: bash
+        run: |
+          sudo apt-get install cmake -y
+
+      - name: Run succeeding test
+        env:
+          TRUNK_PUBLIC_API_ADDRESS: https://api.trunk-staging.io
+        shell: bash
+        run: |
+          ./target/x86_64-unknown-linux-musl/release/trunk-analytics-cli test \
```
pat-trunk-io, line 40: Do we want to run on latest build or the currently latest in the github action. I could see value for both but I think we definitely at least want the latter

cmillar-trunk: Since this is triggered on prs, I want it running on the pr branch. Unless I've horribly misunderstood workflows, using the checkout with `ref: ${{ github.base_ref }}` should be putting us there, right?

pat-trunk-io: Maybe this needs to be a separate test then, but I would really like to test staging services against what is latest in the GH action

dfrankland: Yep, agreed. We want testing of `main` and we want testing of a "pinned" version for both staging and prod.

Additionally, I don't mind a PR smoke test, but I don't think it's as needed right now as the above.

dfrankland, line 40: If most of these steps are shared across smoke test workflows, let's make an action to keep things DRY
https://github.com/trunk-io/analytics-cli/blob/9fbb8d4eb79df5cb78e88e28c8589a2aa4657160/.github/workflows/smoke_test.yml#L24-L46

cmillar-trunk: Ah, gotcha. I'll pivot this to do that instead and post to channel.

dfrankland (approved):
Looks good

PR #241, opened 2024-12-16
Remove parsing errors from empty tags and testsuite names

pv72895:
Baby's first rust PR plz be gentle

## Context
See [here](https://trunk-io.slack.com/archives/C044VBASZ0D/p1734365895854939) for the reasoning behind this. TL:DR is that we were reporting parsing errors for files that don't actually have errors that make them unparsable. These errors were:
1. `StackTraceEmpty` - empty stack trace tag
2. `SystemErrEmpty` empty `<system-err/>`
3. `SystemOutEmpty` empty `system-out/>`
4. `TestSuiteName` empty `testsuite name`

StackAV was getting a ton of errors from this that we were reporting

## Fix
Remove throwing errors relating to these. I left the logic for parsing the tags in there, but just made it so we aren't going to throw an error anymore

## Tests
Update tests in order to addresss

### DIFF
--- a/context-js/tests/parse_and_validate.test.ts
+++ b/context-js/tests/parse_and_validate.test.ts
@@ -72,13 +72,6 @@ describe("context-js", () => {
             <system-err/>
           </testcase>
         </testsuite>
-        <testsuite name="" tests="1" disabled="0" errors="0" failures="1" timestamp="${validTimestamp}">
-          <testcase name="failure-case" file="test.py" classname="MyClass" timestamp="${validTimestamp}" time="1">
-            <failure/>
-            <system-out/>
-            <system-err/>
-          </testcase>
-        </testsuite>
       </testsuites>
     `;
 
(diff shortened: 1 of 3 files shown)


context-js/tests/parse_and_validate.test.ts
```diff
@@ -68,6 +68,8 @@ describe("context-js", () => {
         <testsuite name="my-test-suite" tests="1" disabled="0" errors="0" failures="1" timestamp="${validTimestamp}">
           <testcase name="failure-case" file="test.py" classname="MyClass" timestamp="${validTimestamp}" time="1">
             <failure/>
+            <system-out/>
```
max-trunk, line 71: Worth leaving these in here just to know that we _can_ parse these?

context-py/tests/test_junit_parse.py
```diff
@@ -19,7 +19,8 @@ def test_junit_parse_valid():
           <failure message="AssertionError: assert 'testdata' in '# estdata'">
             FAILURE BODY
           </failure>
-
+          <system-out/>
```
max-trunk, line 22: Same Q here

max-trunk (approved):

context-js/tests/parse_and_validate.test.ts
max-trunk: Oh lol I misread the diff and thought this PR removed them.. my b!!!!!!!!!!!!!!!!!!!!

dfrankland: I'd prefer that these be added into the Rust tests since the bindings tests shouldn't be used to test core functionality since it's 1 degree removed. For example, I could hack in a fix for JS, but not have it work in Rust. If that seems like too much work, let's split a ticket for it.

TylerJang27: +1 to Dylan. Bindings tests should generally be an auxiliary layer to make sure we can consume things correctly. Direct functionality has a better iteration cycle and ownership when it's tested in rust. Can be done in a follow-up though

TylerJang27 (approved):

context-js/tests/parse_and_validate.test.ts
pv72895: Don't worry y'all I understand the benefit of testing closer to the real thing. The parsing tests in rust weren't easy for me to locate - I didn't properly find the rust parsing tests, so I thought that the bindings tests were the main ways you were testing specific functionality. There are no guidelines for how people can contribute to tests - I will update CONTRIBUTING.md to express this

pv72895:
This issue is impeding customers so I'm going to merge this without rust tests because it's taking time for me to do them. I promise I'll follow up with them

PR #242, opened 2024-12-17
(Feat): Parse BEP TestSummary

TylerJang27:
Follow-up to https://github.com/trunk-io/analytics-cli/pull/215 (incomplete, just timestamps) and https://github.com/trunk-io/analytics-cli/pull/226 (more BEP debugging). Parses bazel [TestSummary](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/buildeventstream/proto/build_event_stream.proto#L781) events in order to override the failure status of a given junit. This allows us to properly return a zero exit code if bazel marks a test as flaky, rather than waiting for a quarantine status to properly return.

This should hopefully also allow us to better inform the rules engine for test runner-identified flakiness from bazel

Notes:
- I have attempted to make this as generic as possible without entangling too much logic with bazel parsing
- Unfortunately this logic does have to (and already does) live in the client, since we need to parse the junits in order to properly arrive at a quarantine result
- I have added unit and integration tests here, and also validated this against the repros identified in [thread](https://trunk-io.slack.com/archives/C044VBASZ0D/p1734028034948189).


context/test_fixtures/bep_flaky_summary
TylerJang27, line 1: Once again, sanitized to include real data but stripped for just relevant tests

cli/src/main.rs
```diff
@@ -235,7 +237,13 @@ async fn run(cli: Cli) -> anyhow::Result<i32> {
...
                     bep_result.uncached_xml_files()
                 }
-                None => junit_paths,
+                None => junit_paths
+                    .into_iter()
+                    .map(|p| JunitPathWrapper {
+                        junit_path: p,
+                        status: None,
+                    })
+                    .collect(),
```
dfrankland, line 246: We should probably implement `From<String>` for `JunitPathWrapper` since we repeat that often

bundle/src/files.rs
```diff
@@ -48,6 +51,7 @@ impl FileSet {
     pub fn scan_from_glob(
         repo_root: &str,
         glob_path: String,
+        test_runner_status: Option<TestRunnerJunitStatus>,
```
dfrankland, line 54: Feels like we should just accept `JunitPathWrapper` here instead of making these two separate arguments

context/src/junit/junit_path.rs
```diff
@@ -0,0 +1,29 @@
...
+
+/// Encapsulates the glob path for a junit and, if applicable, the flakiness already
+/// assigned by the user's test runner. See bazel_bep/parser.rs for more.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct JunitPathWrapper {
+    /// Path or glob pattern to the junit file.
+    pub junit_path: String,
+    /// Refers to an optional status parsed from the test runner's output, before junits have been parsed.
+    pub status: Option<TestRunnerJunitStatus>,
+}
```
dfrankland, line 29: Related to genericizing this, I think we ought to consider this a `JunitReportStatus` and `JunitReportFileWithStatus`. Test runners can tell us the status, but it should be fairly trivial for us to derive it from the JUnit report by seeing how many failures and reruns there were. We could even record both what the test runner says and what we derive from the JUnit report for the express reason of checking if those agree and processing quarantining as the test runner says (lest our inference is incorrect). Given that, I think we should always be able to populate `status` and if that's not true, I think we should probably consider adding an `Unknown` variant instead of making `status` `Option<T>`. With this, we can then extend our quarantining check for all JUnit reports rather than just what Bazel or other test runners tell us.

dfrankland: I added a follow ticket to add more generic support for all JUnit report files:
https://linear.app/trunk/issue/TRUNK-13911/parse-junit-files-and-check-for-reruns-for-quarantining-in-analytics

context/src/bazel_bep/parser.rs
```diff
@@ -68,64 +89,100 @@ impl BazelBepParser {
...
+
+                                        test_results.push(TestResult {
+                                            label: id.label.clone(),
+                                            cached,
+                                            xml_files,
+                                            summary_status: None,
+                                        });
+                                        bep_test_events.push(build_event);
+                                    }
+                                    _ => {}
```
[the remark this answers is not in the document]
TylerJang27: Spoke offline, there are 28 different variants here, so probably not worth. I may just make this an `if` though

context/src/junit/junit_path.rs
TylerJang27: Spoke offline, I will undertake the small refactor/renaming here, and then those tickets will capture the follow-up work for parsing the junits themselves

bundle/src/files.rs
```diff
@@ -39,6 +40,8 @@ pub struct FileSet {
     pub file_set_type: FileSetType,
     pub files: Vec<BundledFile>,
     pub glob: String,
+    /// Added in v0.6.11. Populated when parsing from BEP, not from junit globs
+    pub test_runner_status: Option<JunitReportStatus>,
```
TylerJang27, line 44: I ended up making it optional again since that allows us to better deserialize. Otherwise we need to have some versioned logic here for deserializing from meta.json, which doesn't seem worth the effort for this particular case

context/src/junit/junit_path.rs
```diff
@@ -0,0 +1,54 @@
...
+use pyo3::prelude::*;
+#[cfg(feature = "pyo3")]
+use pyo3_stub_gen::derive::gen_stub_pyclass_enum;
+use serde::{Deserialize, Serialize};
+#[cfg(feature = "wasm")]
+use tsify_next::Tsify;
+#[cfg(feature = "wasm")]
+use wasm_bindgen::prelude::*;
+
+use bazel_bep::types::build_event_stream::TestStatus;
```
dfrankland, line 11: nit: these imports can be grouped

cli/src/upload.rs
```diff
@@ -389,7 +395,10 @@ fn handle_xcresult(
                 .map_err(|e| anyhow::anyhow!("Failed to write junit file: {}", e))?;
             let junit_temp_path_str = junit_temp_path.to_str();
             if let Some(junit_temp_path_string) = junit_temp_path_str {
-                temp_paths.push(junit_temp_path_string.to_string());
+                temp_paths.push(JunitReportFileWithStatus {
+                    junit_path: junit_temp_path_string.to_string(),
+                    status: None,
+                });
```
dfrankland, line 401: nit: now that we've implemented `From<String>` for `JunitReportFileWithStatus`, this simplifies to the following:
```suggestion
                temp_paths.push(junit_temp_path_string.into());
```

bundle/src/files.rs
```diff
@@ -47,12 +50,16 @@ impl FileSet {
...
+        glob_with_status: JunitReportFileWithStatus,
         file_counter: &mut FileSetCounter,
         team: Option<String>,
         codeowners: &Option<CodeOwners>,
         start: Option<SystemTime>,
     ) -> anyhow::Result<FileSet> {
+        let JunitReportFileWithStatus {
+            junit_path: glob_path,
+            status: test_runner_status,
+        } = glob_with_status;
```
dfrankland, line 62: You can do this in the args too btw :)

bundle/src/files.rs
gnalh, line 44: Nit: Name. This isn't really for a test runner, but rather the build system around it. Test runner integrations aren't going to use junit.

bundle/src/files.rs
```diff
@@ -147,6 +154,7 @@ impl FileSet {
             file_set_type: FileSetType::Junit,
             files,
             glob: glob_path,
+            test_runner_status,
```
gnalh, line 157: It looks like we only send `glob_with_status` so that we can immediately return it?

gnalh (approved):
Overall this looks fine to me. My main comment is around the naming since test runner integrations will be more direct and not use junit. This logic will really only apply to other build systems we hook into.

bundle/src/files.rs
TylerJang27: Only in the case of extracting failures

PR #245, opened 2024-12-17
[wip] Binary test reports

gnalh:


Cargo.toml
```diff
@@ -12,6 +12,7 @@ members = [
   "smoke-test",
   "test_utils",
   "xcresult",
+  "test_report",
```
gnalh, line 15: This needs to be separate as to prevent circular dependencies.

context-py/src/lib.rs
```diff
@@ -80,6 +81,14 @@ fn junit_parse(xml: Vec<u8>) -> PyResult<Vec<junit::bindings::BindingsReport>> {
         .collect())
 }
 
+#[gen_stub_pyfunction]
+#[pyfunction]
+fn bin_parse(bin: Vec<u8>) -> PyResult<Vec<junit::bindings::BindingsReport>> {
```
gnalh, line 86: I want to decouple the bindings report from quick_junit, but that will be handled in a later task.

context-ruby/spec/spec_helper.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_relative 'trunk_spec_helper'
```
gnalh, line 3: This is what's required to get uploads flowing from rspec directly. Behind the scenes it will automatically parse the results and publish them.

context-ruby/spec/trunk_spec_helper.rb
```diff
@@ -0,0 +1,145 @@
...
+        # TODO: this is where we'll need to override the result once the logic is ready
+        # trunk-ignore(rubocop/Lint/LiteralAsCondition)
+        if true
+          set_exception_core(exception)
+        else
+          # monitor the override in the metadata
+          metadata[:quarantined_exception] = exception
+          nil
+        end
+      end
```
gnalh, line 31: Handling quarantining in a later follow up. Leaving this logic here to show how it can be done.

context-ruby/spec/trunk_spec_helper.rb
```diff
@@ -0,0 +1,145 @@
+# frozen_string_literal: true
```
gnalh, line 1: There are 2 primary components here. One is a custom formatter which is able to parse out the final results. The other is an rspec helper that monitors the test state and modifies the metadata to track values of interest (attempts, quarantining, etc...).

RSpec helpers do not have access to final results because they are considered apart of the run. The way this is defined automatically links the formatter to the test run.

context-ruby/spec/trunk_spec_helper.rb
```diff
@@ -0,0 +1,145 @@
...
+  def description_generated?(example)
+    auto_generated_exp = /^\sis expected to eq .*$/
+    full_description = example.full_description
+    parent_description = example.example_group.description
+    checked_description = full_description.sub(parent_description, '')
+    auto_generated_exp.match(checked_description) != nil
+  end
+
+  def generate_id(example)
+    "#{example.id}-#{example.location}" if description_generated?(example)
```
gnalh, line 102: TBD on if this is the right id generation logic.

context-ruby/test/test_parse_and_validate_spec.rb
```diff
@@ -1,9 +1,10 @@
 # frozen_string_literal: true
 
 require_relative '../lib/context_ruby'
+require_relative '../spec/spec_helper'
 
 # trunk-ignore(rubocop/Metrics/BlockLength)
-describe 'context_ruby' do
+describe do
```
gnalh, line 7: Added as a test for empty parent names.

context/src/junit/validator.rs
```diff
@@ -115,8 +115,8 @@ pub fn validate(report: &Report) -> JunitReportValidation {
             if let Some(raw_test_case_id) = test_case.extra.get("id") {
                 let test_case_id = uuid::Uuid::parse_str(raw_test_case_id).unwrap_or_default();
                 if test_case_id.get_version() != Some(uuid::Version::Sha1) {
-                    test_case_validation.add_issue(JunitValidationIssue::Invalid(
-                        JunitTestCaseValidationIssueInvalid::TestCaseInvalidId(
```
gnalh, line 119: Demote it to suboptimal. `ID` can be set by unrelated junit runners and it's better to just have the ETL / service handle generating it in general.

proto/src/build.rs
```diff
@@ -2,6 +2,10 @@ use std::io::Result;
 
 fn main() -> Result<()> {
     std::env::set_var("PROTOC", protobuf_src::protoc());
-    prost_build::compile_protos(&["proto/test_context.proto"], &["proto/"])?;
+    let mut prost_build = prost_build::Config::new();
+    prost_build
+        .type_attribute(".", "#[derive(serde::Serialize,serde::Deserialize)]")
+        .extern_path(".google.protobuf.Timestamp", "::prost_wkt_types::Timestamp")
+        .compile_protos(&["proto/test_context.proto"], &["proto/"])?;
```
gnalh, line 9: This allows serialization / deserialization. We have to overwrite the google protobuf types with the corresponding prost types in order to get that.

test_report/src/report.rs
```diff
@@ -0,0 +1,269 @@
...
+#[cfg(feature = "wasm")]
+use wasm_bindgen::prelude::wasm_bindgen;
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct TestReport {
+    test_result: TestResult,
+}
+
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
+#[cfg_attr(feature = "ruby", magnus::wrap(class = "TestExecution"))]
```
gnalh, line 16: python bindings aren't needed (yet)

PR #253, opened 2024-12-19
[runner integration - 1] Add support for test report generation

gnalh:
Adds the initial format and integration point for direct test runner integration.

See
* https://github.com/trunk-io/analytics-cli/pull/254
* https://github.com/trunk-io/analytics-cli/pull/255

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,7 @@ members = [
   "smoke-test",
   "test_utils",
   "xcresult",
+  "test_report",
 ]
 exclude = ["context-ruby/ext/context_ruby"]
 resolver = "2"
--- a/cli/src/upload.rs
+++ b/cli/src/upload.rs
@@ -115,6 +115,37 @@ pub struct UploadArgs {
     pub allow_empty_test_results: bool,
 }
 
+impl UploadArgs {
+    pub fn new(
+        token: String,
+        org_url_slug: String,
+        junit_paths: Vec<String>,
+        repo_root: String,
+    ) -> Self {
+        Self {
+            junit_paths,
+            #[cfg(target_os = "macos")]
+            xcresult_path: None,
+            org_url_slug,
+            token,
+            bazel_bep_path: None,
+            repo_root: Some(repo_root),
+            repo_url: None,
+            repo_head_sha: None,
+            repo_head_branch: None,
+            repo_head_commit_epoch: None,
+            tags: Vec::new(),
+            print_files: false,
+            dry_run: false,
+            team: None,
+            // TODO - how are code owners handled?
+            codeowners_path: None,
+            use_quarantining: false,
+            allow_empty_test_results: true,
+        }
+    }
+}
+
 pub async fn run_upload(
     upload_args: UploadArgs,
     test_command: Option<String>,
--- a/proto/proto/test_context.proto
+++ b/proto/proto/test_context.proto
@@ -16,7 +16,6 @@ message TestCaseRun {
   string name = 2;
   string classname = 3;
   string file = 4;
-  // need parent_id
   string parent_name = 5;
   int32 line = 6;
   TestCaseRunStatus status = 7;
--- a/proto/proto/test_context.proto
+++ b/proto/proto/test_context.proto
@@ -4,5 +4,35 @@ import "google/protobuf/timestamp.proto";
 
 package test_context.test_run;
 
+enum TestCaseRunStatus {
+  TEST_CASE_RUN_STATUS_UNSPECIFIED = 0;
+  TEST_CASE_RUN_STATUS_SUCCESS = 1;
+  TEST_CASE_RUN_STATUS_FAILURE = 2;
+  TEST_CASE_RUN_STATUS_SKIPPED = 3;
+}
+
 message TestCaseRun {
+  string id = 1;
+  string name = 2;
+  string classname = 3;
+  string file = 4;
+  // need parent_id
+  string parent_name = 5;
+  int32 line = 6;
+  TestCaseRunStatus status = 7;
+  int32 attempt_number = 8;
+  google.protobuf.Timestamp started_at = 9;
+  google.protobuf.Timestamp finished_at = 10;
+  string status_output_message = 11;
+}
+
+message UploaderMetadata {
+    string version = 1;
+    string origin = 2; // RSpec, jest, etc..
+    google.protobuf.Timestamp upload_time = 3;
+}
+
+message TestResult {
+  repeated TestCaseRun test_case_runs = 1;
+  UploaderMetadata uploader_metadata = 2;
 }
--- a/proto/src/build.rs
+++ b/proto/src/build.rs
@@ -2,6 +2,10 @@ use std::io::Result;
 
 fn main() -> Result<()> {
     std::env::set_var("PROTOC", protobuf_src::protoc());
-    prost_build::compile_protos(&["proto/test_context.proto"], &["proto/"])?;
+    let mut prost_build = prost_build::Config::new();
+    prost_build
+        .type_attribute(".", "#[derive(serde::Serialize,serde::Deserialize)]")
+        .extern_path(".google.protobuf.Timestamp", "::prost_wkt_types::Timestamp")
+        .compile_protos(&["proto/test_context.proto"], &["proto/"])?;
     Ok(())
 }
--- a/test_report/src/report.rs
+++ b/test_report/src/report.rs
@@ -1,0 +1,247 @@
+use chrono::prelude::*;
+#[cfg(feature = "ruby")]
+use magnus::{value::ReprValue, Module, Object};
+use prost_wkt_types::Timestamp;
+use proto::test_context::test_run::{TestCaseRun, TestCaseRunStatus, TestResult, UploaderMetadata};
+use std::cell::RefCell;
+#[cfg(feature = "wasm")]
+use wasm_bindgen::prelude::wasm_bindgen;
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct TestReport {
+    test_result: TestResult,
+}
+
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
+#[cfg_attr(feature = "ruby", magnus::wrap(class = "TestExecution"))]
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TestExecution {
+    id: Option<String>,
+    name: String,
+    classname: String,
+    file: String,
+    parent_name: String,
+    line: Option<i32>,
+    status: Status,
+    attempt_number: i32,
+    started_at: i64,
+    finished_at: i64,
+    output: String,
+}
+
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
+#[cfg_attr(feature = "ruby", magnus::wrap(class = "Status"))]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Status {
+    Success,
+    Failure,
+    Skipped,
+}
+
+#[cfg(feature = "ruby")]
+impl Status {
+    fn new(status: String) -> Self {
+        match status.as_str() {
+            "success" => Status::Success,
+            "failure" => Status::Failure,
+            "skipped" => Status::Skipped,
+            _ => Status::Success,
+        }
+    }
+}
+
+impl Into<&str> for Status {
+    fn into(self) -> &'static str {
+        match self {
+            Status::Success => "success",
+            Status::Failure => "failure",
+            Status::Skipped => "skipped",
+        }
+    }
+}
+
+impl ToString for Status {
+    fn to_string(&self) -> String {
+        String::from(Into::<&str>::into(*self))
+    }
+}
+
+#[cfg(feature = "ruby")]
+impl Status {
+    pub fn to_string(&self) -> &str {
+        (*self).into()
+    }
+}
+
+#[cfg(feature = "ruby")]
+impl magnus::TryConvert for Status {
+    fn try_convert(val: magnus::Value) -> Result<Self, magnus::Error> {
+        let sval: String = val.funcall("to_s", ())?;
+        Ok(Status::new(sval))
+    }
+}
+
+#[cfg_attr(feature = "wasm", wasm_bindgen(getter_with_clone))]
+#[cfg_attr(feature = "ruby", magnus::wrap(class = "TestReport"))]
+#[derive(Debug, Clone, PartialEq)]
+pub struct MutTestReport(RefCell<TestReport>);
+
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
+impl MutTestReport {
+    pub fn new(origin: String) -> Self {
+        let mut test_result = TestResult::default();
+        test_result.uploader_metadata = Some(UploaderMetadata {
+            origin,
+            version: std::env!("CARGO_PKG_VERSION").to_string(),
+            upload_time: None,
+        });
+        Self(RefCell::new(TestReport {
+            test_result: TestResult::default(),
+        }))
+    }
+
+    fn serialize_test_result(&self) -> Vec<u8> {
+        let test_result = self.0.borrow().test_result.clone();
+        let buf: Vec<u8> = prost::Message::encode_to_vec(&test_result);
+        buf
+    }
+
+    // sends out to the trunk api
+    pub fn publish(&self, repo_root: String) -> bool {
+        let path = self.save();
+        if path.is_err() {
+            return false;
+        }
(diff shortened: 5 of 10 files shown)


Cargo.toml
```diff
@@ -12,6 +12,7 @@ members = [
   "smoke-test",
   "test_utils",
   "xcresult",
+  "test_report",
```
gnalh, line 15: This is a separate library (outside of context) to prevent circular dependencies. `cli` depends on `context` and this depends on `cli`.

cli/src/upload.rs
```diff
@@ -113,6 +113,37 @@ pub struct UploadArgs {
     pub allow_empty_test_results: bool,
 }
 
+impl UploadArgs {
```
gnalh, line 116: Add this constructor to simplify creating the upload args.

proto/src/build.rs
```diff
@@ -2,6 +2,10 @@ use std::io::Result;
 
 fn main() -> Result<()> {
     std::env::set_var("PROTOC", protobuf_src::protoc());
-    prost_build::compile_protos(&["proto/test_context.proto"], &["proto/"])?;
+    let mut prost_build = prost_build::Config::new();
+    prost_build
+        .type_attribute(".", "#[derive(serde::Serialize,serde::Deserialize)]")
+        .extern_path(".google.protobuf.Timestamp", "::prost_wkt_types::Timestamp")
+        .compile_protos(&["proto/test_context.proto"], &["proto/"])?;
```
gnalh, line 9: This adds serialization / deserialization support. `Prost` recommends to override the google types with their own in order to use serialization.

test_report/src/report.rs
```diff
@@ -0,0 +1,268 @@
...
+#[cfg(feature = "wasm")]
+use wasm_bindgen::prelude::wasm_bindgen;
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct TestReport {
+    test_result: TestResult,
+}
+
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
+#[cfg_attr(feature = "ruby", magnus::wrap(class = "TestExecution"))]
```
gnalh, line 16: Since we are only supporting rspec and jest (possibly vitest as well) to start, we only need wasm and ruby support.

test_report/src/report.rs
```diff
@@ -0,0 +1,268 @@
...
+            println!("Token or org url slug not set");
+            return false;
+        }
+        if let Some(uploader_metadata) = &mut self.0.borrow_mut().test_result.uploader_metadata {
+            uploader_metadata.upload_time = Some(Timestamp {
+                seconds: Utc::now().timestamp(),
+                nanos: Utc::now().timestamp_subsec_nanos() as i32,
+            });
+        }
+        let upload_args = trunk_analytics_cli::upload::UploadArgs::new(
```
gnalh, line 128: This is where the magic happens. We invoke all of the same upload logic as the cli to reduce fragmentation.

test_report/src/report.rs
```diff
@@ -0,0 +1,268 @@
...
+        name: String,
+        classname: String,
+        file: String,
+        parent_name: String,
+        line: Option<i32>,
+        status: Status,
+        attempt_number: i32,
+        started_at: i64,
+        finished_at: i64,
+        output: String,
```
gnalh, line 179: Normally I'd prefer to use an object here, but `Magnus` was making this difficult with the optional fields. Using the values directly reduces boilerplate needed and makes it easier to use in ruby.

test_report/src/report.rs
```diff
@@ -0,0 +1,268 @@
...
+    }
+}
+
+#[cfg_attr(feature = "wasm", wasm_bindgen(getter_with_clone))]
+#[cfg_attr(feature = "ruby", magnus::wrap(class = "TestReport"))]
+#[derive(Debug, Clone, PartialEq)]
+pub struct MutTestReport(RefCell<TestReport>);
+
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
+impl MutTestReport {
```
gnalh, line 90: Constructed a separate mutable test report as per the suggestions of `magnus`. It reduces the boilerplate needed because we do not need to have bindings for all of the stored types (the proto object in this case).

test_report/src/report.rs
```diff
@@ -0,0 +1,268 @@
...
+
+    // validates the env is set for CI
+    pub fn valid_env(&self) {
+        todo!("Valid env");
+    }
+
+    // validates that we are in a git repo
+    pub fn valid_git(&self) {
+        todo!("Valid git");
+    }
```
gnalh, line 237: There's are all to be implemented at a later date. I can remove them for now if we don't want them as is in the repo.

test_report/tests/report.rs
```diff
@@ -0,0 +1,97 @@
...
+use test_utils::mock_git_repo::setup_repo_with_commit;
+use test_utils::mock_server::MockServerBuilder;
+
+#[tokio::test(flavor = "multi_thread")]
+async fn publish_test_report() {
+    let temp_dir = tempdir().unwrap();
+    let repo_setup_res = setup_repo_with_commit(&temp_dir);
+    assert!(repo_setup_res.is_ok());
+    let set_current_dir_res = std::env::set_current_dir(&temp_dir);
+    assert!(set_current_dir_res.is_ok());
```
dfrankland, line 15: nit: asserting on `is_ok` won't give a track as to what went wrong—I think it's preferred to `unwrap()` in tests for this purpose

test_report/tests/report.rs
```diff
@@ -0,0 +1,97 @@
...
+            0,
+            0,
+            0,
+            "test-message".into(),
+        );
+        let result = test_report.publish(".".into());
+        assert_eq!(result, true);
+    });
+    let thread_res = thread_join_handle.join();
+    assert!(thread_res.is_ok());
```
dfrankland, line 42: Is there a particular reason to do this in a separate thread?

cli/src/upload.rs
```diff
@@ -113,6 +113,37 @@ pub struct UploadArgs {
...
+            repo_url: None,
+            repo_head_sha: None,
+            repo_head_branch: None,
+            repo_head_commit_epoch: None,
+            tags: Vec::new(),
+            print_files: false,
+            dry_run: false,
+            team: None,
+            // TODO - how are code owners handled?
+            codeowners_path: None,
```
TylerJang27, line 140: What are you missing here primarily?

test_report/tests/report.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+
+#[test]
+fn save_test_report() {
+    let temp_dir = tempdir().unwrap();
+    setup_repo_with_commit(&temp_dir).unwrap();
+    let _ = std::env::set_current_dir(&temp_dir);
+    let test_report = MutTestReport::new("test".into());
```
dfrankland, line 50: IMO, we should inspect the bundle that was uploaded in the first test to be more E2E than to look at the intermediary data like this test does

proto/proto/test_context.proto
```diff
@@ -4,5 +4,35 @@ import "google/protobuf/timestamp.proto";
...
+  TEST_CASE_RUN_STATUS_FAILURE = 2;
+  TEST_CASE_RUN_STATUS_SKIPPED = 3;
+}
+
 message TestCaseRun {
+  string id = 1;
+  string name = 2;
+  string classname = 3;
+  string file = 4;
+  // need parent_id
```
TylerJang27, line 19: What's the plan? Populate it on the way out? Add it to the proto?

dfrankland (approved):

proto/proto/test_context.proto
```diff
@@ -4,5 +4,35 @@ import "google/protobuf/timestamp.proto";
...
+  string parent_name = 5;
+  int32 line = 6;
+  TestCaseRunStatus status = 7;
+  int32 attempt_number = 8;
+  google.protobuf.Timestamp started_at = 9;
+  google.protobuf.Timestamp finished_at = 10;
+  string status_output_message = 11;
```
TylerJang27, line 26: Should more of these be optional? I know a lot of test reporters aren't going to report some of these

proto/proto/test_context.proto
```diff
@@ -4,5 +4,35 @@ import "google/protobuf/timestamp.proto";
...
+}
+
+message UploaderMetadata {
+    string version = 1;
+    string origin = 2; // RSpec, jest, etc..
```
TylerJang27, line 31: Should this include version info too?

proto/src/build.rs
TylerJang27: Had to do the same for the bazel-bep stuff. I started work to try to unify these more but hit a bit of a wall: https://github.com/trunk-io/analytics-cli/pull/208

test_report/src/report.rs
```diff
@@ -0,0 +1,247 @@
...
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct TestReport {
+    test_result: TestResult,
+}
+
+#[cfg_attr(feature = "wasm", wasm_bindgen)]
+#[cfg_attr(feature = "ruby", magnus::wrap(class = "TestExecution"))]
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TestExecution {
```
TylerJang27, line 18: This is 1:1 with TestCaseRun in proto, right? Any reason we can't unify/codegen this

test_report/src/report.rs
```diff
@@ -0,0 +1,247 @@
...
+
+    // adds a test to the test report
+    pub fn add_test(
+        &self,
+        id: Option<String>,
+        name: String,
+        classname: String,
+        file: String,
+        parent_name: String,
+        line: Option<i32>,
```
TylerJang27, line 174: I think the optionality of this field is the only discrepancy with the proto. Should the proto field be optional?

test_report/src/report.rs
TylerJang27: Am I missing something? Don't we need the upload.rs logic to be able to handle `.bin` files somewhere?

TylerJang27 (approved):
Looks good, with some small proto design notes/questions

test_report/src/report.rs
```diff
@@ -0,0 +1,247 @@
...
+}
+
+#[cfg(feature = "ruby")]
+impl Status {
+    fn new(status: String) -> Self {
+        match status.as_str() {
+            "success" => Status::Success,
+            "failure" => Status::Failure,
+            "skipped" => Status::Skipped,
+            _ => Status::Success,
```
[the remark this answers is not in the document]
gnalh: Agreed, but `magnus` makes this a real pain. We can `panic` and it would be handled upstream in ruby's case but I'm not sure how this will be handled by other integrations.

test_report/src/report.rs
```diff
@@ -0,0 +1,247 @@
...
+                None,
+                None,
+                None,
+            )) {
+            Ok(_) => return true,
+            Err(e) => {
+                println!("Error uploading: {:?}", e);
+                return false;
+            }
+        }
```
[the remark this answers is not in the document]
gnalh: Yeah I want to add a wrapped error type in the future, but haven't had a chance to go down that rabbit hole with mangos. This is just the initial implementation.

test_report/src/report.rs
```diff
@@ -0,0 +1,247 @@
...
+    pub fn save(&self) -> Result<String, anyhow::Error> {
+        let buf = self.serialize_test_result();
+        if let Ok(named_temp_file) = tempfile::Builder::new().suffix(".bin").tempfile() {
+            std::fs::write(&named_temp_file, buf).unwrap_or_default();
+            let (_, path) = named_temp_file.keep().unwrap();
+            let path_str = path.to_str().unwrap_or_default();
+            return Ok(path_str.to_string());
+        }
+        Err(anyhow::anyhow!("Error saving test report"))
+    }
```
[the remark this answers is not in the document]
gnalh: internal only

test_report/src/report.rs
```diff
@@ -0,0 +1,247 @@
...
+    pub fn to_string(&self) -> String {
+        self.clone().into()
+    }
+}
+
+impl Into<String> for MutTestReport {
+    fn into(self) -> String {
+        serde_json::to_string(&self.0.borrow().test_result).unwrap_or_default()
+    }
+}
```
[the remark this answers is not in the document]
gnalh: I think I'm missing something, where is the cloning happening here?

test_report/tests/report.rs
gnalh: Yes, limitations of tokio. We create a runtime to run the upload sync.

```
Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.
```

cli/src/upload.rs
gnalh: Main question is how are we grabbing code owners when running this way. I don't know what the shape of it is going to look like.

proto/proto/test_context.proto
gnalh: Not sure I understand the question here.

proto/proto/test_context.proto
gnalh: Proto makes everything optional by default, but

proto/proto/test_context.proto
gnalh: Doesn't it above?

test_report/src/report.rs
gnalh: It isn't 1:1. The bindings also make that tricky.

test_report/src/report.rs
gnalh: All proto fields are optional. There isn't a such thing as a required field in proto3.

Cargo.toml
gnalh: Roger Roger. Wanted to make sure that the upload reuse was fine before committing to splitting out.

test_report/src/report.rs
```diff
@@ -0,0 +1,247 @@
...
+            upload_time: None,
+        });
+        Self(RefCell::new(TestReport {
+            test_result: TestResult::default(),
+        }))
+    }
+
+    fn serialize_test_result(&self) -> Vec<u8> {
+        let test_result = self.0.borrow().test_result.clone();
+        let buf: Vec<u8> = prost::Message::encode_to_vec(&test_result);
```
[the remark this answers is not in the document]
gnalh: Good catch, not needed.

proto/proto/test_context.proto
TylerJang27: Nevermind, I parsed version there as a schema version, I was referring to origin version (Rspec plugin 1.0.0, etc.). This should be fine

proto/proto/test_context.proto
TylerJang27: Can you rephrase this comment? Is it a todo or is it justifying `parent_name`?

test_report/src/report.rs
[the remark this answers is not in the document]
dfrankland: I love this typo 😂

test_report/src/report.rs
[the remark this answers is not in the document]
dfrankland: Just above:
https://github.com/trunk-io/analytics-cli/pull/253/files/f0499d7d2757674d16788b74154bba7684de09e7#diff-7d7066e2bdfa3ff5d7cd3b034375972f61028c6174f47dc7c425506d7ca2785fR219-R221

```rust
    pub fn to_string(&self) -> String {
        self.clone().into()
    }
```

test_report/tests/report.rs
dfrankland: I think you need to spin up Tokio on the current thread in the application logic instead of a full on executor. You can see an example of this here:
https://github.com/trunk-io/analytics-cli/blob/0f0b5d7b608ec03caad873c656f7eb373726ba1d/context-py/src/lib.rs#L136-L142

gnalh: Unfortunately using `new_current_thread` does not remove the requirement of doing this work in a separate thread. For the reasons you mentioned, I will still update it to use that.

dfrankland: Oh darn. I fear this will leak implementation :/ I wonder if we should just spawn a thread in the application code then

test_report/src/report.rs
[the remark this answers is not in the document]
gnalh: autocorrect strikes again

Cargo.toml
gnalh: Going to split this out in a separate PR.

test_report/tests/report.rs
gnalh: Done.

PR #254, opened 2024-12-19
[runner integration - 2] Update ETL to handle parsing custom test reports

gnalh:
Adds initial support so that we can parse the custom test reports within the ETL context. This primarily means adding support for the `.bin` files we'll be generating.

See
* https://github.com/trunk-io/analytics-cli/pull/253
* https://github.com/trunk-io/analytics-cli/pull/255


context/src/junit/validator.rs
```diff
@@ -607,12 +607,12 @@ pub enum JunitTestCaseValidationIssueSubOptimal {
         TIMESTAMP_STALE_HOURS
     )]
     TestCaseStaleTimestamp(DateTime<FixedOffset>),
+    #[error("test case id is not a valid uuidv5")]
+    TestCaseInvalidId(String),
```
gnalh, line 611: Going forward, this won't be a real issue. I want to remove it before release, but will leave it as suboptimal for tracking.

context/src/junit/bindings.rs
```diff
@@ -31,6 +32,178 @@ pub struct BindingsReport {
...
+                    system_err: None,
+                    extra: HashMap::new(),
+                }
+            })
+            .collect();
+        let (name, timestamp, timestamp_micros) = match uploader_metadata {
+            Some(t) => (
+                t.origin,
+                Some(t.upload_time.unwrap_or_default().seconds),
+                Some(t.upload_time.unwrap_or_default().nanos as i64 * 1000),
```
TylerJang27, line 110: millis * 1000?

context/src/junit/bindings.rs
```diff
@@ -31,6 +32,178 @@ pub struct BindingsReport {
...
+            Some(started_at) => (
+                Some(started_at.seconds),
+                Some(started_at.nanos as i64 * 1000),
+            ),
+            None => (None, None),
+        };
+        let time = match (started_at, finished_at) {
+            (Some(started_at), Some(finished_at)) => Some(
+                (finished_at.seconds - started_at.seconds) as f64
+                    + (finished_at.nanos - started_at.nanos) as f64 / 1_000_000_000.0,
```
TylerJang27, line 154: Is the intention here to just handle when one is set but the other is not? Wouldn't this double the actual time if both are set? Feels brittle imo

context/src/junit/validator.rs
TylerJang27: Do we have any times where the id is provided to us in the junit to begin with? I asusme we only generate this when unset for a reason...

bundle/src/files.rs
```diff
@@ -126,10 +126,15 @@ impl FileSet {
 
             // Save file under junit/0, junit/1, etc.
             // This is to avoid having to deal with potential file name collisions.
+            let path_formatted = match original_path_abs.ends_with(".xml") {
+                // we currently support junit and internal binary files
+                true => format!("junit/{}", file_counter.count_file()),
+                false => format!("bin/{}", file_counter.count_file()),
```
TylerJang27, line 132: I was of the impression we would have to do a lot more than this PR does to support a new file format, i.e. handle it in `extract_failed_tests`, etc.

TylerJang27 (commented):
Some notes. Deferring review until discussion and/or previous PR is stabilized

context/src/junit/bindings.rs
gnalh: 1000 microseconds is a nanosecond

context/src/junit/bindings.rs
gnalh: No? This is to generate the duration.

bundle/src/files.rs
gnalh: There's a change in the mono repo. But it wasn't that much to get started.

context/src/junit/bindings.rs
TylerJang27: Unless I'm blind, I'm reading this as

`timestamp_micros = upload_time.nanos * 1000` (instead of / 1000). As written this would give you picoseconds

context/src/junit/bindings.rs
TylerJang27: If
`finished_at.seconds = 2`
`started_at.seconds = 1`
`finished_at.nanos = 2_000_000_000`
`started_at.nanos = 1_000_000_000`

this would give you a duration of `2` seconds, would it not?

context/src/junit/bindings.rs
```diff
@@ -31,6 +32,178 @@ pub struct BindingsReport {
...
+            uploader_metadata,
+        }: TestResult,
+    ) -> Self {
+        let test_cases: Vec<BindingsTestCase> = test_case_runs
+            .into_iter()
+            .map(|testcase| BindingsTestCase::from(testcase))
+            .collect();
+        let parent_name_map: HashMap<String, Vec<BindingsTestCase>> =
+            test_cases.iter().fold(HashMap::new(), |mut acc, testcase| {
+                let parent_name = testcase.extra.get("parent_name").unwrap();
```
dfrankland, line 59: This will panic—perhaps we should implement `TryFrom` or we can check this before pushing an entry into the hashmap?

context/src/junit/bindings.rs
```diff
@@ -31,6 +32,178 @@ pub struct BindingsReport {
...
+            None => (None, None),
+        };
+        let time = match (started_at, finished_at) {
+            (Some(started_at), Some(finished_at)) => Some(
+                (finished_at.seconds - started_at.seconds) as f64
+                    + (finished_at.nanos - started_at.nanos) as f64 / 1_000_000_000.0,
+            ),
+            _ => None,
+        };
+        let typed_status = TestCaseRunStatus::try_from(status).ok().unwrap();
```
dfrankland, line 158: This will panic—should we implement `TryFrom` here?

context/src/junit/bindings.rs
```diff
@@ -31,6 +32,178 @@ pub struct BindingsReport {
...
+            timestamp,
+            timestamp_micros,
+            time,
+            status: BindingsTestCaseStatus {
+                status: typed_status.into(),
+                success: {
+                    match typed_status == TestCaseRunStatus::Success {
+                        true => Some(BindingsTestCaseStatusSuccess { flaky_runs: vec![] }),
+                        false => None,
+                    }
```
dfrankland, line 172: nit: use `if...else` for more idiomatic control flow. Same goes for below

context/src/junit/bindings.rs
```diff
@@ -730,3 +903,117 @@ impl BindingsJunitReportValidation {
...
+    use prost_wkt_types::Timestamp;
+    let mut test1 = TestCaseRun::default();
+    test1.id = "test_id1".into();
+    test1.name = "test_name".into();
+    test1.classname = "test_classname".into();
+    test1.file = "test_file".into();
+    test1.parent_name = "test_parent_name1".into();
+    test1.line = 1;
+    test1.status = TestCaseRunStatus::Success.into();
+    test1.attempt_number = 1;
```
dfrankland, line 919: optional: It might save you some characters to use the partial spread. Same goes for other instances of this

```suggestion
    let mut test1 = TestCaseRun {
        id: "test_id1".into(),
        name: "test_name".into(),
        classname: "test_classname".into(),
        file: "test_file".into(),
        parent_name: "test_parent_name1".into(),
        line: 1,
        status: TestCaseRunStatus::Success.into(),
        attempt_number: 1,
        ..Default::default()
    };
```

context/src/junit/bindings.rs
```diff
@@ -31,6 +32,178 @@ pub struct BindingsReport {
...
+impl From<TestCaseRunStatus> for BindingsTestCaseStatusStatus {
+    fn from(value: TestCaseRunStatus) -> Self {
+        match value {
+            TestCaseRunStatus::Success => BindingsTestCaseStatusStatus::Success,
+            TestCaseRunStatus::Failure => BindingsTestCaseStatusStatus::NonSuccess,
+            TestCaseRunStatus::Skipped => BindingsTestCaseStatusStatus::Skipped,
+            TestCaseRunStatus::Unspecified => todo!(),
+        }
+    }
+}
```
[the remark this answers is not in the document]
gnalh: It's `todo` because I'm planning to address it later. I can resolve it now if you prefer.

context/src/junit/bindings.rs
gnalh: Good catch, didn't mean to leave the unwrap in.

context/src/junit/bindings.rs
gnalh: `nanos` captures the nanos relative to the second. We can enforce that it's never a full second, but that's how it's supposed to work with google timestamp .

context/src/junit/validator.rs
gnalh: Yes, there's a few test runners that set an id field.

context/src/junit/bindings.rs
TylerJang27: Ah, didn't know the relative part

context-py/src/lib.rs
```diff
@@ -6,6 +6,7 @@ use bundle::{
 };
 use codeowners::{BindingsOwners, CodeOwners};
 use context::{env, junit, meta, repo};
+use prost::Message;
```
dfrankland, line 9: My eyes might be deceiving me, but is this trait used anywhere?

context/src/junit/bindings.rs
```diff
@@ -31,6 +32,193 @@ pub struct BindingsReport {
...
+impl From<TestResult> for BindingsReport {
+    fn from(
+        TestResult {
+            test_case_runs,
+            uploader_metadata,
+        }: TestResult,
+    ) -> Self {
+        let test_cases: Vec<BindingsTestCase> = test_case_runs
+            .into_iter()
+            .map(|testcase| BindingsTestCase::from(testcase))
```
dfrankland, line 55: totally optional: this can be simplified to
```suggestion
            .map(BindingsTestCase::from)
```

context/src/junit/bindings.rs
```diff
@@ -31,6 +32,193 @@ pub struct BindingsReport {
...
+            id,
+            file,
+            line,
+            attempt_number,
+        }: TestCaseRun,
+    ) -> Self {
+        let (timestamp, timestamp_micros) = match started_at {
+            Some(started_at) => (
+                Some(started_at.seconds),
+                Some(started_at.nanos as i64 * 1000),
```
dfrankland, line 154: nit: we can use `chrono` for this too

context/src/junit/bindings.rs
```diff
@@ -31,6 +32,193 @@ pub struct BindingsReport {
...
+            ),
+            None => (None, None),
+        };
+        let time = match (started_at, finished_at) {
+            (Some(started_at), Some(finished_at)) => Some(
+                (finished_at.seconds - started_at.seconds) as f64
+                    + (finished_at.nanos - started_at.nanos) as f64 / 1_000_000_000.0,
+            ),
+            _ => None,
+        };
```
dfrankland, line 164: nit: `chrono` can be used for this too. Any reason why not to use `prost-wkt` / `prost-wkt-types` to convert proto `Timestamp` to `chrono` for querying and manipulation? Seems like it would make a lot of these operations much easier:
https://docs.rs/prost-wkt-types/latest/prost_wkt_types/struct.Timestamp.html#method.from-2

dfrankland (approved):

context-py/src/lib.rs
gnalh: It's necessary for decode to be used.
```
no function or associated item named `decode` found for struct `proto::test_context::test_run::TestResult` in the current scope items from traits can only be used if the trait is in scope
```

PR #255, opened 2024-12-19
[runner integration - 3] Add initial rspec plugin

gnalh:
This adds the initial logic for integrating rspec with our custom test reporters. This will handle generating the report and publishing it to our services.

See
* https://github.com/trunk-io/analytics-cli/pull/253
* https://github.com/trunk-io/analytics-cli/pull/254

Follow-up (in progress):
* add upload invocation in the workflow
* add in missing test count parsing during upload (this is visible on the uploads tab)
* direct quarantining 
* js test runner integration


TylerJang27 (commented):
Don't know ruby but on initial glance LGTM

dfrankland (approved):
LOOKING GOOD!!!!

gnalh:
moved to protox for this PR because for some reason it would fail because `protoc` was missing when building the CLI target. What's odd is that context-ruby is supposed to be built separately and this PR was a change purely to that.

PR #256, opened 2024-12-20
no cmake, no protoc, no problems

dfrankland:
* Simplify setup and build by removing reliance on `cmake` and `protoc`
* [Fixes `Cargo.lock` issue](https://trunk-io.slack.com/archives/C02CMS4TFUJ/p1734464030801369) seen in `context-ruby`
* Fix running Clippy
* Fix running `rust-analyzer`
* Fix `cargo check` lints
* Fix imports of dependencies with inconsistent versions

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -17,4 +17,4 @@ use_trunk.io() {
 use trunk.io
 
 # Needed by `rust-analyzer`. Hash should be the same for all platforms.
-export RUST_SRC_PATH="$HOME/.cache/trunk/tools/rust/2024-11-15-41630b9ccc37d4c3f29f3f2c80409ef9/rust-src-nightly/rust-src/lib/rustlib/src/rust/library"
+export RUST_SRC_PATH="$HOME/.cache/trunk/tools/rust/2024-05-01-d4daa7f4db5f9d74e00218df573de113/rust-src-nightly/rust-src/lib/rustlib/src/rust/library"
--- a/.envrc
+++ b/.envrc
@@ -15,3 +15,6 @@ use_trunk.io() {
 }
 
 use trunk.io
+
+# Needed by `rust-analyzer`. Hash should be the same for all platforms.
+export RUST_SRC_PATH="$HOME/.cache/trunk/tools/rust/2024-11-15-41630b9ccc37d4c3f29f3f2c80409ef9/rust-src-nightly/rust-src/lib/rustlib/src/rust/library"
--- a/.envrc
+++ b/.envrc
@@ -17,4 +17,4 @@ use_trunk.io() {
 use trunk.io
 
 # Needed by `rust-analyzer`. Hash should be the same for all platforms.
-export RUST_SRC_PATH="$HOME/.cache/trunk/tools/rust/2024-05-01-d4daa7f4db5f9d74e00218df573de113/rust-src-nightly/rust-src/lib/rustlib/src/rust/library"
+export RUST_SRC_PATH="$HOME/.cache/trunk/tools/rust-src/2024-05-01-d4daa7f4db5f9d74e00218df573de113/rust-src-nightly/rust-src/lib/rustlib/src/rust/library"
--- a/.trunk/trunk.yaml
+++ b/.trunk/trunk.yaml
@@ -97,20 +97,6 @@ actions:
     - trunk-fmt-pre-commit
     - trunk-upgrade-available
 downloads:
-  - name: protoc
-    downloads:
-      - os:
-          linux: linux
-          macos: osx
-        cpu:
-          x86_64: x86_64
-          arm_64: aarch_64
-        url: https://github.com/protocolbuffers/protobuf/releases/download/v${version}/protoc-${version}-${os}-${cpu}.zip
-      - os:
-          windows: windows
-        cpu:
-          x86_64: x86_64
-        url: https://github.com/protocolbuffers/protobuf/releases/download/v29.1/protoc-29.1-win64.zip
   - name: rust
     downloads:
       - os:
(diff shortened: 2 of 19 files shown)


.envrc
```diff
@@ -15,3 +15,6 @@ use_trunk.io() {
 }
 
 use trunk.io
+
+# Needed by `rust-analyzer`. Hash should be the same for all platforms.
+export RUST_SRC_PATH="$HOME/.cache/trunk/tools/rust/2024-05-01-d4daa7f4db5f9d74e00218df573de113/rust-src-nightly/rust-src/lib/rustlib/src/rust/library"
```
dfrankland, line 20: There doesn't seem to be a way to have Rust components added via `trunk` other than referring to them like this. Luckily, this isn't platform dependent and the path can be found by `cat "$(which rust-src)"` since `trunk` creates a wrapper script to the actual location.

.trunk/trunk.yaml
```diff
@@ -14,8 +14,8 @@ runtimes:
   enabled:
     - node@18.12.1
     - python@3.10.8
-    - rust@1.81.0
     - ruby@3.1.0
+    - rust@2024-05-01
```
dfrankland, line 18: Rust couldn't be upgraded to a more recent nightly version since it requires a version of `GLIBC` that's newer than the version of Ubuntu our CI runners use.

bazel-bep/src/lib.rs
```diff
@@ -45,6 +45,8 @@ pub mod types {
     }
 }
 
+// NOTE: `build_event_stream(.serde).rs` has some failing clippy lints
+#[allow(clippy::all)]
```
dfrankland, line 49: `trunk.yaml` doesn't seem to be able to ignore these generated files (perhaps because Clippy reports absolute paths?)

cli/src/upload.rs
```diff
@@ -17,8 +17,6 @@ use bundle::{
 use codeowners::CodeOwners;
 use constants::{EXIT_FAILURE, EXIT_SUCCESS};
 #[cfg(target_os = "macos")]
-use context::junit::junit_path::JunitReportStatus;
```
dfrankland, line 20: This was reported from `cargo check` which `trunk check` doesn't support yet and is not reported by Clippy

context/Cargo.toml
```diff
@@ -8,7 +8,6 @@ pretty_assertions = "0.6"
 junit-mock = { path = "../junit-mock" }
 test_utils = { path = "../test_utils" }
 tempfile = "3.2.0"
-prost-wkt-types = "0.6.0"
```
dfrankland, line 11: not needed in `dev-dependencies` since it's a regular dependency

context/Cargo.toml
```diff
@@ -34,7 +33,7 @@ uuid = { version = "1.10.0", features = ["v5"] }
 wasm-bindgen = { version = "0.2.95", optional = true }
 magnus = { version = "0.7.1", optional = true, default-features = false }
 proto = { path = "../proto" }
-prost-wkt-types = "0.6.0"
+prost-wkt-types = { version = "0.5.1", features = ["vendored-protox"] }
```
dfrankland, line 36: Until https://github.com/fdeantoni/prost-wkt/pull/70 lands, `prost-wkt-types` prevents us from having `prost`'s version be consistent which is required for generated code to compile properly

context/src/junit/bindings.rs
```diff
@@ -107,15 +107,18 @@ impl From<TestResult> for BindingsReport {
...
+                (
+                    t.origin,
+                    Some(upload_time.seconds),
+                    Some(
+                        chrono::Duration::nanoseconds(upload_time.nanos as i64)
+                            .num_microseconds()
+                            .unwrap_or_default(),
+                    ),
+                )
+            }
```
dfrankland, line 121: Since `prost-wkt-types` got downgraded, `Timestamp` requires `clone`

context/src/lib.rs
```diff
@@ -1,5 +1,3 @@
-// NOTE: This lint isn't applicable since we compile with nightly
-/* trunk-ignore(clippy/E0554) */
```
dfrankland, line 2: No longer needed now that Clippy is working properly 🤷

proto/src/build.rs
```diff
@@ -1,10 +1,8 @@
...
-    prost_build
+fn main() {
+    let file_descriptors = protox::compile(["proto/test_context.proto"], ["proto/"]).unwrap();
+    prost_build::Config::new()
         .type_attribute(".", "#[derive(serde::Serialize,serde::Deserialize)]")
         .extern_path(".google.protobuf.Timestamp", "::prost_wkt_types::Timestamp")
-        .compile_protos(&["proto/test_context.proto"], &["proto/"])?;
-    Ok(())
+        .compile_fds(file_descriptors)
+        .unwrap();
```
dfrankland, line 7: Changing to protox was serendipitously done at the same time by Gabe. I changed this to be a little more succinctly/consistently by using `unwrap`s here—this is a `build.rs` file which is run prior to compilation, so `unwrap` is fine to use here

bazel-bep/src/lib.rs
TylerJang27: I would expect us to handle this properly by adding a lint.ignore, but fine for now

TylerJang27 (approved):
Look at him go

bazel-bep/src/lib.rs
dfrankland: Exactly my thoughts. I added one and it didn't work
https://github.com/trunk-io/analytics-cli/pull/256/files#diff-45a5a35dd7ad9a52d0fd3540f6c0d16a763575de51cc9a0a6c6f6a9f1da62ecdR78-R83

PR #269, opened 2025-01-02
Part 4 - re-organize upload command

dfrankland:
Attempts to split out functions for ease of reading the upload command logic

### DIFF
--- a/cli/src/upload.rs
+++ b/cli/src/upload.rs
@@ -9,7 +9,8 @@ use std::{
 use api::BundleUploadStatus;
 use bundle::{
     parse_custom_tags, BundleMeta, BundleMetaBaseProps, BundleMetaDebugProps, BundleMetaJunitProps,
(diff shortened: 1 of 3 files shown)


cli/src/upload.rs
```diff
@@ -148,10 +149,130 @@ impl UploadArgs {
...
+        for file_set in &meta.base_props.file_sets {
+            println!(
+                "  File set ({:?}): {}",
+                file_set.file_set_type, file_set.glob
+            );
+            for file in &file_set.files {
+                println!("    {}", file.original_path);
+            }
+        }
+    }
```
gnalh, line 185: When do we want to print over log?

gnalh (approved):
Overall LGTM, just one question

cli/src/upload.rs
dfrankland: All of the logging needs to be redone—this is just old code

PR #272, opened 2025-01-07
Fix TestReport parsing

gnalh:
I accidentally generated the status in the wrong way. This was silently swallowed by the validation when being transformed from a BindingsReport into the quick_junit Report and ultimately caused test cases to be missing during the parsing step of the ETL. This addresses the immediate issue and adds a corresponding test.

Long term, we should migrate away from converting back to junit when doing validation and accept the BindingsReport directly. If that's undesirable then we should not being using try_into [[1](https://github.com/trunk-io/analytics-cli/blob/0fdcc4b6ce3020cf8801c64f6468923a80c54c4e/context/src/junit/bindings.rs#L551), [2](https://github.com/trunk-io/analytics-cli/blob/0fdcc4b6ce3020cf8801c64f6468923a80c54c4e/context-py/src/lib.rs#L92-L98), [3](https://github.com/trunk-io/analytics-cli/blob/0fdcc4b6ce3020cf8801c64f6468923a80c54c4e/context/src/junit/bindings.rs#L406-L413)] when converting the statuses. It needs to be surfaced during validation. This second part is outside the scope of this fix and am happy to discuss further.

### DIFF
--- a/context/src/junit/bindings.rs
+++ b/context/src/junit/bindings.rs
@@ -176,7 +176,7 @@ impl From<TestCaseRun> for BindingsTestCase {
                     }
                 },
                 non_success: {
-                    if typed_status == TestCaseRunStatus::Success {
+                    if typed_status == TestCaseRunStatus::Failure {
                         Some(BindingsTestCaseStatusNonSuccess {
                             kind: BindingsNonSuccessKind::Failure,
                             message: Some(status_output_message.clone()),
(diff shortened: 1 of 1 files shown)


context/src/junit/bindings.rs
```diff
@@ -1023,4 +1024,38 @@ fn parse_test_report_to_bindings() {
...
+            assert_eq!(issue.1.level, JunitValidationLevel::SubOptimal);
+            if issue.0 == 0 {
+                assert_eq!(issue.1.error_type, JunitValidationType::Report);
+                assert_eq!(
+                    issue.1.error_message,
+                    "report has old (> 30 day(s)) timestamps"
+                );
+            } else {
+                assert_eq!(issue.1.error_type, JunitValidationType::TestCase);
+                assert_eq!(issue.1.error_message, "test case id is not a valid uuidv5");
```
gnalh, line 1048: I'm planning on removing this error case before release, but left it in the test as validation for the time being.

TylerJang27 (approved):

PR #276, opened 2025-01-08
Part 1 - refactor file set building and failed test extraction

dfrankland:
This fixes two different abstractions / separation of concerns
* File sets
  * Previous
    * File counter: counts and keeps index for file naming of JUnit files in the bundle
    * Array of file sets
      * scan for JUnit files, gather relevant information, structure information into format for meta JSON all in one go
      * uses / mutates the file counter to uniquely name JUnit files in the bundle
    * Issues with this approach
      * main logic had to keep track of both of these separately everywhere even though they are closely related
      * finding files, counting files, and making metadata were all tightly intertwined making it difficult to figure out where to change logic for one of them
      * logic for checking when we don't find files is duplicated for each command
  * Now
    * File set builder
      * scan for JUnit files
      * then, in a loop for each file
        * try to generate metadata
        * try to count / increment file index
      * Advantages of this approach are
        * easy to follow flow of logic for finding, counting, and generate metadata for JUnit files
        * easy to see close relationship of file counts, checking for no files found, and retrieving file metadata
* Failed test extraction
  * Previous
    * Array of tests: just a filtered array of tests
    * Issues with this approach
      * hard to understand if this was filtered or if this is just an array of all tests
      * cannot associate other logic like exit code based on this
  * Now
    * Failed test extractor: struct which associates result data with logic of filtering and deriving exit code
    * Advantages: easy to see close relationship of data

### DIFF
--- a/bundle/src/files.rs
+++ b/bundle/src/files.rs
@@ -1,12 +1,18 @@
-use std::{format, time::SystemTime};
+use std::{
+    fmt::Debug,
+    format,
+    path::{Path, PathBuf},
+    time::SystemTime,
+};
 
 use codeowners::{CodeOwners, Owners, OwnersOfPath};
 use constants::ALLOW_LIST;
 use context::junit::junit_path::{JunitReportFileWithStatus, JunitReportStatus};
+use glob::glob;
 #[cfg(feature = "pyo3")]
 use pyo3::prelude::*;
 #[cfg(feature = "pyo3")]
-use pyo3_stub_gen::derive::gen_stub_pyclass;
+use pyo3_stub_gen::derive::{gen_stub_pyclass, gen_stub_pyclass_enum};
 use regex::Regex;
 use serde::{Deserialize, Serialize};
 #[cfg(feature = "wasm")]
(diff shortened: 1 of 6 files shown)


bundle/src/files.rs
```diff
@@ -1,35 +1,141 @@
...
+                        }
+                        Ok(acc)
+                    },
+                )?;
+                acc.count = count;
+                acc.file_sets.push(FileSet::new(
+                    bundled_files,
+                    junit_wrapper.junit_path.clone(),
+                    junit_wrapper.status.clone(),
+                ));
```
dfrankland, line 105: simple convert to metadata + counting

bundle/src/files.rs
```diff
@@ -45,125 +151,144 @@ pub struct FileSet {
...
+#[cfg_attr(feature = "pyo3", gen_stub_pyclass, pyclass(get_all))]
+#[cfg_attr(feature = "wasm", derive(Tsify))]
+pub struct BundledFile {
+    pub original_path: String,
+    /// Added in v0.5.33
+    pub original_path_rel: Option<String>,
+    pub path: String,
+    pub owners: Vec<String>,
+    pub team: Option<String>,
+}
```
dfrankland, line 188: Pretty much everything below this is moved from `types.rs` because it's only relevant to the structs and logic in `files.rs`

bundle/src/types.rs
```diff
@@ -43,7 +47,7 @@ impl Test {
         file: Option<String>,
         id: Option<String>,
         org_slug: &str,
-        repo: &BundleRepo,
+        repo: &RepoUrlParts,
```
dfrankland, line 50: We don't need `BundleRepo`—we just referenced `RepoUrlParts` within `BundleRepo` in the nested logic. This cleans up a bunch of code trying to do `repo.repo` or creating the larger `BundleRepo` struct for testing

bundle/src/types.rs
```diff
@@ -94,52 +96,6 @@ pub struct BundleUploader {
...
-    pub team: Option<String>,
-}
-
-impl BundledFile {
-    pub fn get_print_path(&self) -> &str {
-        self.original_path_rel
-            .as_ref()
-            .unwrap_or(&self.original_path)
-    }
-}
```
dfrankland, line 141: moved to `files` since it's only relevant there

cli/src/runner.rs
```diff
@@ -27,15 +28,14 @@ pub enum JunitSpec {
...
-    org_slug: &str,
     command: &String,
     args: Vec<&String>,
     junit_spec: JunitSpec,
     team: Option<String>,
     codeowners: &Option<CodeOwners>,
 ) -> anyhow::Result<RunResult> {
     let start = SystemTime::now();
-    let exit_code = run_test_and_get_exit_code(command, args).await?;
+    let exit_code = run_test_and_get_exit_code(command, args)?;
```
dfrankland, line 38: This function never used `await` within it, so I removed `async` from `fn`

cli/src/runner.rs
```diff
@@ -68,12 +68,12 @@ pub async fn run_test_command(
...
     Ok(RunResult {
         exit_code,
-        failures,
+        file_set_builder,
         exec_start: Some(start),
     })
 }
 
-async fn run_test_and_get_exit_code(command: &String, args: Vec<&String>) -> anyhow::Result<i32> {
+fn run_test_and_get_exit_code(command: &String, args: Vec<&String>) -> anyhow::Result<i32> {
```
dfrankland, line 69: There was no `await` within this function 🤷

cli/src/runner.rs
```diff
@@ -94,58 +94,8 @@ async fn run_test_and_get_exit_code(command: &String, args: Vec<&String>) -> any
...
-                    team.clone(),
-                    codeowners,
-                    exec_start,
-                )
-            })
-            .collect::<anyhow::Result<Vec<FileSet>>>()?;
-    }
-
-    Ok((file_sets, file_counter))
-}
```
dfrankland, line 145: Moved this logic to `FileSetBuilder` in `files.rs`

bundle/src/files.rs
```diff
@@ -1,35 +1,141 @@
...
+                            codeowners.clone(),
+                            exec_start,
+                        )? {
+                            acc.0 += 1;
+                            acc.1.push(bundled_file);
+                        }
+                        Ok(acc)
+                    },
+                )?;
+                acc.count = count;
```
max-trunk, line 100: Should this be `acc.count += count`?

cli/src/upload.rs
```diff
@@ -219,62 +219,51 @@ pub async fn run_upload(
...
                     repo: repo.repo.clone(),
                     org_url_slug: org_url_slug.clone(),
-                    test_identifiers: failures.clone(),
+                    test_identifiers: failed_tests_extractor.failed_tests().to_vec(),
                 },
-                failures,
-                exit_code,
+                &file_set_builder,
+                Some(failed_tests_extractor),
+                None,
```
max-trunk, line 249: Just wanted to sanity check that passing `None` here for the exit code was intentional (without being super familiar with this code path)

max-trunk (commented):
Looks great! just a couple Qs

cli/src/upload.rs
dfrankland: Yep, that exit code is for the test command

bundle/src/files.rs
dfrankland: Nope since we are incrementing the count in the inner fold 👍

max-trunk (approved):

bundle/src/files.rs
```diff
@@ -1,35 +1,141 @@
...
+        if file_set_builder.count == 0 {
+            let junit_paths_with_glob = junit_paths
+                .iter()
+                .cloned()
+                .map(|mut junit_wrapper| {
+                    junit_wrapper.junit_path = PathBuf::from(junit_wrapper.junit_path)
+                        .join("**/*.xml")
+                        .to_string_lossy()
+                        .to_string();
+                    junit_wrapper
```
gnalh, line 55: This should be update to support the `.bin` as well (and not assume xml). I don't believe it is strictly required right now because we don't support uploading those directly from the cli yet, but would be good to add.

bundle/src/files.rs
```diff
@@ -45,125 +151,144 @@ pub struct FileSet {
 }
 
 impl FileSet {
-    /// Scan a file set from a glob path.
-    /// And generates file set using file counter.
-    ///
-    pub fn scan_from_glob(
-        repo_root: &str,
-        glob_with_status: JunitReportFileWithStatus,
```
TylerJang27, line 53: I think the change that you added removes the work done in https://github.com/trunk-io/analytics-cli/pull/242/files#diff-5fba23e112b0ab064ebf51f2e6ce7e7454395e6046cd7142b14f2fe0c8591e0eR157 to populate `resolved_status` (i.e. when coming from BEP) and use it as an override when determining failed_tests.

TylerJang27 (commented):
Concern about `resolved_status` not being correctly forwarded inside `file_sets_from_glob`

bundle/src/files.rs
dfrankland: This is just copied logic from here since this is a refactor:
https://github.com/trunk-io/analytics-cli/pull/276/files/741eda7912d000e6181f3fb6b3342ed7231312e6#diff-3b165d3ba4073981ae2d1c4353a6baaf0d9739f55fad738758350ecedd8a5479L128

I'd rather add new functionality afterwards, I'll create a ticket for you:
https://linear.app/trunk/issue/TRUNK-14012/analytics-cli-add-junit-path-fallback-to-search-for-bin-files

bundle/src/files.rs
```diff
@@ -1,35 +1,141 @@
...
+        let file_set_builder = Self::file_sets_from_glob(
+            repo_root,
+            junit_paths,
+            team.clone(),
+            codeowners,
+            exec_start,
+        )?;
+
+        // Handle case when junit paths are not globs.
+        if file_set_builder.count == 0 {
```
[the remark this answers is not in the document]
dfrankland: Is that an invalid glob? Either way, this is the exact same logic from before:
https://github.com/trunk-io/analytics-cli/pull/276/files/741eda7912d000e6181f3fb6b3342ed7231312e6#diff-3b165d3ba4073981ae2d1c4353a6baaf0d9739f55fad738758350ecedd8a5479L120-L142

bundle/src/files.rs
dfrankland: Isn't this that exact logic?
https://github.com/trunk-io/analytics-cli/pull/276/files#diff-5fba23e112b0ab064ebf51f2e6ce7e7454395e6046cd7142b14f2fe0c8591e0eR101-R105

bundle/src/files.rs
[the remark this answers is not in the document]
TylerJang27: Nvmd, yeah this is the same as before. More I meant invalid in the sense that it returns no files found. 👍

bundle/src/files.rs
TylerJang27: Ah that's it. I can't read, was looking for `resolved_status` instead of `status` 🤦

PR #280, opened 2025-01-10
[TRUNK-13980] bindings for multithreaded codeowners parsing and association

max-trunk:
[TRUNK-13980](https://linear.app/trunk/issue/TRUNK-13980/implement-code-owners-process-parallelization-on-rust-side)

Adds bindings for new pyfunctions intended to be used in the periodic ETL to 'batch' associate codeowners, leveraging Rust's multithreading capabilities, in an attempt to improve the overall performance of the codeowners association step.

Summary of new pyfunctions:
 - `parse_many_codeowners_multithreaded()` -> accepts a batch of `Vec<(bundle_upload_id, codeowners_bytes)>` and returns a `HashMap<bundle_upload_id, CodeOwners obj>`, splitting work over a requested number of threads
 - `associate_codeowners_multithreaded()` -> accepts a `HashMap<bundle_upload_id, CodeOwners obj>` and a `Vec<(bundle_upload_id, file path to associate owners for)>` and returns a `Vec<Vec<associated owner>>`, splitting work over a requested number of threads **and maintaining order of elements in the resulting vector**

See example usage in https://github.com/trunk-io/trunk/pull/20958

Tried my best with the multithreaded implementation but I'm sure it's not the most idiomatic way to do it :) pls hammer away!

### DIFF
--- a/codeowners/src/codeowners.rs
+++ b/codeowners/src/codeowners.rs
@@ -1,6 +1,9 @@
 use std::{
+    collections::HashMap,
     fs::File,
     path::{Path, PathBuf},
+    sync::{Arc, Mutex, RwLock},
+    thread,
 };
 
 use constants::CODEOWNERS_LOCATIONS;
@@ -17,7 +20,13 @@ use wasm_bindgen::prelude::*;
 
 #[cfg(feature = "pyo3")]
 use crate::{github::BindingsGitHubOwners, gitlab::BindingsGitLabOwners};
-use crate::{github::GitHubOwners, gitlab::GitLabOwners, traits::FromReader};
+use crate::{
+    github::GitHubOwners,
+    gitlab::GitLabOwners,
+    traits::{FromReader, OwnersOfPath},
+};
+
+pub type BundleUploadIDAndCodeOwnersBytes = (String, Option<Vec<u8>>);
 
 // TODO(TRUNK-13628): Implement serializing and deserializing for CodeOwners
 #[derive(Default, Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
@@ -86,6 +95,36 @@ impl CodeOwners {
             owners: owners_result.ok(),
         }
     }
+
+    pub fn parse_many_multithreaded(
+        to_parse: Vec<BundleUploadIDAndCodeOwnersBytes>,
+        num_threads: usize,
+    ) -> HashMap<String, Option<Self>> {
+        let chunk_size = (to_parse.len() + num_threads - 1) / num_threads;
+        let mut handles = Vec::with_capacity(num_threads);
+        let results_map: Arc<Mutex<HashMap<String, Option<Self>>>> =
+            Arc::new(Mutex::new(HashMap::new()));
+
+        for chunk in to_parse.chunks(chunk_size) {
+            let chunk = chunk.to_vec();
+            let results_map = Arc::clone(&results_map);
+            let handle = thread::spawn(move || {
+                for (bundle_upload_id, codeowners_bytes) in chunk.into_iter() {
+                    let codeowners = codeowners_bytes.map(Self::parse);
+                    let mut results_map = results_map.lock().unwrap();
+                    results_map.insert(bundle_upload_id, codeowners);
+                }
+            });
+
+            handles.push(handle);
+        }
+
+        for handle in handles {
+            handle.join().unwrap();
+        }
+
+        Arc::try_unwrap(results_map).unwrap().into_inner().unwrap()
+    }
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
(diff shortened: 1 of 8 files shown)


codeowners/src/codeowners.rs
```diff
@@ -135,3 +244,60 @@ where
...
+        let num_codeowners_files = 100;
+        let num_files_to_associate_owners = 1000;
+        let num_threads = 4;
+
+        let codeowners_files: Vec<BundleUploadIDAndCodeOwnersBytes> = (0..num_codeowners_files)
+            .map(|i| (i.to_string(), Some(make_codeowners_bytes(i))))
+            .collect();
+        let to_associate: Vec<(String, Option<String>)> = (0..num_files_to_associate_owners)
+            .map(|i| {
+                let mut file = "foo".to_string();
```
TylerJang27, line 267: Is the intention here for these odd files to not be associated with any codeowners? nit: would rename to "unassociated".to_string() or something

TylerJang27 (commented):
LGTM. Deferring to someone who knows more rust :)

codeowners/src/codeowners.rs
```diff
@@ -85,6 +94,38 @@ impl CodeOwners {
             owners: owners_result.ok(),
         }
     }
+
+    pub fn parse_many_multithreaded(
+        to_parse: Vec<BundleUploadIDAndCodeOwnersBytes>,
+        num_threads: usize,
+    ) -> HashMap<String, Option<Self>> {
```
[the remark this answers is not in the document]
max-trunk: The idea was to give back something that allowed for quick lookups of codeowners objects by bundle upload ID. Mainly had the ETL usage in mind

codeowners/src/codeowners.rs
max-trunk: Yep!

codeowners/src/codeowners.rs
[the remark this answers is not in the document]
dfrankland: I hear you, but this crate is pretty far removed from that logically—ie I think that logic is better handled outside of this function as a matter of separation of concerns. Looking at this logic only in this crate, it's hard to discern its purpose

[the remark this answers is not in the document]
tenesttang-trunk: Just to verify -- we can guarantee that the order of parsed CodeOwners objects is the same as the input codeowners bytes?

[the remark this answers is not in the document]
max-trunk: Yep!

codeowners/src/codeowners.rs
```diff
@@ -85,6 +94,38 @@ impl CodeOwners {
...
+                    results_map.insert(bundle_upload_id, codeowners);
+                }
+            });
+
+            handles.push(handle);
+        }
+
+        for handle in handles {
+            handle.join().unwrap();
+        }
```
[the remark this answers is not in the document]
max-trunk: Alrighty - refactored to use tokio. Suprise suprise, much cleaner :D

codeowners/src/codeowners.rs
```diff
@@ -121,6 +155,58 @@ impl BindingsOwners {
...
+            Owners::GitLabOwners(glo) => glo
+                .of(file)
+                .unwrap_or_default()
+                .iter()
+                .map(ToString::to_string)
+                .collect(),
+        },
+        _ => Vec::new(),
+    }
+}
```
dfrankland, line 181: Same comment as before, I don't think that `BundleId` is related to codeowners, so we shouldn't have logic here related to it here. This method can really get simplified to the following:

```suggestion
fn associate_codeowners<T: AsRef<str>>(
    owners: &Owners,
    file: T,
) -> Vec<String> {
    match owners {
        Owners::GitHubOwners(gho) => gho
            .of(file)
            .unwrap_or_default()
            .iter()
            .map(ToString::to_string)
            .collect(),
        Owners::GitLabOwners(glo) => glo
            .of(file)
            .unwrap_or_default()
            .iter()
            .map(ToString::to_string)
            .collect(),
    }
}```

File: codeowners/src/codeowners.rs [hunk 5]
```diff
@@ -121,6 +155,58 @@ impl BindingsOwners {
...
+        tasks.push(task);
+    }
+
+    for task in tasks {
+        let (i, result) = task.await?;
+        results[i] = Some(result);
+    }
+
+    Ok(results.into_iter().flatten().collect())
+}
```
dfrankland, line 208: Very similar feedback to the above. `BundleId` isn't related to codeowners, so we shouldn't have any logic related to it here:

```suggestion
pub async fn associate_codeowners_multithreaded<T: AsRef<Path>>(
    to_associate: Vec<(Owners, T)>,
) -> Result<Vec<Vec<String>>> {
    let tasks = to_associate.iter().enumerate().map(|(i, (owners, file))| {
        task::spawn(async move {
            (i, associate_codeowners(&owners, file))
        })
    }).collect::<Vec<_>>();

    let mut results = Vec::with_capacity(task.len());
    for task in tasks {
        let (i, result) = task.await?;
        results[i] = result;
    }

    Ok(results)
}
```

context-py/src/lib.rs
```diff
@@ -178,13 +181,98 @@ fn meta_validation_level_to_string(
...
+#[gen_stub_pyfunction]
+#[pyfunction]
+fn parse_many_codeowners_multithreaded(
+    to_parse: Vec<Option<Vec<u8>>>,
+) -> PyResult<Vec<Option<BindingsOwners>>> {
+    let rt = tokio::runtime::Builder::new_multi_thread()
+        .enable_all()
+        .build()?;
+    parse_many_codeowners_multithreaded_impl(rt, to_parse)
+}
```
dfrankland, line 216: Can we combine these by making `num_threads` `Option<usize>`? Same with the other similar function below

context-py/src/lib.rs
```diff
@@ -178,13 +181,98 @@ fn meta_validation_level_to_string(
...
+    to_parse: Vec<Option<Vec<u8>>>,
+) -> PyResult<Vec<Option<BindingsOwners>>> {
+    let parsed_codeowners = rt
+        .block_on(CodeOwners::parse_many_multithreaded(to_parse))
+        .map_err(|err| PyTypeError::new_err(err.to_string()))?;
+    Ok(parsed_codeowners
+        .into_iter()
+        .map(|codeowners| codeowners.and_then(|codeowners| codeowners.owners.map(BindingsOwners)))
+        .collect())
+}
```
dfrankland, line 229: At this top-level is where I would implement the logic related to how the data is passed from elsewhere. It can follow the very familiar pattern that we've seen in the other `codeowners` crate. Same goes for `associate_codeowners_multithreaded_impl` below

```suggestion
fn parse_many_codeowners_multithreaded_impl(
    rt: tokio::runtime::Runtime,
    to_parse: Vec<Option<Vec<u8>>>,
) -> PyResult<Vec<Option<BindingsOwners>>> {
    let to_parse_len = to_parse.len();
    let parsed_indexes = to_parse
        .iter()
        .enumerate()
        .filter_map(|(i, bytes)| -> Option<usize> {
            bytes.map(|| i)
        })
        .collect::<Vec<_>>();
    let parsed_codeowners = rt
        .block_on(
            CodeOwners::parse_many_multithreaded(
                to_parse
                    .into_iter()
                    .filter_map(|tp| tp)
                    .collect()
            )
        )
        .map_err(|err| PyTypeError::new_err(err.to_string()))?;

    let mut results: Vec<Option<BindingsOwners>> = vec![None; to_parse_len];
    for (i, codeowners) in parsed_codeowners.into_iter().enumerate() {
        results[parsed_indexes[i]] = Some(BindingsOwners(codeowners.owners));
    }
    Ok(results)
}
```

context-py/src/lib.rs
max-trunk: This is what I initially tried, but was unsuccessful. All function pyo3 function arguments are required by default, but can be configured optional via the `#[pyo3(signature = (...))]` macro https://pyo3.rs/main/function/signature. This part worked; but on the Python consumer side the function arguments are `Unknown` and not typed. I couldn't seem to declare argument types with the `#[pyo3(signature = (...))]` macro and have them carry over to the generated bindings, so this is what I settled for

context-py/src/lib.rs
max-trunk: Implemented your suggestions! Thanks again

dfrankland (approved):

PR #288, opened 2025-01-13
Part 6 - simplify and abstract gathering context for upload

dfrankland:
Abstract and reuse functions as much as possible to simplify test, upload, and quarantine. Also helps prevent a lot of duplicate execution.


cli/src/context.rs
```diff
@@ -0,0 +1,358 @@
...
+    let envs: HashMap<String, String> = ENVS_TO_GET
+        .iter()
+        .filter_map(|&env_var| {
+            env::var(env_var)
+                .map(|env_var_value| (env_var.to_string(), env_var_value))
+                .ok()
+        })
+        .collect();
+
+    let meta = BundleMeta {
```
TylerJang27, line 93: 🙏

cli/src/context.rs
```diff
@@ -0,0 +1,358 @@
...
+                env!("CARGO_PKG_VERSION"),
+                env!("VERGEN_GIT_SHA"),
+                env!("VERGEN_RUSTC_SEMVER")
+            ),
+            bundle_upload_id: String::with_capacity(0),
+            tags: parse_custom_tags(&tags)?,
+            file_sets: Vec::with_capacity(0),
+            envs,
+            upload_time_epoch: SystemTime::now().duration_since(UNIX_EPOCH)?.as_secs(),
+            test_command: None,
```
TylerJang27, line 117: Are you missing a call to set `test_command`? I don't see it elsewhere...

cli/src/test.rs
```diff
@@ -32,166 +24,48 @@ pub struct TestArgs {
...
-    command: &String,
-    args: Vec<&String>,
-    junit_spec: JunitSpec,
-    team: Option<String>,
-    codeowners: &Option<CodeOwners>,
-) -> anyhow::Result<RunResult> {
-    let start = SystemTime::now();
-    let exit_code = run_test_and_get_exit_code(command, args)?;
-    log::info!("Command exit code: {}", exit_code);
+pub async fn run_test(
```
TylerJang27, line 34: gorgeous

cli/src/test.rs
```diff
@@ -32,166 +24,48 @@ pub struct TestArgs {
...
+pub async fn run_test_command<T: AsRef<str>>(command: &[T]) -> anyhow::Result<TestRunResult> {
+    let exec_start = SystemTime::now();
+    let mut child = Command::new(command.first().map(|s| s.as_ref()).unwrap_or_default())
+        .args(
+            command
+                .iter()
+                .skip(1)
+                .map(|s| s.as_ref())
+                .collect::<Vec<_>>(),
+        )
```
TylerJang27, line 64: What was the reason for this change and the generic?

TylerJang27 (approved):
good shtuff. Couple small callouts

cli/src/test.rs
dfrankland: `&String` is an odd type since that's usually represented by `&str`, but to make whatever calls this even more simplified we can accept anything that can be referenced as `&str` with the generic `AsRef<str>`. It's essentially a bit of clean up.

This is very opinionated, so I'm open to doing things differently. Here's a bit more background:
https://steveklabnik.com/writing/when-should-i-use-string-vs-str/

cli/src/context.rs
dfrankland: I totally did forget this, so I've updated it here:
[`4ea761b` (#288)](https://github.com/trunk-io/analytics-cli/pull/288/commits/4ea761bd29bbcf1c93e6ca8b37e782aba98fd12d#diff-f744e9c93b77a61a62a445a9ac43b43c353d329a0656b033aca02edc89676e5bR169)

And I added a test for that, so we don't mess up again here:
[`4ea761b` (#288)](https://github.com/trunk-io/analytics-cli/pull/288/files#diff-75cb9d34ff9cc8c9e6bdbcea3c7014f3c8f2f96d8885d780b7a989acc6446488R138-R141)

PR #292, opened 2025-01-14
Part 8 - add quarantine command

dfrankland:


cli-tests/src/quarantine.rs
```diff
@@ -0,0 +1,108 @@
...
+            let mut test_ids = get_quarantine_bulk_test_status_request
+                .test_identifiers
+                .into_iter()
+                .map(|t| t.id)
+                .collect::<Vec<_>>();
+            let called_count = CALLED_COUNT.load(Ordering::SeqCst);
+            let quarantined_tests = match called_count {
+                0 => Vec::new(),
+                1 => test_ids.split_off(1),
+                2..=3 => test_ids,
```
TylerJang27, line 42: ngl mixed feelings about this test structure setup. At this scale, it's very easy to see what this test is doing (since you've added the comments below), but I worry that as it scales/drifts, it will become readable. Any thoughts here or other input? I don't have a better solution, so certainly won't block on this, but I think merits mini discussion

cli/src/quarantine_command.rs
```diff
@@ -0,0 +1,24 @@
...
+
+use crate::upload_command::{run_upload, UploadArgs, UploadRunResult};
+
+#[derive(Args, Clone, Debug)]
+pub struct QuarantineArgs {
+    #[command(flatten)]
+    upload_args: UploadArgs,
+}
+
+pub async fn run_quarantine(QuarantineArgs { upload_args }: QuarantineArgs) -> anyhow::Result<i32> {
```
TylerJang27, line 11: optional: add comment that this is basically an alias. I think as `run_quarantine` matures, `run_upload` might start to lose meaning as a "run_upload_and_maybe_quarantine" function

TylerJang27 (approved):
LGTM

cli-tests/src/quarantine.rs
dfrankland: Nah, it's a good comment. I think a much better way would be to use an enum that's stored in this static reference. I'll update it before running each command so that it's much more declarative.

PR #293, opened 2025-01-14
Part 9 - rename dry-run flag to no-upload

dfrankland:


TylerJang27 (approved):
LGTM. Is this flag still used anywhere in practice?

dfrankland:
The plan is to allow people to use quarantining without upload

PR #299, opened 2025-01-15
Part 1 - reorganize xcresult test fixtures

dfrankland:
I did this to clean up a few things, but more importantly to track changes to fixtures in git. I believe this will help with reviews and to track changes to output over time.


xcresult/tests/data/test-ExpectedFailures.junit.xml
dfrankland, line 1: To reiterate the PR description, I think having the XML plain text in the repo will help with reviews and track changes to output overtime.

xcresult/tests/data/test-ExpectedFailures.xcresult.tar.gz
dfrankland, line 1: Archiving the individual `.xcresult` bundles will likely cause less churn since we don't need to update the individual bundles

xcresult/tests/xcresult.rs
```diff
@@ -1,40 +1,44 @@
...
-fn init() {
-    let path = "tests/data.tar.gz";
-    let file = File::open(path).unwrap();
-    let decoder = GzDecoder::new(file);
-    let mut archive = Archive::new(decoder);
-    match archive.unpack(TEMP_DIR.as_ref()) {
-        Ok(_) => (),
-        Err(e) => panic!("failed to unpack data.tar.gz: {}", e),
-    }
-}
```
dfrankland, line 32: I don't think `ctor` gets us any more functionality than what `lazy_static` already allows, which is to lazily initialize anything and stick it in a static reference. Let me know if this is off and there's something more I'm not seeing.

xcresult/Cargo.toml
```diff
@@ -25,7 +25,6 @@ serde_json = "1.0.133"
 uuid = { version = "1.10.0", features = ["v5"] }
 
 [dev-dependencies]
-ctor = "0.2.8"
```
gnalh, line 28: Any nice tools for automatically finding unused deps? I tried a couple but they ended up breaking builds by mis-identifying unused dependencies.

gnalh (approved):

xcresult/Cargo.toml
dfrankland: I haven't used any personally—you probably know more than I do.

PR #304, opened 2025-01-16
Remove uuidv5 generation for xcresult

gnalh:
This has been moved into the ETL. Both forms will be supported until we EOL 0.6.11. Confirmed that uploads are still mapping to the same tests. 

Example:
Latest CLI and local CLI with this change
Screenshot 2025-01-21 at 9 40 25 AM

Only one test exists in new repo.
Screenshot 2025-01-21 at 9 40 39 AM

2 test runs matching the uploads
Screenshot 2025-01-21 at 9 40 48 AM
Screenshot 2025-01-21 at 9 40 44 AM

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5416,7 +5416,6 @@ dependencies = [
  "serde_json",
  "tar",
  "temp_testdir",
- "uuid",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


dfrankland (approved):

cmillar-trunk (approved):

gnalh:
Moving this back into review, looks like it breaks the quarantining logic when calling the api.

PR #308, opened 2025-01-17
Move api-client out of the cli

gnalh:
In an effort to remove having `libs` under the cli, separate this out into the api crate. This PR also cleans up some of the naming and adds corresponding request / response types where previously missing.


api/src/call_api.rs
```diff
@@ -74,7 +74,7 @@ where
             }
         });
 
-        let result = Retry::spawn(default_delay(), || (&mut self.action).run()).await;
+        let result = Retry::spawn(default_delay(), || self.action.run()).await;
```
gnalh, line 77: clippy changed this

api/src/client.rs
```diff
@@ -329,8 +340,8 @@ mod tests {
             .unwrap()
             .iter()
             .filter(|(_, message)| message.starts_with("Uploading bundle to S3"))
-            .cloned()
             .take(2)
+            .cloned()
```
gnalh, line 344: This was also changed by clippy.

cli/src/upload_command.rs
```diff
@@ -1,11 +1,11 @@
-use api::BundleUploadStatus;
+use api::client::ApiClient;
+use api::message::BundleUploadStatus;
```
dfrankland, line 2: nit: can we group these for consistency?

dfrankland (approved):
Oh! I just figured out that this is the config to make imports get auto-formatted!

https://rust-lang.github.io/rustfmt/?version=v1.8.0&search=#group_imports

cli/src/upload_command.rs
gnalh: Will do

PR #315, opened 2025-01-18
Fix quarantining logic

gnalh:
When `use-quarantining` is set to true we should defer to the cloud config. When it is set to false we should disable quarantining. Previously we were ignoring the state of that flag and treating everything as though it should be quarantined by always inspecting failures when exit codes were not specified. These exit codes are not specified for the `upload` and `quarantine` commands, so you would essentially always have quarantining enabled when using those commands. In the future this should be renamed and flipped, but this PR updates it to handle those cases appropriately. 

* Add tests to validate upload and quarantine behaviors match expectations

### DIFF
--- a/cli-tests/src/quarantine.rs
+++ b/cli-tests/src/quarantine.rs
@@ -15,6 +15,20 @@ use predicates::prelude::*;
 use tempfile::tempdir;
 use test_utils::mock_server::{MockServerBuilder, SharedMockServerState};
 
+#[derive(Debug, Clone, Copy)]
+enum QuarantineConfigResponse {
+    Disabled,
+    None,
+    Some,
+    All,
+}
+
+#[derive(Debug, Clone, Copy)]
+enum CreateBundleResponse {
+    Error,
+    Success,
+}
+
 // NOTE: must be multi threaded to start a mock server
 #[tokio::test(flavor = "multi_thread")]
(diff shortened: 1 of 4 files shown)


cli-tests/src/quarantine.rs
```diff
@@ -15,6 +15,20 @@ use predicates::prelude::*;
 use tempfile::tempdir;
```
gnalh, line 15: The tests here are very similar across upload and quarantine. I would like to parameterize them in a follow-up to make sure we are handling all variations of quarantining across the various sub commands. For the sake of getting this out sooner rather than later, I've opted to just add them with boilerplate and all (as the existing patterns has been doing).

cli/src/context_quarantine.rs
```diff
@@ -167,7 +167,7 @@ pub async fn gather_quarantine_context(
         )
     });
 
-    let exit_code = test_run_exit_code.unwrap_or_else(|| failed_tests_extractor.exit_code());
```
gnalh, line 170: This is the problematic logic. `test_run_exit_code` is always empty for `quarantine` and `upload`, so we end up looking at the exit code of the failed tests extractor, which is determined entirely by the results of the tests.

dfrankland (approved):

PR #318, opened 2025-01-27
[1] Resolve Pre-test to include Rspec debug parameters

gnalh:
This addresses 2 points.

1. It allows us to understand how rspec is being invoked to improve debugging
2. It addresses a `KERN_INVALID_ADDRESS` that occurs when using arm64-darwin cydlib builds. This error happens because we are unable to access `env::args` in this mode.

### DIFF
--- a/test_report/tests/report.rs
+++ b/test_report/tests/report.rs
@@ -25,7 +25,7 @@ async fn publish_test_report() {
     env::set_var("TRUNK_ORG_URL_SLUG", "test-org");
 
     let thread_join_handle = thread::spawn(|| {
-        let test_report = MutTestReport::new("test".into());
+        let test_report = MutTestReport::new("test".into(), "".into());
         test_report.add_test(
             Some("1".into()),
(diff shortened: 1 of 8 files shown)


test_report/tests/report.rs
```diff
@@ -25,7 +25,7 @@ async fn publish_test_report() {
     env::set_var("TRUNK_ORG_URL_SLUG", "test-org");
 
     let thread_join_handle = thread::spawn(|| {
-        let test_report = MutTestReport::new("test".into());
+        let test_report = MutTestReport::new("test".into(), "".into());
```
cmillar-trunk, line 28: This had `",".into()` before, but is now `"".into()`, without the comma. Is that a consequence of moving it up here?

gnalh: The second parameter is the underlying command for the test report. The first parameter is the origin. Both are optional and are for debugging purposes.

cmillar-trunk (approved):

PR #319, opened 2025-01-27
[2] Add gem generation support to context-ruby

gnalh:
* Update the ruby version we use
* Slight restructure to better match example repos -- mainly Cargo.lock and adding a workspace.
* Add pinned rb_sys
* Add gem compilation for all ruby 3.* versions and arm64/x86_64+darwin/linux


context-ruby/ext/context_ruby/Cargo.toml
```diff
@@ -11,3 +11,18 @@ anyhow = "1.0.93"
 context = { path = "../../../context", features = ["git-access", "ruby"] }
 magnus = { version = "0.7" }
 test_report = { path = "../../../test_report", features = ["ruby"] }
+openssl = { version = "0.10.66", features = ["vendored"], optional = true }
+openssl-src = { version = "=300.3.1", optional = true }
```
cmillar-trunk, line 15: I don't see either of these being used in this pr; are they used in the third pr?

context-ruby/Gemfile
```diff
@@ -2,10 +2,10 @@
 
 source 'https://rubygems.org'
 
+gemspec
+
 gem 'rake'
 gem 'rake-compiler', '1.2.0'
```
cmillar-trunk, line 8: What's the rule for which of these needs to be represented in the gemspec? I see `rspec` and `rspec-core`, but not any of the others here.

cmillar-trunk (approved):
lgtm, a few optional questions.

context-ruby/ext/context_ruby/Cargo.toml
gnalh: Good catch, it's a holdover from when I was debugging an openssl issue and doesn't need to be here.

PR #331, opened 2025-01-30
Add stable branches to parsing context

riya-n:
This will be used for customers who want to specify their stable branch. There shouldn't be any changes to the cli, just to the context parsers. If a customer hasn't specified a stable branch, we default to main/master in the etl, so stable branches should never be empty in the bindings. 

TRD: https://docs.google.com/document/d/13obb1nx-kyZbrBpAXTCvM40vsAMlc1qsPH0WGL-Aln8/edit?tab=t.0#heading=h.b93n5poqpkwb
Will be used here, left a TODO: https://github.com/trunk-io/trunk/pull/21965/files#diff-c62d1e1316815bde052dfcc010f4823873a059a58ca6c33ba4b078b89a9548d2R190


context/src/env/parser.rs
```diff
@@ -594,22 +620,28 @@ impl<'a> EnvParser<'a> {
...
+        ));
     }
 }
 
 #[cfg(feature = "ruby")]
 pub fn ruby_init(ruby: &magnus::Ruby) -> Result<(), magnus::Error> {
     let ci_platform = ruby.define_class("CIPlatform", ruby.class_object())?;
     ci_platform.define_method("to_s", magnus::method!(CIPlatform::to_string, 0))?;
+    let branch_class = ruby.define_class("BranchClass", ruby.class_object())?;
+    branch_class.define_method("to_s", magnus::method!(BranchClass::to_string, 0))?;
```
riya-n, line 644: I think this was missed in the ruby bindings, I was seeing an error when trying to get branch_class in the ruby tests

context/src/env/parser.rs
```diff
@@ -481,7 +502,12 @@ impl From<(&str, Option<usize>, Option<GitLabMergeRequestEventType>)> for Branch
             BranchClass::PullRequest
         } else if branch_name.starts_with("remotes/pull/") || branch_name.starts_with("pull/") {
             BranchClass::PullRequest
-        } else if matches!(branch_name, "master" | "main") {
+        } else if stable_branches
+            .filter(|branches| !branches.is_empty())
+            .unwrap_or(vec![String::from("main"), String::from("master")])
+            .iter()
+            .any(|branch| branch == branch_name)
```
riya-n, line 509: if stable branches is an empty vector or is none, we should default to main/master

context-ruby/ext/context_ruby/src/lib.rs
```diff
@@ -3,10 +3,13 @@ use std::collections::HashMap;
 use context::{env, repo};
 use test_report::report;
 
-pub fn env_parse(env_vars: magnus::RHash) -> Option<env::parser::CIInfo> {
+pub fn env_parse(
+    env_vars: magnus::RHash,
+    stable_branches: Option<Vec<String>>,
```
riya-n, line 8: should this use magnus::RArray ?

context-py/src/lib.rs
```diff
@@ -20,9 +20,13 @@ define_stub_info_gatherer!(stub_info);
 
 #[gen_stub_pyfunction]
 #[pyfunction]
-fn env_parse(env_vars: HashMap<String, String>) -> Option<env::parser::CIInfo> {
+#[pyo3(signature = (env_vars, stable_branches))]
```
riya-n, line 23: I believe this should fix the trailing optional warning: 
```
use of deprecated constant `__pyfunction_env_parse::SIGNATURE`: this function has implicit defaults for the trailing `Option<T>`
```
ref: https://pyo3.rs/v0.23.3/function/signature.html

context/src/env/parser.rs
dfrankland, line 509: We don't really need a `Vec<String>` here, so I think we can remove extra `clones`, passing my value, and memory allocations by working with `&[str]`. Example:

```rust
const DEFAULT_STABLE_BRANCHES: &[&str] = &["master", "main"];

fn check_branch_name_is_stable_branch(branch_name: &str, stable_branches: Option<&[&str]>) -> bool {
    stable_branches
        .filter(|branches| !branches.is_empty())
        .unwrap_or(DEFAULT_STABLE_BRANCHES)
        .contains(&branch_name)
}

fn main() {
    dbg!(check_branch_name_is_stable_branch("main", None));
    dbg!(check_branch_name_is_stable_branch("dev", None));

    let my_stable_branches_owned = vec![String::from("dev")];
    let my_stable_branches_ref = &my_stable_branches_owned
        .iter()
        .map(String::as_str)
        .collect::<Vec<&str>>();

    dbg!(check_branch_name_is_stable_branch(
        "main",
        Some(my_stable_branches_ref)
    ));
    dbg!(check_branch_name_is_stable_branch(
        "dev",
        Some(my_stable_branches_ref)
    ));
}
```

Playground link:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=44385122e111732cbf3a21117c3afb6e

There's a couple of other things to take note of here to:
* `unwrap_or` should be used when the "or" is a value, not an expression—when using an expression `unwrap_or_else` should be used. This prevents doing extra work when a value is `Some`
* `iter().any()` shouldn't be needed here since we are just checking if a value is present in the `Vec`, so `contains` should be just fine and faster

context-ruby/ext/context_ruby/src/lib.rs
gnalh: It's fine as is, but FYI I was planning to restructure and drop this logic since it isn't actually needed by rspec.

context/src/env/parser.rs
gnalh: Similar to the above, this will most likely be dropped in the future.

context/tests/env.rs
```diff
@@ -510,6 +510,53 @@ fn test_simple_github_graphite_merge_queue() {
...
+
+    let env_vars = EnvVars::from_iter(vec![
+        (String::from("GITHUB_ACTIONS"), String::from("true")),
+        (String::from("GITHUB_RUN_ID"), String::from(&run_id)),
+        (String::from("GITHUB_ACTOR"), String::from(&actor)),
+        (String::from("GITHUB_REPOSITORY"), String::from(&repository)),
+        (String::from("GITHUB_REF"), String::from(&branch)),
+        (String::from("GITHUB_WORKFLOW"), String::from(&workflow)),
+        (String::from("GITHUB_JOB"), String::from(&job)),
+    ]);
```
gnalh, line 530: It isn't in place yet, but we are working on centralizing our env variables. @cmillar-trunk for viz.

context-js/src/lib.rs
```diff
@@ -8,7 +8,10 @@ use wasm_bindgen::prelude::*;
 use wasm_streams::{readable::sys, readable::ReadableStream};
 
 #[wasm_bindgen]
-pub fn env_parse(env_vars: js_sys::Object) -> Option<env::parser::CIInfo> {
+pub fn env_parse(
+    env_vars: js_sys::Object,
+    stable_branches: Option<Vec<String>>,
```
riya-n, line 13: does this even need to be optional if we update bindings + add this param to the method in the same pr (in the monorepo)?

context-js/src/lib.rs
```diff
@@ -22,8 +25,15 @@ pub fn env_parse(env_vars: js_sys::Object) -> Option<env::parser::CIInfo> {
             }
         })
         .collect();
+
+    let stable_branches_unwrapped = stable_branches.unwrap_or_default();
+    let stable_branches_ref: &[&str] = &stable_branches_unwrapped
+        .iter()
+        .map(String::as_str)
+        .collect::<Vec<&str>>();
```
riya-n, line 33: thought it made more sense to send an empty array over an optional but let me know if we should keep using optionals instead

context/src/meta/bindings.rs
```diff
@@ -23,8 +23,14 @@ pub struct BindingsMetaContext {
...
     #[new]
-    pub fn new(ci_info: &CIInfo, repo: &BundleRepo) -> Self {
-        BindingsMetaContext::from(MetaContext::new(ci_info, repo))
+    #[pyo3(signature = (ci_info, repo, stable_branches))]
+    pub fn new(ci_info: &CIInfo, repo: &BundleRepo, stable_branches: Option<Vec<String>>) -> Self {
+        let stable_branches_unwrapped = stable_branches.unwrap_or_default();
+        let stable_branches_ref: &[&str] = &stable_branches_unwrapped
+            .iter()
+            .map(String::as_str)
+            .collect::<Vec<&str>>();
```
riya-n, line 32: unsure if this logic can be extracted into a common helper somewhere

context-js/src/lib.rs
riya-n: Yeah, defaults to main/master if the stable branches array is empty. I can update this to use Optionals all the way (and coalesce in BranchClass::From)

riya-n: Actually, planning on making this non-optional and defaulting to main/master in the etl. i.e., the stable_branches array should never be empty

context-py/tests/test_meta_validate.py
```diff
@@ -18,6 +19,42 @@ def test_branch_supplied_by_env():
     )
 
 
+def test_branch_supplied_by_env_stable_branches():
+    from context_py import env_parse  # trunk-ignore(pyright/reportUnknownVariableType)
```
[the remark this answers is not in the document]
riya-n: No longer an issue since we removed stable branches as optional

dfrankland (approved):

PR #334, opened 2025-01-31
Add reusable test framework for cli tests

cmillar-trunk:
Abstracts out the arg and env orchestration for cli invocation to a builder pattern. The builder fills in default values for parts that are common across existing usages. While we have all of the flags supported by the cli in the builder, we don't have methods for setting all of them because not every flag is actually tested, so those methods would be unused.

### DIFF
--- a/cli-tests/src/command_builder.rs
+++ b/cli-tests/src/command_builder.rs
@@ -1,0 +1,397 @@
+use std::{collections::HashMap, path::Path};
+
+use assert_cmd::Command;
+use constants::{TRUNK_API_CLIENT_RETRY_COUNT_ENV, TRUNK_PUBLIC_API_ADDRESS_ENV};
+
+use crate::utils::CARGO_RUN;
+
+const DEFAULT_JUNIT_PATHS: &str = "./*";
+
+pub struct UploadArgs {
+    repo_root: Option<String>,
+    repo_url: Option<String>,
+    repo_head_sha: Option<String>,
+    repo_head_branch: Option<String>,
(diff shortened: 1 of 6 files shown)


cli-tests/src/command_builder.rs
```diff
@@ -0,0 +1,397 @@
+use std::{collections::HashMap, path::Path};
+
+use assert_cmd::Command;
+use constants::{TRUNK_API_CLIENT_RETRY_COUNT_ENV, TRUNK_PUBLIC_API_ADDRESS_ENV};
+
+use crate::utils::CARGO_RUN;
+
+const DEFAULT_JUNIT_PATHS: &str = "./*";
+
+pub struct UploadArgs {
```
gnalh, line 10: Why duplicate this instead of importing it? Or is it subtly different?

gnalh (approved):
Love all of the code removal.

cli-tests/src/command_builder.rs
cmillar-trunk: It is slightly different:
- `junit-paths` and `bazel-bep-path` got moved up into `CommandBuilder.paths_state` because they're genuinely common to all subcommands and also mutually exclusive, and I wanted that captured in the types for the builder (this is also why you see some tests still use `.args()` to set a junit or bep path argument - they're testing that exclusivity, and the builder tries to avoid letting you create illegal argument sets).
- `org-url-slug` is set by this, but since no tests use it, it isn't a variable
- All of the bools are Options here, but mandatory there thanks to the defaults.

PR #346, opened 2025-02-04
Test no cross contamination of env vars

cmillar-trunk:
Adds a test to confirm that in the case where env vars are present for two possible ci providers, we only choose from one, to avoid a situation where a messy env causes cross-contamination.


gnalh (approved):
Love it!

PR #347, opened 2025-02-04
Try to discover repo root when not explicitly specified

gnalh:
* If repo root if provided, use it directly
* If repo root is not provided, use discover to find the root

### DIFF
--- a/test_report/src/report.rs
+++ b/test_report/src/report.rs
@@ -113,7 +113,7 @@ impl MutTestReport {
         let resolved_path_str = resolved_path.path().to_str().unwrap_or_default();
         let token = env::var("TRUNK_API_TOKEN").unwrap_or_default();
         let org_url_slug = env::var("TRUNK_ORG_URL_SLUG").unwrap_or_default();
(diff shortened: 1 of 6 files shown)


test_report/src/report.rs
```diff
@@ -113,7 +113,10 @@ impl MutTestReport {
         let resolved_path_str = resolved_path.path().to_str().unwrap_or_default();
         let token = env::var("TRUNK_API_TOKEN").unwrap_or_default();
         let org_url_slug = env::var("TRUNK_ORG_URL_SLUG").unwrap_or_default();
-        let repo_root = env::var("REPO_ROOT").unwrap_or(".".to_string());
+        let repo_root = match env::var("REPO_ROOT") {
```
cmillar-trunk, line 116: nit, optional: I think `env::var("REPO_ROOT").ok()` will do this without manually writing the match.

cmillar-trunk (approved):

test_report/src/report.rs
gnalh: Ah good catch, will add.

PR #358, opened 2025-02-05
Add missing upload parameters

gnalh:
We were missing the test count and version information.

Test count was caused by us only parsing out the count for junit. Instead we'll allow the uploader specify it. Otherwise we have a circular dependency between upload and test report.

Version information was missing because we weren't updating it when publishing the gem.

Screenshot 2025-02-05 at 10 34 55 AM

### DIFF
--- a/.github/workflows/release_ruby_gem.yml
+++ b/.github/workflows/release_ruby_gem.yml
@@ -55,7 +55,14 @@ jobs:
           bundler-cache: true
           working-directory: context-ruby
 
-      - name: Update version
+      - name: Update CLI version
+        uses: ./.github/actions/update_version
+        if: "!cancelled()"
+        with:
+          version: ${{ github.event.inputs.release_tag }}
+          file: ./cli/Cargo.toml
+
+      - name: Update Gem version
         uses: ./.github/actions/update_version
         if: "!cancelled()"
         with:
--- a/cli/src/context.rs
+++ b/cli/src/context.rs
@@ -354,6 +354,10 @@ fn parse_num_tests(file_sets: &[FileSet]) -> usize {
         .filter_map(|(file, bundled_file)| {
             let file_buf_reader = BufReader::new(file);
             let mut junit_parser = JunitParser::new();
+            // skip .bin files
+            if !bundled_file.original_path.ends_with(".xml") {
+                return None;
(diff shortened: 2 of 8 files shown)


test_report/src/report.rs
```diff
@@ -124,12 +124,12 @@ impl MutTestReport {
                 nanos: Utc::now().timestamp_subsec_nanos() as i32,
             });
         }
-        // TODO - handle finding the repo root automatically
```
gnalh, line 127: Holdover comment. This has been resolved.

cli/src/context.rs
```diff
@@ -354,6 +354,10 @@ fn parse_num_tests(file_sets: &[FileSet]) -> usize {
         .filter_map(|(file, bundled_file)| {
             let file_buf_reader = BufReader::new(file);
             let mut junit_parser = JunitParser::new();
+            // skip .bin files
+            if !bundled_file.original_path.ends_with(".xml") {
+                return None;
+            }
```
gnalh, line 360: This works fine as is, just adding this as a safety feature in case we forget to handle it when updating the junit parser.

.github/workflows/release_ruby_gem.yml
```diff
@@ -55,7 +55,14 @@ jobs:
...
 
-      - name: Update version
+      - name: Update CLI version
+        uses: ./.github/actions/update_version
+        if: "!cancelled()"
+        with:
+          version: ${{ github.event.inputs.release_tag }}
+          file: ./cli/Cargo.toml
+
+      - name: Update Gem version
```
gnalh, line 65: This was the missing version info. I'd like to be able to define this in a single place in the future.

gnalh:
Forgot to update / add the test, will push that up shortly.

bundle/src/files.rs
```diff
@@ -252,7 +252,8 @@ impl BundledFile {
         // When start is provided, check if file is stale
         if let Some(start) = start {
             let modified = path.metadata()?.modified()?;
-            if modified < start {
+            let difference = start.duration_since(modified);
+            if difference.is_err() {
```
gnalh, line 256: The previous method was imprecise and would cause the test_report test to sometimes fail.

PR #362, opened 2025-02-05
[TRUNK-13843] Don't retry on invalid json or some errors

cmillar-trunk:
Adds logic to check for specific errors or invalid json, and only retry when we don't hit those conditions. The list of errors is mostly derived from https://stackoverflow.com/questions/47680711/which-http-errors-should-never-trigger-an-automatic-retry

### DIFF
--- a/api/src/call_api.rs
+++ b/api/src/call_api.rs
@@ -1,8 +1,11 @@
 use std::{env, time::Duration};
 
 use constants::TRUNK_API_CLIENT_RETRY_COUNT_ENV;
+use http::StatusCode;
 use tokio::time::{self, Instant};
-use tokio_retry::{strategy::ExponentialBackoff, Action, Retry};
+use tokio_retry::{strategy::ExponentialBackoff, Action, RetryIf};
+
+use crate::client::{NOT_FOUND_CONTEXT, UNAUTHORIZED_CONTEXT};
 
 // Tokio-retry uses base ^ retry * factor formula.
 // This will give us 8ms, 64ms, 512ms, 4096ms, 32768ms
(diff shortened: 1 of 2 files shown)


api/src/call_api.rs
```diff
@@ -34,6 +37,81 @@ fn default_delay() -> std::iter::Take<ExponentialBackoff> {
         )
 }
 
+pub trait AbortableRetry {
+    fn should_retry(&self) -> bool;
+}
+
+impl AbortableRetry for anyhow::Error {
+    fn should_retry(&self) -> bool {
+        println!("Should retrying anyhow error {:#?}", self);
```
gnalh, line 46: Nit: remove `println` as they seem to just be for debugging. Use `log::debug!` instead.

api/src/call_api.rs
```diff
@@ -34,6 +37,81 @@ fn default_delay() -> std::iter::Take<ExponentialBackoff> {
...
+        let self_text = format!("{self}");
+        println!("it had text {:#?}", self_text);
```
gnalh, line 48: Nit: remove `println` as they seem to just be for debugging. Use `log::debug!` instead.

api/src/call_api.rs
```diff
@@ -34,6 +37,81 @@ fn default_delay() -> std::iter::Take<ExponentialBackoff> {
...
+            && self.chain().fold(true, |acc: bool, cause| {
+                let cause_should_retry =
+                    if let Some(reqwest_error) = cause.downcast_ref::<reqwest::Error>() {
+                        reqwest_error.should_retry()
+                    } else {
+                        true
+                    };
+                acc && cause_should_retry
+            });
+        println!("anyhow result was {:#?}", x);
```
gnalh, line 60: Nit: remove `println` as they seem to just be for debugging. Use `log::debug!` instead.

api/src/call_api.rs
```diff
@@ -34,6 +37,81 @@ fn default_delay() -> std::iter::Take<ExponentialBackoff> {
...
+                    StatusCode::VARIANT_ALSO_NEGOTIATES => true,
+                    StatusCode::INSUFFICIENT_STORAGE => true,
+                    StatusCode::LOOP_DETECTED => true,
+                    StatusCode::NOT_EXTENDED => true,
+                    StatusCode::NETWORK_AUTHENTICATION_REQUIRED => true,
+
+                    _ => false,
+                }
+            }));
+        println!("reqwest Result was {:#?}", x);
```
gnalh, line 110: Nit: remove `println` as they seem to just be for debugging. Use `log::debug!` instead.

api/src/call_api.rs
```diff
@@ -34,6 +37,81 @@ fn default_delay() -> std::iter::Take<ExponentialBackoff> {
...
+pub trait AbortableRetry {
+    fn should_retry(&self) -> bool;
+}
+
+impl AbortableRetry for anyhow::Error {
+    fn should_retry(&self) -> bool {
+        println!("Should retrying anyhow error {:#?}", self);
+        let self_text = format!("{self}");
+        println!("it had text {:#?}", self_text);
+        let x = !self_text.contains(UNAUTHORIZED_CONTEXT)
```
gnalh, line 49: nit: `x` is ambiguous in terms of naming. If we remove the `println!` at the bottom we can return directly.

api/src/call_api.rs
```diff
@@ -34,6 +37,81 @@ fn default_delay() -> std::iter::Take<ExponentialBackoff> {
...
+            });
+        println!("anyhow result was {:#?}", x);
+        x
+    }
+}
+
+impl AbortableRetry for reqwest::Error {
+    fn should_retry(&self) -> bool {
+        println!("Should retrying reqwest error {:#?}", self);
+        let x = !(self.is_decode()
```
gnalh, line 68: Same comment as above. `x` is ambiguous in terms of naming. If we remove the `println!` at the bottom we can return directly.

gnalh (approved):
Nice, just make sure to cleanup the println statements.

api/src/call_api.rs
cmillar-trunk: Thanks for the catch, I'd forgotten to remove these.

PR #364, opened 2025-02-06
Add custom env to debug info reports

cmillar-trunk:
Adds the custom env vars to the debug info reports since this was missed when first adding them.


gnalh (approved):
Is this something we can test?

cmillar-trunk:
> Is this something we can test?

Not easily afaict, but I suppose we could build a linter or somesuch that trawls the source for uses of env::var to enumerate what we use it for.

PR #375, opened 2025-02-07
Add sentry-log reporting

gnalh:
From RSpec gem: https://trunkio.sentry.io/issues/6279840516/?environment=development&project=4507772986982400&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&statsPeriod=7d&stream_index=0
From CLI: https://trunkio.sentry.io/issues/6277208418/?environment=development&project=4507772986982400&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&statsPeriod=7d&stream_index=2

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -2,7 +2,7 @@ use std::{env, io::Write};
 
 use clap::{Parser, Subcommand};
 use clap_verbosity_flag::{log::LevelFilter, InfoLevel, Verbosity};
-use constants::SENTRY_DSN;
+use third_party::sentry;
 use trunk_analytics_cli::{
     quarantine_command::{run_quarantine, QuarantineArgs},
     test_command::{run_test, TestArgs},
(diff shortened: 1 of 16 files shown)


test_report/src/report.rs
```diff
@@ -151,12 +174,12 @@ impl MutTestReport {
...
                 Ok(_) => true,
                 Err(e) => {
-                    println!("Error uploading: {:?}", e);
+                    log::error!("Error uploading: {:?}", e);
                     false
                 }
             }
         } else {
-            println!("Error gathering pre test context");
+            log::error!("Error gathering pre test context");
```
gnalh, line 182: Logging from the cli library now appears in the rspec plugin

cli/src/main.rs
```diff
@@ -112,6 +104,6 @@ fn setup_logger(log_level_filter: LevelFilter) -> anyhow::Result<()> {
     if let Ok(log) = std::env::var("TRUNK_LOG") {
         builder.parse_filters(&log);
     }
-    builder.init();
+    sentry::logger(builder)?;
```
cmillar-trunk, line 107: If `sentry::logger` always sets `max_level(log::LevelFilter::Info)`, but we set the `filter_level(log_level_filter)` based on verbosity flags passed in, do those conflict (eg if we set the verbosity up to debug, does the max_level block that)?

gnalh: They do not conflict. `max_level` is global, while `filter_level` is application specific so it takes precedence here. For reasons I haven't dug deep into, sentry does not allow our logger to output logs without setting the max_level. We can set them to match by having it passed in if that makes things clearer here.

cmillar-trunk (approved):

PR #377, opened 2025-02-09
Run ruby gem release when file changed

gnalh:
This won't publish the gem but will run through all of the steps. Similar to the `release.yml` workflow.

### DIFF
--- a/.github/workflows/release_ruby_gem.yml
+++ b/.github/workflows/release_ruby_gem.yml
@@ -1,6 +1,9 @@
 ---
 name: Release Ruby Gem
 on:
+  pull_request:
+    paths:
+      - .github/workflows/release_ruby_gem.yml
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


.github/workflows/release_ruby_gem.yml
```diff
@@ -57,14 +60,16 @@ jobs:
...
           file: ./cli/Cargo.toml
 
       - name: Update Gem version
         uses: ./.github/actions/update_version
-        if: "!cancelled()"
+        # only empty during testing
+        if: github.event.inputs.release_tag != ''
         with:
           version: ${{ github.event.inputs.release_tag }}
           file: context-ruby/context_ruby.gemspec
```
cmillar-trunk, line 75: Can we separate out the release to a separate action/step from the setup/testing and gate it behind a single check?

PR #380, opened 2025-02-10
[TRUNK-14129] Add structured logging

cmillar-trunk:
Switches our logging to use the tracing library, which gives us support for structured logs, using the sentry publisher so that these logs go to sentry. Adds a single span for execution so we can map out executions as we encounter issues, and converts our use of error to warn to avoid spamming oncall.

Logging with quiet flag:
2025-02-10-075854_1119x70_scrot

With default verbosity:
2025-02-10-075802_1161x70_scrot

With single verbosity:
2025-02-10-075815_1119x144_scrot

With double verbosity:
2025-02-10-075833_1400x231_scrot

Fake error uploaded to sentry: https://trunkio.sentry.io/issues/6293011644/?environment=development&project=4507772986982400&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&statsPeriod=14d&stream_index=0


cli/src/main.rs
```diff
@@ -89,22 +102,22 @@ async fn run(cli: Cli) -> anyhow::Result<i32> {
...
     }
-    sentry::logger(builder, log_level_filter)?;
+}
+
+fn setup_logger(log_level_filter: LevelFilter) -> anyhow::Result<()> {
+    tracing_subscriber::registry()
+        .with(tracing_subscriber::fmt::layer().with_target(false))
+        .with(sentry_tracing::layer())
+        .with(to_trace_filter(log_level_filter))
+        .init();
```
gnalh, line 121: We'll need something similar inside of `test_report` so that we have traces for the rspec gem.

cli/src/main.rs
```diff
@@ -49,15 +61,16 @@ fn main() -> anyhow::Result<()> {
             let cli = Cli::parse();
             let log_level_filter = cli.verbose.log_level_filter();
             setup_logger(log_level_filter)?;
+            tracing::info_span!("Running command", command = cli.command.command_name());
```
gnalh, line 64: We should use `env::args` to capture the full command

third-party/src/sentry.rs
```diff
@@ -22,7 +23,16 @@ pub fn init(
...
+    opts.dsn = match SENTRY_DSN.into_dsn() {
+        Ok(dsn) => dsn,
+        Err(_) => {
+            // Logging is not set up at this point, but this is a pure code error
+            // that can only be caused by bad formatting of the dsn const.
+            None
+        }
+    };
+
+    sentry::init(opts)
```
gnalh, line 35: For my own knowledge, why is this needed?

cli/src/main.rs
cmillar-trunk: Gotcha, will add.

third-party/src/sentry.rs
cmillar-trunk: `sentry::init` afaik is to setup the client for sending to sentry; as opposed to the subscriber which handles bundling events up. I'm not sure why I needed to set the dsn specifically, I suspect just a difference between how tracing uses sentry vs the baseline sentry crate.

cli/src/main.rs
gnalh, line 64: Where are we tracking the repo + org_slug?

api/src/client.rs
```diff
@@ -456,10 +456,10 @@ mod tests {
         guard.flush(None);
```
gnalh, line 456: Can we also track the endpoint being hit in the trace? This will help us understand staging vs prod issues.

gnalh (approved):

cli/src/main.rs
cmillar-trunk: I decided not to do that because I don't believe we should be logging the token.

cmillar-trunk: Added the org slug to the overall info, and the repo to `context/src/repo/mod.rs`, where we actually open the repo.

api/src/client.rs
cmillar-trunk: Yes, added it to ApiClient::new when we resolve the host.

cli/src/main.rs
gnalh: We have logic to handle that
https://github.com/trunk-io/analytics-cli/blob/efdad1f14b2929e26db74531e51b07c02ca8e157/cli/src/context.rs#L44-L51

cmillar-trunk: oop, switched to use that.

PR #392, opened 2025-02-11
Fix force sentry dev

cmillar-trunk:
Github actions needed the boolean param to be compared to a single-quoted string to recognize it.


gnalh (approved):
I'm having dejavu here

PR #398, opened 2025-02-12
[sentry] Update config parameters

gnalh:
* allow sentry to be disabled via override (this will be used within a testing context)
* grab correct release information

### DIFF
--- a/third-party/src/sentry.rs
+++ b/third-party/src/sentry.rs
@@ -1,17 +1,25 @@
+use std::borrow::Cow;
+
 pub const SENTRY_DSN: &str =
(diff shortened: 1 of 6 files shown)


third-party/src/sentry.rs
```diff
@@ -1,18 +1,26 @@
...
         {
             opts.environment = Some("development".into());
         }
     } else {
         opts = options.unwrap_or_default();
     }
+    opts.release = Some(release_name);
+    if std::env::var("DISABLE_SENTRY").is_ok() {
+        opts.sample_rate = 0.0;
+    }
```
gnalh, line 23: There isn't an explicit disable for sentry, so this seemed like the best path forward.

cmillar-trunk (approved):

PR #404, opened 2025-02-14
Add better logging

cmillar-trunk:
Adds the logic requested by BB-340, with the exception of showing status since logs, as we can get all of those without needing to modify the flaky tests server.

Added logs outlined in red:
command_screenshot

### DIFF
--- a/api/src/urls.rs
+++ b/api/src/urls.rs
@@ -1,0 +1,64 @@
+use bundle::Test;
+use context::repo::RepoUrlParts;
+use url::{form_urlencoded, ParseError, Url};
+
+pub fn url_for_test_case(
+    public_api_address: &String,
+    org_url_slug: &String,
+    repo: &RepoUrlParts,
+    test_case: &Test,
+) -> Result<String, ParseError> {
+    let mut url = Url::parse(convert_to_app(public_api_address).as_str())?;
+    url.set_path(test_path(org_url_slug, test_case).as_str());
+    url.set_query(Some(repo_query(repo).as_str()));
+    Ok(url.to_string())
+}
+
+fn convert_to_app(public_api_address: &String) -> String {
+    public_api_address.replace("https://api.", "https://app.")
+}
+
+fn test_path(org_url_slug: &String, test_case: &Test) -> String {
+    format!("{}/flaky-tests/test/{}", org_url_slug, test_case.id)
+}
+
+fn repo_query(repo: &RepoUrlParts) -> String {
+    let value: String =
+        form_urlencoded::byte_serialize(format!("{}/{}", repo.owner, repo.name).as_bytes())
+            .collect();
+    format!("repo={}", value)
+}
+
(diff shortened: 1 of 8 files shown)


api/src/urls.rs
```diff
@@ -0,0 +1,64 @@
...
+    repo: &RepoUrlParts,
+    test_case: &Test,
+) -> Result<String, ParseError> {
+    let mut url = Url::parse(convert_to_app(public_api_address).as_str())?;
+    url.set_path(test_path(org_url_slug, test_case).as_str());
+    url.set_query(Some(repo_query(repo).as_str()));
+    Ok(url.to_string())
+}
+
+fn convert_to_app(public_api_address: &String) -> String {
```
gnalh, line 17: +1 to the linter error around using String here.

api/src/urls.rs
```diff
@@ -0,0 +1,64 @@
...
+    public_api_address.replace("https://api.", "https://app.")
+}
+
+fn test_path(org_url_slug: &String, test_case: &Test) -> String {
+    format!("{}/flaky-tests/test/{}", org_url_slug, test_case.id)
```
gnalh, line 22: @federicoweber we'll need to make sure these urls are stable.

cli/src/main.rs
```diff
@@ -121,9 +125,28 @@ fn to_trace_filter(filter: log::LevelFilter) -> tracing::level_filters::LevelFil
...
+        )
         .with(sentry_tracing::layer())
         .with(to_trace_filter(log_level_filter))
         .init();
     Ok(())
 }
+
+// Uses a raw string to avoid needing to escape quotes in the title card. This is mostly just so you can see
+// what it looks like in code rather than needing to print.
+const TITLE_CARD: &str = r#"
```
gnalh, line 141: How does this look on CI and different terminal sizes?

gnalh (approved):
LGTM, just a handful of comments.

cli/src/main.rs
cmillar-trunk: On smaller terminals it'll wrap around in weird ways - classic limitation of ascii art.

Here's what it looks like on ci, and on a small ci window (the weird wraparound is the same sort of thing you get on small terminals).
2025-02-14-134622_824x263_scrot
2025-02-14-134643_498x395_scrot

cli/src/context_quarantine.rs
```diff
@@ -225,6 +225,16 @@ pub async fn gather_quarantine_context(
...
                 failure.id
             );
+            if quarantine_failure {
+                if let Ok(url) = url_for_test_case(
+                    &api_client.host,
+                    &request.org_url_slug,
+                    &request.repo,
+                    &failure,
+                ) {
+                    tracing::info!("Learn more > {}", url);
```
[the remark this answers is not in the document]
cmillar-trunk: Sure, will do.

PR #406, opened 2025-02-18
Small sentry fixes

gnalh:
Addresses the following issues:
* The command is being filtered because `token` exists in the string
* sentry isn't treating out release as semver compatible

### DIFF
--- a/test_report/src/report.rs
+++ b/test_report/src/report.rs
@@ -115,7 +115,7 @@ impl MutTestReport {
 
     // sends out to the trunk api
     pub fn publish(&self) -> bool {
-        let release_name = env!("CARGO_PKG_VERSION");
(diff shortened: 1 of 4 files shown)


test_report/src/report.rs
```diff
@@ -115,7 +115,7 @@ impl MutTestReport {
 
     // sends out to the trunk api
     pub fn publish(&self) -> bool {
-        let release_name = env!("CARGO_PKG_VERSION");
+        let release_name = format!("rspec-flaky-tests@{}", env!("CARGO_PKG_VERSION"));
```
gnalh, line 118:
```suggestion
        let release_name = format!("rspec-trunk-flaky-tests@{}", env!("CARGO_PKG_VERSION"));
```

cmillar-trunk (approved):

PR #409, opened 2025-02-19
Update smoke test CLI version

gnalh:


.github/workflows/smoke_test.yml
```diff
@@ -6,7 +6,7 @@ on:
   schedule:
     - cron: 0 0/2 * * *
 env:
-  RELEASE: 0.6.11
+  RELEASE: 0.6.16
```
gnalh, line 9:
```suggestion
  RELEASE: 0.6.18
```

.github/workflows/smoke_test.yml
```diff
@@ -6,7 +6,7 @@ on:
   schedule:
     - cron: 0 0/2 * * *
 env:
-  RELEASE: 0.6.11
+  RELEASE: 0.6.18
```
cmillar-trunk, line 9: Is this the version zillow is now using?

gnalh: Yes

cmillar-trunk (approved):

PR #425, opened 2025-02-20
Upload results to prod

gnalh:
This will help validate that cli changes work in both environments


.github/actions/upload_test_results/action.yaml
```diff
@@ -1,23 +0,0 @@
-name: Upload test results
```
gnalh, line 1: This action is only used in one place and would collapse the staging + prod uploads. Moved them to the workflow to improve readability.

cmillar-trunk (approved):

PR #426, opened 2025-02-21
Update title card

cmillar-trunk:
The title card has some, shall we say, display issues. Changing it to avoid those.


dfrankland (approved):
This is a nice creative update!

gnalh (approved):

PR #427, opened 2025-02-21
Group failures in logs

cmillar-trunk:
Groups failures in logs by whether they're quarantined or not.

2025-02-21-081123_1613x93_scrot


gnalh (approved):
The spacing for the quarantined looks a little off in your screenshot. Was it an old one before adding a space? Other than that, this looks great!

PR #432, opened 2025-02-21
Remove update bundle upload status call

gnalh:
This is handled elsewhere now.

### DIFF
--- a/api/src/message.rs
+++ b/api/src/message.rs
@@ -3,18 +3,6 @@ use context::repo::RepoUrlParts;
 use serde::{Deserialize, Serialize};
 
 #[derive(Debug, Serialize, Clone, Deserialize, PartialEq, Eq)]
-pub enum BundleUploadStatus {
-    #[serde(rename = "PENDING")]
-    Pending,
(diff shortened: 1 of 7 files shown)


api/src/message.rs
```diff
@@ -61,13 +49,6 @@ pub struct CreateRepoRequest {
...
 
-#[derive(Debug, Serialize, Clone, Deserialize, PartialEq, Eq)]
-pub struct UpdateBundleUploadRequest {
-    pub id: String,
-    #[serde(rename = "uploadStatus")]
-    pub upload_status: BundleUploadStatus,
-}
-
 #[derive(Debug, Serialize, Clone, Deserialize, PartialEq, Eq)]
 pub struct UpdateBundleUploadResponse {}
```
cmillar-trunk, line 53: Can we get rid of the response too?

cmillar-trunk (approved):

api/src/message.rs
gnalh: Good catch. Missed those.

gnalh:
Screenshot 2025-02-21 at 1 28 40 PM
Confirmed it works as expected

PR #434, opened 2025-02-21
Remove create repo calls

gnalh:
These are going to be handled elsewhere.

### DIFF
--- a/api/src/call_api.rs
+++ b/api/src/call_api.rs
@@ -179,16 +179,16 @@ mod tests {
 
     use lazy_static::lazy_static;
     use reqwest::Response;
+    use serde::{Deserialize, Serialize};
     use tokio::time;
 
     use super::{
         CallApi, CHECK_PROGRESS_INTERVAL_SECS, REPORT_SLOW_PROGRESS_TIMEOUT_SECS,
         RETRY_COUNT_DEFAULT,
     };
-    use crate::{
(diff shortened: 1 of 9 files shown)


api/src/call_api.rs
```diff
@@ -318,7 +318,7 @@ mod tests {
                 time::sleep(Duration::from_secs(CHECK_PROGRESS_INTERVAL_SECS - 1)).await;
                 retry_count.fetch_add(1, Ordering::Relaxed);
                 Response::from(http::Response::new("{'invalid': 'json'"))
-                    .json::<CreateRepoResponse>()
+                    .json::<EmptyResponse>()
```
cmillar-trunk, line 321: I believe this could be done with any json struck we have, rather than needing a new `EmptyResponse` struct?

cli-tests/src/upload.rs
```diff
@@ -45,7 +45,7 @@ async fn upload_bundle() {
         .failure();
 
     let requests = state.requests.lock().unwrap().clone();
-    assert_eq!(requests.len(), 4);
+    assert_eq!(requests.len(), 3);
```
cmillar-trunk, line 48: It's really nice seeing these counts go down.

cmillar-trunk (approved):

api/src/call_api.rs
gnalh: Opted to be explicit here that it's intentionally empty since we aren't using it.

PR #436, opened 2025-02-24
[TRUNK-14145] Disable quarantining flag

cmillar-trunk:
Replaces the enable-quarantining flag with a disable-quarantining flag.

### DIFF
--- a/cli-tests/src/command_builder.rs
+++ b/cli-tests/src/command_builder.rs
@@ -18,6 +18,7 @@ pub struct UploadArgs {
     no_upload: Option<bool>,
     team: Option<String>,
     codeowners_path: Option<String>,
+    use_quarantining: Option<bool>,
     disable_quarantining: Option<bool>,
     allow_empty_test_results: Option<bool>,
 }
@@ -35,6 +36,7 @@ impl UploadArgs {
             no_upload: None,
             team: None,
             codeowners_path: None,
+            use_quarantining: None,
             disable_quarantining: None,
             allow_empty_test_results: None,
         }
@@ -119,6 +121,17 @@ impl UploadArgs {
                 .into_iter()
                 .flat_map(|codeowners_path: String| {
                     vec![String::from("--codeowners-path"), codeowners_path]
+                }),
+        )
+        .chain(
+            self.use_quarantining
+                .into_iter()
+                .flat_map(|use_quarantining: bool| {
+                    if use_quarantining {
+                        vec![String::from("--use-quarantining")]
+                    } else {
+                        vec![String::from("--use-quarantining=false")]
+                    }
                 }),
         )
         .chain(
@@ -222,6 +235,20 @@ impl CommandType {
                 empty
             }
         }
+    }
+
+    pub fn use_quarantining(&mut self, new_flag: bool) -> &mut Self {
+        match self {
+            CommandType::Upload { upload_args, .. } => {
+                upload_args.use_quarantining = Some(new_flag)
+            }
+            CommandType::Quarantine { upload_args, .. } => {
+                upload_args.use_quarantining = Some(new_flag)
+            }
+            CommandType::Test { upload_args, .. } => upload_args.use_quarantining = Some(new_flag),
+            CommandType::Validate { .. } => (),
+        }
+        self
     }
 
     pub fn disable_quarantining(&mut self, new_flag: bool) -> &mut Self {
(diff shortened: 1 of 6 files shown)


cli-tests/src/command_builder.rs
```diff
@@ -122,13 +122,13 @@ impl UploadArgs {
...
                 .into_iter()
-                .flat_map(|use_quarantining: bool| {
-                    if use_quarantining {
-                        vec![String::from("--use-quarantining")]
+                .flat_map(|disable_quarantining: bool| {
+                    if disable_quarantining {
+                        vec![String::from("--disable-quarantining")]
                     } else {
-                        vec![String::from("--use-quarantining=false")]
+                        vec![String::from("--disable-quarantining=false")]
```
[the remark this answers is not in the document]
cmillar-trunk: Ah, gotcha. I'll re-add it and add tests that mirror behaviour for both flags.

cli-tests/src/quarantine.rs
```diff
@@ -225,3 +225,102 @@ async fn do_no_quarantines_tests_when_use_quarantined_disabled() {
     *CREATE_BUNDLE_RESPONSE.lock().unwrap() = CreateBundleResponse::Error;
     command.assert().success().stdout(upload_failure.clone());
 }
+
+#[tokio::test(flavor = "multi_thread")]
+async fn do_not_quarantines_tests_when_quarantine_disabled_set() {
```
gnalh, line 230: Nit: This can stay here for now, but I'd like to remove the `quarantine` subcommand in the future. It isn't used by anyone and doesn't have a clear story. I'd prefer this test exist in `test` / `upload` instead, but understand if it's here because it's easier to understand.

cli-tests/src/quarantine.rs
```diff
@@ -225,3 +225,102 @@ async fn do_no_quarantines_tests_when_use_quarantined_disabled() {
...
+    *CREATE_BUNDLE_RESPONSE.lock().unwrap() = CreateBundleResponse::Success;
+    command
+        .assert()
+        .success()
+        .stdout(upload_failure.clone().not());
+
+    // Fifth run will run with quarantining disabled, but will log upload failure
+    *QUARANTINE_CONFIG_RESPONSE.lock().unwrap() = QuarantineConfigResponse::Disabled;
+    *CREATE_BUNDLE_RESPONSE.lock().unwrap() = CreateBundleResponse::Error;
+    command.assert().success().stdout(upload_failure.clone());
```
gnalh, line 325: Shouldn't this be `failure` if quarantining is disabled and there was a test failed?

cli-tests/src/quarantine.rs
cmillar-trunk: I agree on the reasoning, but I put it here because this test mirrors the test above it, which is for the quarantine_enabled=false case.

cli-tests/src/quarantine.rs
cmillar-trunk: It should not. The reason is that quarantine is running on a stored file, and we pass `None` to `run_upload`'s `test_run_result` argument. Then that gets passed to `gather_exit_code_and_quarantined_tests_context`, where we see that it's `None` and return `QuarantineContext::default()`, and `i32`'s default is 0.

gnalh: Then I don't think this is actually testing what it says it is in the comment.
> // Fifth run will run with quarantining disabled, but will log upload failure

We aren't actually verifying this.

gnalh (approved):
LGTM, just make sure to add the missing test.

cli-tests/src/quarantine.rs
cmillar-trunk: We are: `command.assert().success().stdout(upload_failure.clone())` will panic if the log message we're expecting doesn't get printed.

gnalh: Maybe I'm missing something here, but it seems like we aren't verifying that with this flag that quarantining is actually disabled. The way I see it is we have:
1. a test that validates quarantining succeeds because there was a failure in a test but it was quarantined (quarantining is enabled)
2. a test that validates quarantining fails because there was a failure in a test but it was not quarantined (quarantining is set to disabled with this flag).

cmillar-trunk: Ah, gotcha. I'll add tests to verify that.

PR #443, opened 2025-02-27
[TRUNK-14370] Do not error on forbidden/unauthorized

cmillar-trunk:
Changes our exit logic to not use error reporting on forbidden/unauthorized errors. While I was there, added a nudge to check your token.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -2,6 +2,7 @@ use std::env;
 
 use clap::{Parser, Subcommand};
 use clap_verbosity_flag::{log::LevelFilter, InfoLevel, Verbosity};
+use http::StatusCode;
 use third_party::sentry;
 use tracing_subscriber::prelude::*;
 use trunk_analytics_cli::{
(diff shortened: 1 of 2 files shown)


cli/src/main.rs
```diff
@@ -74,18 +75,30 @@ fn main() -> anyhow::Result<()> {
...
-                        tracing::error!("Error: {:?}", e);
-                        guard.flush(None);
-                        std::process::exit(exitcode::SOFTWARE);
+
+                    if let Some(reqwest_error) = root_cause.downcast_ref::<reqwest::Error>() {
+                        if let Some(status) = reqwest_error.status() {
+                            if status == StatusCode::UNAUTHORIZED || status == StatusCode::FORBIDDEN {
+                                tracing::warn!("Unauthorized to access trunk, are you sure your token is correct? {:?}", error);
+                                guard.flush(None);
+                                std::process::exit(exitcode::OK);
```
gnalh, line 93: This might be a footgun when enabling trunk. If we exit success even though it's misconfigured you'll see the CI job succeed and assume all is well.

cmillar-trunk: Fair; I'll switch it to use the software exit code.

cli/src/context_quarantine.rs
```diff
@@ -182,7 +184,7 @@ pub async fn gather_quarantine_context(
         let result = api_client.get_quarantining_config(request).await;
 
         if let Err(ref err) = result {
-            tracing::error!("{}", err);
+            log_error(err, None);
```
gnalh, line 187: When should this be used? I assume only in this context, but want to make sure.

gnalh (approved):

cli/src/context_quarantine.rs
cmillar-trunk: Yeah, basically only when the cli crate wants to report an error.

PR #450, opened 2025-03-03
[codeowners] Don't fail to parse for invalid rows

gnalh:
A bit of a code cleanup in code owners around error generation. If one line within a section is invalid, we should not fail to parse the entire code owners files. This update adds a way to identify whether a parsing action is fatal and opts to return an empty vector for those cases. We will still fail if large sections of the code owners are deemed incorrect.


codeowners/src/gitlab/mod.rs
```diff
@@ -100,7 +100,12 @@ impl FromReader for GitLabOwners {
         if !file.valid() {
             let error_messages: Vec<String> =
                 file.errors().iter().map(ToString::to_string).collect();
-            return Err(anyhow::Error::msg(error_messages.join("\n")));
+            // check if failure should stop parsing
+            if file.errors().iter().any(|e| e.is_fatal()) {
+                return Err(anyhow::Error::msg(error_messages.join("\n")));
+            } else {
+                tracing::warn!("Suboptimal errors found: {:?}", error_messages.join("\n").replace("\t", " "));
```
gnalh, line 107: This is surfaced like this when using `validate`
```
All 6 files are valid! (6 files with validation warnings) ✅
 WARN Suboptimal errors found: ":187 missing entry owner"
```

cmillar-trunk (approved):

PR #456, opened 2025-03-04
[TRUNK-14308] Update merge-catching logic

cmillar-trunk:
The logic for checking merge commits was looking for refs of refs/remotes/pull/$pr_number/merge, but sometimes we get refs of refs/pull/$pr_number/merge. So we update the regex to support that case.

### DIFF
--- a/context/src/repo/mod.rs
+++ b/context/src/repo/mod.rs
@@ -405,3 +405,31 @@ impl RepoUrlParts {
         Self { host, owner, name }
     }
 }
+
+#[test]
+fn matches_gh_remote_refs() {
+    assert!(GH_MERGE_BRANCH_REGEX.is_match("refs/remotes/pull/114342/merge"));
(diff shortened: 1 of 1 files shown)


context/src/repo/mod.rs
```diff
@@ -19,7 +19,7 @@ pub const GIT_REMOTE_ORIGIN_URL_CONFIG: &str = "remote.origin.url";
 
 lazy_static! {
     static ref GH_MERGE_BRANCH_REGEX: Regex =
-        Regex::new(r"refs\/remotes\/pull\/[0-9]+\/merge").unwrap();
+        Regex::new(r"refs\/(remotes\/)?pull\/[0-9]+\/merge").unwrap();
```
gnalh, line 22: We should unit test this regex.

gnalh (approved):

PR #458, opened 2025-03-05
Get event type

cmillar-trunk:
Smoke tests were falling back to testing everything for release. GH docs indicate that we were using the wrong field to fetch the repository dispatch event type, so updating to the correct value.


gnalh (approved):
Good catch!

PR #460, opened 2025-03-05
Specify remote urls in getQuarantineConfig

gnalh:
That endpoint is being updated to handle creating a repo if one doesn't exist, so we need to add this data.


gnalh:
Once the endpoint update is released, this should unblock TRUNK-14336

cmillar-trunk (approved):

PR #461, opened 2025-03-05
[TRUNK-14363] Add settings to auth messages

cmillar-trunk:
Adds the settings page link to auth error messages.
2025-03-05-142300_2449x611_scrot

### DIFF
--- a/api/src/client.rs
+++ b/api/src/client.rs
@@ -377,7 +377,7 @@ fn adds_settings_if_domain_present() {
         add_settings_url_to_context("base_context", domain, &String::from("fake-org-slug"));
     assert_eq!(
         final_context,
-        "base_context\nYour settings page can be found at: https://app.fake-trunk.io/fake-org-slug/settings",
+        "base_context\nHint - Your settings page can be found at: https://app.fake-trunk.io/fake-org-slug/settings",
     )
 }
 
--- a/api/src/client.rs
+++ b/api/src/client.rs
@@ -298,8 +298,8 @@ pub(crate) fn status_code_help<T: FnMut(&Response) -> String>(
     check_unauthorized: CheckUnauthorized,
     check_not_found: CheckNotFound,
     mut create_error_message: T,
-    api_host: &String,
-    org_url_slug: &String,
+    api_host: &str,
+    org_url_slug: &str,
 ) -> anyhow::Result<Response> {
     let base_error_message = create_error_message(&response);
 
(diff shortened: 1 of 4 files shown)


api/src/client.rs
```diff
@@ -313,6 +332,48 @@ pub(crate) fn status_code_help<T: FnMut(&Response) -> String>(
...
+fn adds_settings_if_domain_present() {
+    let domain = url::Url::parse("https://api.fake-trunk.io/")
+        .ok()
+        .map(|url| url.domain().map(|domain| String::from(domain)))
+        .flatten();
+    let final_context =
+        add_settings_url_to_context("base_context", domain, &String::from("fake-org-slug"));
+    assert_eq!(
+        final_context,
+        "base_context Your settings page can be found at: https://app.fake-trunk.io/fake-org-slug/settings",
```
gnalh, line 366: The message we print is fairly long, can we split this up into smaller chunks using new lines? Also, I'd love if we hid the `HTTP status error: ***` message. It's useful for us to track, but not for users since we're already providing a call to action here.

cmillar-trunk: Good points, I can add the newlines easily, but avoiding showing the base error is a bit trickier because we're storing these messages in the error atm. I'll rework it in another pr if that's alright?

gnalh: Sure, that works. We'll just make sure that both changes are in the same release.

cli-tests/src/test.rs
```diff
@@ -254,7 +254,7 @@ async fn quarantining_resets_fail_code() {
         vec![
             String::from("bash"),
             String::from("-c"),
-            String::from("touch ./*; exit 1"),
+            String::from("sleep 1; touch ./*; exit 1"),
```
gnalh, line 257: Why is this sleep needed?

api/src/client.rs
```diff
@@ -313,6 +340,54 @@ pub(crate) fn status_code_help<T: FnMut(&Response) -> String>(
...
+                .into_iter()
+                .map(String::from)
+                .collect::<Vec<String>>()
+        })
+        .next();
+    let final_context =
+        add_settings_url_to_context("base_context", domain, &String::from("fake-org-slug"));
+    assert_eq!(
+        final_context,
+        "base_context\nYour settings page can be found at: https://app.fake-trunk.io/fake-org-slug/settings",
```
gnalh, line 380: Nit: Maybe call out that this is a hint?
```suggestion
        "base_context\nHint - Your settings page can be found at: https://app.fake-trunk.io/fake-org-slug/settings",
```

gnalh (approved):

PR #468, opened 2025-03-07
Don't show errors in console

cmillar-trunk:
Changes our requests to simply log context, and adds a hidden_in_console field filter so that when we have logs that should go to sentry only, we can hide them from the user. Uses this to hide internal rust errors from our existing logging.

2025-03-07-085547_1834x598_scrot


cli/src/error_report.rs
```diff
@@ -19,17 +18,14 @@ pub fn log_error(error: &anyhow::Error, base_message: Option<&str>) -> i32 {
...
-                );
+                tracing::warn!("{}", message(base_message, "unauthorized to access trunk"),);
                 return exitcode::SOFTWARE;
             }
         }
     }
 
-    tracing::error!("{}: {:?}", message(base_message, "error"), error,);
+    tracing::error!("{}", message(base_message, "error"));
+    tracing::error!(hidden_in_console = true, "Caused by error: {:#?}", error);
```
gnalh, line 28: Love this

PR #471, opened 2025-03-10
[TRUNK-14377] Tags for sentry uploads

cmillar-trunk:
Uses spans to add tags to sentry uploads. Also removes logging levels since it's a one-liner.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -3,6 +3,7 @@ use std::env;
 use clap::{Parser, Subcommand};
 use clap_verbosity_flag::{log::LevelFilter, InfoLevel, Verbosity};
 use third_party::sentry;
+use tracing::Instrument;
 use tracing_subscriber::prelude::*;
 use trunk_analytics_cli::{
     context::gather_debug_props,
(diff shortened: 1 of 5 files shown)


cli/src/main.rs
```diff
@@ -119,19 +157,46 @@ fn to_trace_filter(filter: log::LevelFilter) -> tracing::Level {
...
-        _ => sentry_tracing::EventFilter::Ignore,
+fn setup_logger(
+    log_level_filter: LevelFilter,
+    command_name: &str,
+    org_url_slug: String,
+    repo_root: String,
+) -> anyhow::Result<()> {
+    let command_string = String::from(command_name);
+    let sentry_layer = sentry_tracing::layer().event_mapper(move |event, context| {
+        // trunk-ignore(clippy/match_ref_pats)
```
gnalh, line 168: What's being ignored here? Which match is missing/

gnalh (approved):
Love this!

cli/src/main.rs
cmillar-trunk: None, there's a lint issue with using `&` to define the matches. This existed with the `.event_filter` before this change, I just had to move the `trunk-ignore` inside the closure because the curly braces changed which lines counted.

PR #472, opened 2025-03-11
Address issues with printing debug command

gnalh:


cli/src/context.rs
```diff
@@ -39,16 +41,19 @@ pub struct PreTestContext {
...
 }
 
+lazy_static! {
+    static ref COMMAND_REGEX: Regex = Regex::new(r"--token[=]?").unwrap();
+}
+
 // This function is used to gather debug properties for the bundle meta.
 // It will trigger EXC_BAD_ACCESS on arm64-darwin builds when compiled under cdylib
-pub fn gather_debug_props(token: String) -> BundleMetaDebugProps {
+pub fn gather_debug_props(args: Vec<String>, token: String) -> BundleMetaDebugProps {
```
gnalh, line 50: Added this as input to make testings easier.

cmillar-trunk:
Just to double-check, the result of this is that we just don't include the `--token abcd` when printing the command? What happens if I write the argument as `--token=abcd`?

cmillar-trunk (approved):

PR #481, opened 2025-03-12
Propagate `--allow-empty-test-result` to bep parsing

gnalh:


cli/src/context.rs
```diff
@@ -192,7 +191,7 @@ fn coalesce_junit_path_wrappers(
     #[cfg(target_os = "macos")] xcresult_path: Option<String>,
     #[cfg(target_os = "macos")] repo: &RepoUrlParts,
     #[cfg(target_os = "macos")] org_url_slug: String,
-    #[cfg(target_os = "macos")] allow_empty_test_results: bool,
+    allow_empty_test_results: bool,
```
gnalh, line 194: This will now be used outside of macOS specific applications. It was previously only used by xcresult.

cmillar-trunk (approved):

PR #482, opened 2025-03-12
Bail early on bad token

gnalh:
We shouldn't keep trying our endpoints if we've already detected an issue with the token / org combination.

### DIFF
--- a/cli/src/context_quarantine.rs
+++ b/cli/src/context_quarantine.rs
@@ -160,7 +160,7 @@ pub async fn gather_quarantine_context(
     file_set_builder: &FileSetBuilder,
     failed_tests_extractor: Option<FailedTestsExtractor>,
     test_run_exit_code: Option<i32>,
-) -> QuarantineContext {
+) -> anyhow::Result<QuarantineContext> {
     let failed_tests_extractor = failed_tests_extractor.unwrap_or_else(|| {
         FailedTestsExtractor::new(
             &request.repo,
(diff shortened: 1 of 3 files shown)


cli-tests/src/upload.rs
```diff
@@ -793,3 +793,84 @@ async fn telemetry_does_not_impact_return() {
...
+    let state = mock_server_builder.spawn_mock_server().await;
+
+    let assert = CommandBuilder::upload(temp_dir.path(), state.host.clone())
+        .command()
+        .assert()
+        .failure();
+
+    // Verify only the quarantine config request was made
+    let requests = state.requests.lock().unwrap().clone();
+    assert_eq!(requests.len(), 0);
```
cmillar-trunk, line 821: Why are we expecting 0 requests instead of 1?

cli/src/context_quarantine.rs
```diff
@@ -238,19 +238,27 @@ pub async fn gather_quarantine_context(
...
+            Err(err) => {
+                log_error(
+                    &err,
+                    Context {
+                        base_message: Some("Unable to find quarantine config".into()),
+                        org_url_slug: request.org_url_slug.clone(),
+                    },
+                );
+                if is_user_error(&err) {
+                    println!("user error {}", err);
```
cmillar-trunk, line 252: Leftover from debugging?

cli-tests/src/upload.rs
gnalh: The request didn't go through because it raises an error above.

cli/src/context_quarantine.rs
gnalh, line 252:
```suggestion
```

cli-tests/src/upload.rs
cmillar-trunk: But how do you get the error from the state server without sending a request to it? Doesn't the cli have to send a request to get the 404 from the server?

gnalh: The logic in the mock_server is such that if there's an error during the request then this will not be populated if there is an error. I agree it's confusing given the name and we should move towards it always being populated.

cmillar-trunk (approved):

gnalh:
Closing this since it isn't as easy as we'd hoped

PR #484, opened 2025-03-18
RSpec quarantining support

gnalh:


api/src/message.rs
```diff
@@ -66,9 +66,7 @@ pub struct TelemetryUploadMetricsRequest {
 
 #[derive(Debug, Serialize, Clone, Deserialize, PartialEq)]
 pub struct PageQuery {
```
gnalh, line 68: The api updated the names before release.

cmillar-trunk (approved):

PR #486, opened 2025-03-19
Add support for parsing `Internal` type

gnalh:
Addresses a long standing bug where we assume everything is junitxml. We also need this distinction for RSpec quarantining.


bundle/src/files.rs
```diff
@@ -183,6 +196,7 @@ impl FileSet {
 pub enum FileSetType {
     #[default]
     Junit,
+    Internal,
```
gnalh, line 199: We could call this `bin` instead, but I think internal captures the intent better.

bundle/src/files.rs
```diff
@@ -48,10 +48,17 @@ impl FileSetBuilder {
...
-                    junit_wrapper.junit_path = PathBuf::from(junit_wrapper.junit_path)
-                        .join("**/*.xml")
-                        .to_string_lossy()
-                        .to_string();
+                    if junit_wrapper.junit_path.ends_with(".xml") {
+                        junit_wrapper.junit_path = PathBuf::from(junit_wrapper.junit_path)
+                            .join("**/*.xml")
+                            .to_string_lossy()
+                            .to_string();
+                    } else {
```
cmillar-trunk, line 56: Since the body of this block is assuming an extension of `.bin`, should we be if-elsing on `.bin` with an error (or log) on a catchall `else`? (Also applies for the if/else on line 116 this file.

gnalh: Yeah, good call. Let's throw an error.

bundle/src/files.rs
```diff
@@ -106,6 +118,11 @@ impl FileSetBuilder {
                     bundled_files,
                     junit_wrapper.junit_path.clone(),
                     junit_wrapper.status.clone(),
+                    if junit_wrapper.junit_path.ends_with(".xml") {
```
cmillar-trunk, line 121: Same if else-ing here?

cmillar-trunk (approved):

PR #490, opened 2025-03-19
Move away from using quick_junit directly

gnalh:
Not everything the CLI uses will be directly supported by `quick_junit`. We should move away from assuming the incoming failures are stored in junit and move towards the shared bindings type. This is already how we process this data within our services and this makes the CLI more aligned with it. This is required to support quarantining with rspec and reporting what has been quarantined.

### DIFF
--- a/cli/src/context_quarantine.rs
+++ b/cli/src/context_quarantine.rs
@@ -1,13 +1,22 @@
-use std::collections::HashMap;
+use std::{
+    collections::HashMap,
+    io::{BufReader, Read},
+};
 
 use api::{client::ApiClient, urls::url_for_test_case};
-use bundle::{FileSet, FileSetBuilder, QuarantineBulkTestStatus, Test};
+use bundle::{FileSet, FileSetBuilder, FileSetType, QuarantineBulkTestStatus, Test};
 use constants::{EXIT_FAILURE, EXIT_SUCCESS};
 use context::{
-    junit::{junit_path::JunitReportStatus, parser::JunitParser},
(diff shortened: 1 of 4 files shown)


gnalh:
No tests were added here because this should be a drop in replacement for the previous logic. Existing tests succeeding should give us confident that this works as intended.

cli/src/context_quarantine.rs
```diff
@@ -21,17 +30,13 @@ fn convert_case_to_test<T: AsRef<str>>(
...
-    let xml_string_to_string = |s: &quick_junit::XmlString| String::from(s.as_str());
-    let class_name = case.classname.as_ref().map(xml_string_to_string);
-    let file = case.extra.get("file").map(xml_string_to_string);
-    let timestamp_millis = case
-        .timestamp
-        .or(suite.timestamp)
-        .map(|t| t.timestamp_millis());
+    let class_name = case.classname.clone();
+    let file = case.extra().get("file").cloned();
+    let timestamp_millis = case.timestamp.or(suite.timestamp);
```
cmillar-trunk, line 39: nit: should this just be `timestamp` since we're not calling `.timestamp_millis()`?

cmillar-trunk (approved):

cli/src/context_quarantine.rs
gnalh: Looks like this should be millis. I fixed it, but I don't think it's used on our side. There's a text for the binding directly in the `context-py` tests, which this wouldn't have touched.

PR #496, opened 2025-03-20
[bazel-bep] Properly support the test subcommand

gnalh:
We were checking for the bep.json before the command was invoked, which means it didn't exist and we wouldn't be able to locate tests. Test cases missed this because they use a stable bazel-bep file. I've confirmed that the test added fails on main.


cli/src/test_command.rs
```diff
@@ -55,13 +55,13 @@ pub async fn run_test(
     }: TestArgs,
 ) -> anyhow::Result<i32> {
     let token = upload_args.token.clone();
-    let pre_test_context = gather_pre_test_context(
+    tracing::info!("running command: {:?}", command);
+    let mut test_run_result = run_test_command(&command).await?;
+    let test_context = gather_pre_test_context(
```
gnalh, line 60: Should this be renamed? It doesn't seem to match the intent anymore, but I don't know what the original goal was for it.

mmatheson (approved):

cli/src/test_command.rs
cmillar-trunk: Yeah, it should definitely be renamed - it's gathering context after the test runs, and the `pre_test` makes it look like having it placed here looks like a bug.

cmillar-trunk (approved):

TylerJang27 (approved):
Thanks for the fix

cli/src/test_command.rs
gnalh: renaming `gather_test_context`

PR #498, opened 2025-03-26
Add ability to hide flaky tests banner

gnalh:


cli/src/upload_command.rs
```diff
@@ -117,6 +117,17 @@ pub struct UploadArgs {
...
+        long,
+        help = "Hide the top-level flaky tests banner",
+        action = ArgAction::Set,
+        required = false,
+        require_equals = true,
+        num_args = 0..=1,
+        default_value = "false",
+        default_missing_value = "true",
+    )]
+    pub hide_banner: bool,
```
cmillar-trunk, line 130: nit: should this be global so that it can apply to validate commands as well?

cmillar-trunk (approved):

cli/src/upload_command.rs
gnalh: Skipped validate since it is only run in local contexts, but I'll add it as well.

PR #504, opened 2025-03-27
Remove tags from being set

gnalh:
This functionality is unused (and always has been) and will be covered by a new flag.


cli/src/upload_command.rs
```diff
@@ -162,6 +163,11 @@ pub async fn run_upload(
     pre_test_context: Option<PreTestContext>,
     test_run_result: Option<TestRunResult>,
 ) -> anyhow::Result<UploadRunResult> {
+    if upload_args.tags.len() > 0 {
+        tracing::error!(
```
gnalh, line 167: This is intentionally an error so it reports in sentry.

cmillar-trunk (approved):

PR #507, opened 2025-03-28
Add variant support in the CLI

gnalh:


bundle/src/bundle_meta.rs
```diff
@@ -121,6 +121,31 @@ pub struct BundleMetaV0_6_3 {
...
+#[cfg_attr(feature = "wasm", derive(Tsify))]
+pub struct BundleMetaV0_7_6 {
+    #[serde(flatten)]
+    pub base_props: BundleMetaBaseProps,
+    #[serde(flatten)]
+    pub junit_props: BundleMetaJunitProps,
+    #[serde(flatten)]
+    pub debug_props: BundleMetaDebugProps,
+    pub bundle_upload_id_v2: String,
+    pub variant: Option<String>,
```
gnalh, line 135: I didn't see a good way to add variant to the `base_props` with our versioning system. It looks like the process is to move it out (similar to `bundle_upload_id_v2`. @TylerJang27 let me know if there's a better way here.

cmillar-trunk (approved):

PR #508, opened 2025-03-28
Do not truncate the command printed

gnalh:
Added a test that validates this logic and I've verified that it fails on main.

### DIFF
--- a/cli-tests/src/test.rs
+++ b/cli-tests/src/test.rs
@@ -72,6 +72,62 @@ async fn test_command_fails_with_successful_upload() {
 }
 
 #[tokio::test(flavor = "multi_thread")]
+async fn test_run_test_with_really_long_command() {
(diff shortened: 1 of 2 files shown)


cmillar-trunk (approved):

cli-tests/src/test.rs
```diff
@@ -71,6 +71,62 @@ async fn test_command_fails_with_successful_upload() {
     .code(1);
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn test_run_test_with_really_long_command() {
+    let temp_dir = tempdir().unwrap();
```
gnalh, line 76: I'm going to make this a unit test instead. As-is, this is overkill.

gnalh: Nevermind, we grab directly from env:args which makes this tricky.

PR #514, opened 2025-04-01
Deprecate tags

gnalh:
The first attempt had to be reverted because the upload processor couldn't handle the field being missing. We need to handle this in a series of PRs instead of all at once.


cli/src/upload_command.rs
```diff
@@ -78,7 +79,7 @@ pub struct UploadArgs {
         help = "Run metrics CLI without uploading to API."
     )]
     pub no_upload: bool,
-    #[arg(long, help = "Value to tag team owner of upload.")]
+    #[arg(long, help = "Value to tag team owner of upload.", hide = true)]
```
cmillar-trunk, line 82: Are we also deprecating the `--team` flag?

gnalh: Yes, it was never used.

PR #517, opened 2025-04-02
[rspec] Make generated description identification less strict

gnalh:
There is
`is expected to include`
`is expected to receive`
`is expected to redirect`

etc...

### DIFF
--- a/context-ruby/lib/trunk_spec_helper.rb
+++ b/context-ruby/lib/trunk_spec_helper.rb
@@ -29,7 +29,7 @@ def escape(str)
 end
 
 def description_generated?(example)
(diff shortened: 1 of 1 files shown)


context-ruby/lib/trunk_spec_helper.rb
```diff
@@ -29,7 +29,7 @@ def escape(str)
 end
 
 def description_generated?(example)
-  auto_generated_exp = /^\s?is expected to eq .*$/
+  auto_generated_exp = /^\s?is expected to.*$/
```
gnalh, line 32:
```suggestion
  auto_generated_exp = /^\s?is expected to .*$/
```

cmillar-trunk (approved):

PR #529, opened 2025-04-08
Pre-parse junit uploads

gnalh:


cli/src/context.rs
```diff
@@ -194,6 +196,50 @@ pub fn gather_post_test_context<U: AsRef<Path>>(
...
+                        junit_parser.parse(BufReader::new(file_contents.as_bytes()));
+                    if let Err(e) = parsed_results {
+                        tracing::warn!("Failed to parse JUnit file {:?}: {:?}", file, e);
+                        continue;
+                    }
+                    test_case_runs.extend(junit_parser.into_test_case_runs());
+                }
+            }
+        }
+        // todo - internal is already one file
```
cmillar-trunk, line 220: Is this something we'll need to resolve in the future?

gnalh: The comment actually isn't valid. It was a holdover from when I was imlpementing this and I ended up going in a different direction. Removing it.

cmillar-trunk (approved):

PR #539, opened 2025-04-11
remove invalid id check

gnalh:
This hasn't been valid for a long time. Clean it up so that we stop populating it in the table for rspec. This check is `suboptimal` and doesn't impact uploads / isn't visible in the UI. It was added for internal tracking only but is not needed.


cmillar-trunk:
Is this uuid just not used at all? It feels odd to be removing validation if the field is still being used.

gnalh:
> Is this uuid just not used at all? It feels odd to be removing validation if the field is still being used.

I added this check when we first added on xcresult support. It's no longer valid and probably shouldn't have ever been added. We support sending non uuidv5 ids (the ETL generates them).

cmillar-trunk:
> I added this check when we first added on xcresult support. It's no longer valid and probably shouldn't have ever been added. We support sending non uuidv5 ids (the ETL generates them).

In that case, should we be validating that it is at least a uuid (or do we intend to accept literally any string)?

gnalh:
It doesn't have to be a uuid

PR #546, opened 2025-04-15
Allow defining fallback exit code

gnalh:
This mimics the behavior of `test` in `upload` by allowing users to set the exit code of the previous result.

### DIFF
--- a/cli/src/context.rs
+++ b/cli/src/context.rs
@@ -315,7 +315,7 @@ pub async fn gather_exit_code_and_quarantined_tests_context(
     disable_quarantining: bool,
     api_client: &ApiClient,
     file_set_builder: &FileSetBuilder,
-    test_run_result: &Option<TestRunResult>,
+    default_exit_code: Option<i32>,
 ) -> i32 {
     // Run the quarantine step and update the exit code.
     let failed_tests_extractor = FailedTestsExtractor::new(
@@ -332,9 +332,9 @@ pub async fn gather_exit_code_and_quarantined_tests_context(
             },
     } = if disable_quarantining {
         // use the exit code of the test run result if exists
-        if let Some(test_run_result) = test_run_result {
+        if let Some(exit_code) = default_exit_code {
             QuarantineContext {
-                exit_code: test_run_result.exit_code,
+                exit_code,
                 quarantine_status: QuarantineBulkTestStatus {
                     quarantine_results: failed_tests_extractor
                         .failed_tests()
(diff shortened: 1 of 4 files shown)


cli/src/upload_command.rs
```diff
@@ -130,6 +130,14 @@ pub struct UploadArgs {
...
     )]
     pub variant: Option<String>,
+    #[arg(
+        long,
+        help = "The exit code to use when not all tests are quarantined.",
+        required = false,
+        num_args = 1,
+        hide = true
+    )]
+    pub previous_exit_code: Option<i32>,
```
gnalh, line 140: Open to a better name here. I had `fallback_exit_code` and `default_exit_code`, but didn't like either option. It's currently hidden while we experiment.

cmillar-trunk: Perhaps `override_exit_code_with`, since it's more explicit about what you should expect to have happen when you set it? (If not, I have to admit I prefer either option in your comment over `previous`, just because previous is describing it in the context of a workflow run, and not describing what it's for).

cli/src/context.rs
```diff
@@ -377,7 +377,7 @@ pub async fn gather_exit_code_and_quarantined_tests_context(
             },
             file_set_builder,
             Some(failed_tests_extractor),
-            test_run_result.as_ref().map(|t| t.exit_code),
+            default_exit_code,
```
cmillar-trunk, line 380: Should this `orElse` to the `test_run_result.exit_code`, so that we still use that if the default_exit_code isn't set?

gnalh: No, this doesn't need the `test_run_result` and never really did. That is and should be handled higher up in the call stack.

cli/src/upload_command.rs
gnalh: >just because previous is describing it in the context of a workflow run
But isn't that exactly what this is for? We want to allow users to propagate the exit code or result of a preceding job from their test run when they are not wrapping the underlying command.

cmillar-trunk: That's one thing it can be used for, and our intended use for it, but saying 'this is the previous code' doesn't tell us about how it's used - if someone is looking at arguments to solve a problem where they want to modify the exit code, they need to trawl through help because `previous_exit_code` does not tell you the behaviour of this flag. They need to infer that `previous` -> 'overrides a standard exit code' , or else trawl through the help for all arguments. It's harder to discover, and trawling through the whole help text is not a thing people will do, as thousands of stackoverflow questions about using bash commands demonstrate.

cmillar-trunk (approved):

PR #552, opened 2025-04-18
[TRUNK-14636] Allow overriding local repos

cmillar-trunk:
Some customers run tests where they do not check out the repo under test at all in their workflow (an integration test that is testing a backend service by deploying it and then hitting its endpoints, for example). In order to support this, we add the ability to entirely override local repository reading.


cli/src/upload_command.rs
```diff
@@ -138,6 +138,30 @@ pub struct UploadArgs {
...
+        default_value = "false",
+        default_missing_value = "true",
+        requires = "repo_url",
+        requires = "repo_head_sha",
+        requires = "repo_head_branch",
+        requires = "repo_head_commit_epoch",
+        requires = "repo_head_author_name",
+        conflicts_with = "repo_root"
+    )]
+    pub use_uncloned_repo: bool,
```
gnalh, line 164: I think it would be preferable to have this either be hidden or under an environment variable or both. This is an internal flag that I _believe_ we want to hide and only use within the context of custom integrations (such as the GitHub action).

cmillar-trunk: I've made it hidden

gnalh (approved):

PR #555, opened 2025-04-22
Address timestamp issue with internal binaries

gnalh:
* Corrects timestamp
* Handle empty classname

### DIFF
--- a/context/src/junit/parser.rs
+++ b/context/src/junit/parser.rs
@@ -958,7 +958,22 @@ mod tests {
             assert_eq!(run_binding.time, report_binding.time);
             assert_eq!(run_binding.system_out, report_binding.system_out);
             assert_eq!(run_binding.system_err, report_binding.system_err);
-            //assert_eq!(run_binding.properties, report_binding.properties);
+            // check that the properties match
+            for property in run_binding.properties.iter() {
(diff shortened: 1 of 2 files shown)


cmillar-trunk (approved):

context/src/junit/parser.rs
```diff
@@ -898,4 +898,67 @@ mod tests {
...
+            assert_eq!(run_binding.status.status, report_binding.status.status);
+            assert_eq!(run_binding.timestamp, report_binding.timestamp);
+            assert_eq!(
+                run_binding.timestamp_micros,
+                report_binding.timestamp_micros
+            );
+            assert_eq!(run_binding.time, report_binding.time);
+            assert_eq!(run_binding.system_out, report_binding.system_out);
+            assert_eq!(run_binding.system_err, report_binding.system_err);
+            //assert_eq!(run_binding.properties, report_binding.properties);
```
dfrankland, line 961: still needed?

dfrankland (approved):
tests make it easy to understand. nice

context/src/junit/parser.rs
gnalh: Good catch, forgot to add the check here.

PR #563, opened 2025-05-01
[rspec] Add ability to dump internal bin 

gnalh:


test_report/src/report.rs
```diff
@@ -269,12 +269,31 @@ impl MutTestReport {
...
+            .unwrap_or(named_temp_file.path().to_path_buf());
+        let resolved_path = match self.save(desired_path.clone()) {
+            Ok(path) => path,
+            Err(e) => {
+                tracing::error!("Error saving test results: {:?}", e);
+                return false;
+            }
+        };
+        if trunk_local_upload_dir.is_ok() {
+            println!("Saved test results to {:?}. Please run `trunk flakytests upload` to upload the results.", resolved_path);
```
cmillar-trunk, line 293: It feels really odd that a function called `publish` now explicitly doesn't send data depending on the status of an env var (which is not visible in the signature). Is is possible to move the 'do we publish or do we save locally' decision up the call stack to make it more visible?

gnalh: Fair enough, updated this to be called directly by the ruby code.

cmillar-trunk (approved):

PR #566, opened 2025-05-02
[TRUNK-14827] Add test for gitlab env vars

cmillar-trunk:
Adds testing for gitlab env vars, so that we can track this on main as well.
Screenshot from 2025-05-06 13-01-39

### DIFF
--- a/.github/actions/perform_gitlab_smoke_test/action.yaml
+++ b/.github/actions/perform_gitlab_smoke_test/action.yaml
@@ -29,7 +29,7 @@ runs:
         GITHUB_ACTIONS: false
         GITLAB_CI: true
         CI_JOB_URL: mock_gitlab.com/${{ env.GITHUB_REPOSITORY }}/actions/runs/${{ env.GITHUB_RUN_ID }}
-        GI_COMMIT_REF_NAME: ${{ env.GITHUB_REF }}
+        CI_COMMIT_REF_NAME: ${{ env.GITHUB_REF }}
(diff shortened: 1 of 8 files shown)


.github/actions/perform_gitlab_smoke_test/action.yaml
```diff
@@ -0,0 +1,118 @@
...
+        CI_COMMIT_MESSAGE: mock_commit_message
+        CI_MERGE_REQUEST_TITLE: ${{ env.PR_TITLE }}
+        CI_JOB_NAME: ${{ env.GITHUB_WORKFLOW }}
+        CI_JOB_STAGE: ${{ env.GITHUB_JOB }}
+      shell: bash
+      run: |
+        ./${{ inputs.cli-binary-location }} test \
+        --org-url-slug trunk-staging-org \
+        --junit-paths ${{ github.workspace }}/target/**/*junit.xml \
+        --token ${{ inputs.staging-api-token }} \
```
gnalh, line 46: This also needs the repo-url override provided since we interact with the repo service using the repo extracted from it.

cmillar-trunk: Gotcha, would you recommend using an obviously default gitlab url (eg `https://gitlab.com/gitlab-org/gitlab`, or spoof one for trunk)? Ideally I'd do the spoof, but I worry that that opens us to a namespace squatting issue since this is going into prod & staging.

gnalh (approved):

PR #575, opened 2025-05-08
init

gnalh:


xcresult/xcrun-xcresulttool-formatDescription-get---format-json---legacy-json-schema.json
```diff
@@ -632,6 +632,21 @@
...
+            },
+            "_values": {
+              "type": "array",
+              "items": {
+                "$ref": "#/$defs/ActionTestFailureSummary"
+              }
+            }
+          },
+          "required": ["_values"]
+        },
```
gnalh, line 649: There is no way to comment this directly in the json, but this is manually added. These values are not surfaced by xcrun's tooling but are always there.

xcresult/src/xcresult_legacy.rs
```diff
@@ -24,7 +24,58 @@ pub struct XCResultTestLegacy {
...
+                                            call_stack
+                                                .symbol_info
+                                                .as_ref()
+                                                .and_then(|symbol_info| {
+                                                    symbol_info.location.as_ref().and_then(
+                                                        |location| location.file_path.as_ref(),
+                                                    )
+                                                })
+                                                .map(|file_path| {
+                                                    file_path.value.replace(' ', "%20")
```
dfrankland, line 59: optional: we may want to consider URL decoding this instead of only replacing spaces

Cargo.lock
dfrankland, line 1: This lock file update seems unrelated to your changes

xcresult/src/xcresult_legacy.rs
gnalh: Yeah I thought the same thing, but the uri libraries fail on this input since it is t a valid uri.

xcresult/src/main.rs
```diff
@@ -39,6 +39,7 @@ fn main() -> anyhow::Result<()> {
         path,
         org_url_slug.unwrap_or_default(),
         repo_url_parts.repo_full_name(),
+        true,
```
cmillar-trunk, line 42: Why is this main function hardcoding the feature toggle to true?

gnalh: That's an internal only binary used for running the xcresults and seeing the output. Customers don't use this, just us.

dfrankland (approved):
neat

PR #583, opened 2025-05-20
[codeowners] Store the codeowners in the internal representation

gnalh:
Part 1 - Add code owners into the internal reports. The follow-up (part 2) will handle surfacing this information to the ETL using the bindings report.


context/src/junit/parser.rs
```diff
@@ -244,12 +245,46 @@ impl JunitParser {
...
                         .get(extra_attrs::FILE)
                         .or_else(|| test_case.extra.get(extra_attrs::FILEPATH))
                         .map(|v| v.to_string())
                         .unwrap_or_default();
+                    if !file.is_empty() && codeowners.is_some() {
+                        let codeowners: Vec<CodeOwner> = codeowners
+                            .as_ref()
+                            .map(|o| match o.owners.as_ref() {
+                                Some(Owners::GitHubOwners(owners)) => {
+                                    let owners = owners.of(file.clone()).unwrap_or_default();
```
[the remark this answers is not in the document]
mmatheson: @gemini-code-assist

> this will likely cause a compile error, or if it compiles due to some specific trait implementations not visible here, it would panic if owners.of() returns an Err, as codeowners::Error does not implement Default.

Your statement here is confidently incorrect. -- especially considering all of the tests pass.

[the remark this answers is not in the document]
gnalh: @gemini-code-assist Again, you are incorrect. The `of` function does not accept a reference to a string and therefore requires the clone occur. In order to pass along a reference, the function parameters and usage must be updated.

test_report/src/report.rs
```diff
@@ -305,7 +315,7 @@ impl MutTestReport {
             token,
             org_url_slug,
             vec![resolved_path_str.into()],
-            repo_root,
+            None,
```
cmillar-trunk, line 318: Why do we no longer pass the repo_root here?

test_report/src/report.rs
```diff
@@ -411,7 +421,41 @@ impl MutTestReport {
...
-        test.file = file;
+        // trunk-ignore(clippy/assigning_clones)
+        test.file = file.clone();
+        if !test.file.is_empty() {
+            test.codeowners = self
+                .0
+                .borrow_mut()
+                .codeowners
+                .as_ref()
+                .map(|o| match o.owners.as_ref() {
```
cmillar-trunk, line 432: The closure in this map looks the same as in junit/parser.rs:257. Can it be extracted out so we only need to update this once if another ci adds their own codeowners?

test_report/src/report.rs
gnalh: It isn't needed and was a holdover from no longer used code.

cmillar-trunk (approved):

PR #594, opened 2025-05-22
Remove the quarantine command

gnalh:
This was never broadcast out and hasn't been used. Remove it because the usage is currently ambiguous and we don't want to support it.


cmillar-trunk (approved):
You can probably also remove the quarantine branch of the command_builder in cli-tests now, too.

PR #606, opened 2025-05-27
Migrate to superconsole

gnalh:


cli/src/main.rs
```diff
@@ -210,16 +219,13 @@ fn setup_logger(
...
-        .with_level(false)
+        .with_target(true)
+        .with_level(true)
         .with_writer(std::io::stdout.with_max_level(to_trace_filter(log_level_filter)))
         .with_filter(FilterFn::new(|metadata| {
-            !metadata
-                .fields()
-                .iter()
-                .any(|field| field.name() == "hidden_in_console")
+            metadata.level() == &tracing::Level::DEBUG || metadata.level() == &tracing::Level::TRACE
```
gnalh, line 228: If you run with `-v` or `-vv`, we'll show all of the tracing logs. By default, they will be hidden and we'll only show superconsole output.

cli/src/main.rs
```diff
@@ -210,16 +219,13 @@ fn setup_logger(
         }
     });
 
+    // make console layer toggle based on vebosity
     let console_layer = tracing_subscriber::fmt::Layer::new()
-        .without_time()
-        .with_target(false)
-        .with_level(false)
+        .with_target(true)
+        .with_level(true)
```
gnalh, line 225: The goal is to make these logs more verbose since we are no longer showing them by default.

cmillar-trunk (approved):
Cleaning up the error reporting to get to just use Results is really nice.

PR #612, opened 2025-05-29
re-add print-files but as a no-op

gnalh:
We had a customer using this flag because it was mentioned in our docs. That reference has been dropped and we'll track the usage of it in sentry.


cmillar-trunk (approved):

cli/src/upload_command.rs
```diff
@@ -206,6 +213,12 @@ pub async fn run_upload(
         chrono::Utc::now().into()
     };
 
+    if upload_args.print_files.is_some() {
+        tracing::error!(
+            "The --print-files flag is deprecated and will be removed in a future version."
+        );
+    }
```
gnalh, line 220: This will track usage in sentry.

PR #620, opened 2025-06-03
Parse internal bin from tarball

pv72895:
Similar to `parse_metadata_from_tarball`, this PR aims to introduce a rust method that can be used to extract and parse the internal.bin file from an upload (as opposed to having consumers extract the tarball from the S3 stream themselves)

### DIFF
--- a/bundle/src/bundler.rs
+++ b/bundle/src/bundler.rs
@@ -10,6 +10,8 @@ use async_tar_wasm::Archive;
 use codeowners::CodeOwners;
 use context::bazel_bep::common::BepParseResult;
 use futures_io::AsyncBufRead;
+use prost::Message;
+use proto::test_context::test_run::TestResult;
 use tempfile::TempDir;
 #[cfg(feature = "wasm")]
 use tsify_next::Tsify;
@@ -27,6 +29,7 @@ pub struct BundlerUtil {
 }
 
 const META_FILENAME: &str = "meta.json";
+const INTERNAL_BIN_FILENAME: &str = "internal.bin";
 
 impl BundlerUtil {
     const ZSTD_COMPRESSION_LEVEL: i32 = 15; // This gives roughly 10x compression for text, 22 gives 11x.
@@ -123,9 +126,10 @@ impl BundlerUtil {
     }
 }
 
-/// Reads and decompresses a .tar.zstd file from an input stream into just a `meta.json` file
-///
-pub async fn parse_meta_from_tarball<R: AsyncBufRead>(input: R) -> anyhow::Result<VersionedBundle> {
+async fn parse_file_from_tarball<R: AsyncBufRead>(
+    input: R,
+    filename: &str,
+) -> anyhow::Result<Vec<u8>> {
     let zstd_decoder = ZstdDecoder::new(Box::pin(input));
     let archive = Archive::new(zstd_decoder);
 
@@ -138,13 +142,23 @@ pub async fn parse_meta_from_tarball<R: AsyncBufRead>(input: R) -> anyhow::Resul
             .to_owned();
 
         if path_str == META_FILENAME {
-            let mut meta_bytes = Vec::new();
-            owned_first_entry.read_to_end(&mut meta_bytes).await?;
+            let mut file_bytes = Vec::new();
+            owned_first_entry.read_to_end(&mut file_bytes).await?;
 
-            return parse_meta(meta_bytes);
+            return Ok(file_bytes);
         }
     }
-    Err(anyhow::anyhow!("No meta.json file found in the tarball"))
+    Err(anyhow::anyhow!(format!(
+        "No file at {} found in the tarball",
+        filename
+    )))
+}
+
+/// Reads and decompresses a .tar.zstd file from an input stream into just a `meta.json` file
+///
+pub async fn parse_meta_from_tarball<R: AsyncBufRead>(input: R) -> anyhow::Result<VersionedBundle> {
+    let meta_bytes = parse_file_from_tarball(input, META_FILENAME).await?;
+    parse_meta(meta_bytes)
 }
 
 pub fn parse_meta(meta_bytes: Vec<u8>) -> anyhow::Result<VersionedBundle> {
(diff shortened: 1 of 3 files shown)


.trunk/trunk.yaml
```diff
@@ -122,11 +125,9 @@ tools:
   runtimes:
     - ruby
     - node
-    - rust
   enabled:
     - gh@2.62.0
     - pnpm@9.14.2
     # Update `RUST_SRC_PATH` path to `rust-src` if this is updated
     # IfChange
-    - rust-src@2024-05-01
```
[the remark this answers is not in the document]
TylerJang27: IfChange you mean?

.trunk/trunk.yaml
TylerJang27, line 1: What prompted these changes?

.trunk/trunk.yaml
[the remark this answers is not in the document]
pv72895: Yah

TylerJang27 (commented):
LGTM but also I'm rusty

.trunk/trunk.yaml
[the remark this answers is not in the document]
TylerJang27: trunk-toolbox needs to be enabled for that to work. `trunk check enable trunk-toolbox`

.trunk/trunk.yaml
pv72895: https://trunk-io.slack.com/archives/C08AEDGMZNH/p1748962373354059

bundle/src/bundler.rs
```diff
@@ -27,6 +29,7 @@ pub struct BundlerUtil {
 }
 
 const META_FILENAME: &str = "meta.json";
+const INTERNAL_BIN_FILENAME: &str = "internal.bin";
```
gnalh, line 32: Let's also make sure to update `generate_internal_file` to use this as well. I don't want them diverging.

bundle/src/bundler.rs
```diff
@@ -163,3 +171,37 @@ pub fn parse_meta(meta_bytes: Vec<u8>) -> anyhow::Result<VersionedBundle> {
     let base_bundle = serde_json::from_slice(&meta_bytes)?;
     Ok(VersionedBundle::V0_5_29(base_bundle))
 }
+
+/// Reads and decompresses a .tar.zstd file from an input stream into just a `meta.json` file.
+/// This assumes that the `meta.json` file will be the first entry in the tarball.
+pub async fn parse_meta_from_tarball<R: AsyncBufRead>(input: R) -> anyhow::Result<VersionedBundle> {
```
gnalh, line 177: Let me know if I'm missing something, but this function should just invoke `parse_file_from_tarball` -- specifying meta.json as the filename. We don't need to duplicate code here.

Also, the assumption at the top -- assuming that the meta.json file is the first entry -- looks like a footgun. Why is that necessary?

gnalh (commented):
We haven't worked much in that section of the code since taking over this repo, but it does look like `bundle` as a whole is missing tests. I understand the desire to not add new tests in a language you are unfamiliar with, but I think there's value in generating those tests -- especially when introducing new code. Here's an [example](https://github.com/trunk-io/analytics-cli/blob/2a00e8bb41cce668fcb0f3720a9dad849de12b3b/context/src/junit/bindings.rs#L986-L1293) of defining a unit test in file. If the test is more complex, it should be in a standalone `test` directory.

bundle/src/bundler.rs
TylerJang27: We guarantee that meta.json is the first thing in the tarball (IIRC we do have tests to that effect at least).

This allows for the most efficient loading of the bundle info during EarlyTL and ETL. At least that was the requirement when I first implemented, not sure if that's changed.

.trunk/trunk.yaml
dfrankland: I have to make some changes to the bindings and I ran into the same issues btw. Disabling Rust in Trunk tools and recompiling everything made everything work. I think it's outside the scope to fix these runtime issues and that disabling Rust in Trunk tools is the appropriate action.

bundle/src/bundler.rs
dfrankland: Agreed. I think if we make a test for `meta.json` being the first entry that's enough to ensure we don't have regression with overfetching / perf in EarlyTL. We do have a test, right? 👀

bundle/src/bundler.rs
gnalh: @pv72895 this should be addressed as well (in case you missed it)

bundle/src/bundler.rs
```diff
@@ -123,28 +126,33 @@ impl BundlerUtil {
...
-/// Reads and decompresses a .tar.zstd file from an input stream into just a `meta.json` file
-///
-pub async fn parse_meta_from_tarball<R: AsyncBufRead>(input: R) -> anyhow::Result<VersionedBundle> {
+/// Reads and decompresses a .tar.zstd file from an input stream into just the specified file.
+/// This does not assume any specific order of files in the tarball, so it will read through all entries
+/// until it finds the specified file.
+async fn parse_file_from_tarball<R: AsyncBufRead>(
+    input: R,
+    file_name: &str,
+) -> anyhow::Result<Vec<u8>> {
```
dfrankland, line 135: We're making some similar-ish code over in the ETL: #26797

A comment I made there that I think is relevant here:
https://github.com/trunk-io/trunk/pull/26797#discussion_r2125602708

tl;dr: If we are pulling and processing multiple files, we should split the operations that pull files and the operations for processing the files. Reason being that you don't want to re-read the tar multiple times potentially doubling the IO needed.

I can go deeper on how to execute this, but I figured you'd probably prefer to think through the solution.

context-js/src/lib.rs
```diff
@@ -134,3 +137,31 @@ pub async fn parse_meta_from_tarball(
...
+            body.into_stream()
+                .map_ok(|js_value| js_value.dyn_into::<Uint8Array>().unwrap_throw().to_vec())
+                .map_err(|_js_error| {
+                    std::io::Error::new(std::io::ErrorKind::Other, "failed to read")
+                })
+                .into_async_read(),
+        ),
+    };
+
+    let buf_reader = BufReaderAsync::new(async_read);
```
dfrankland, line 161: Would it make sense to make a reusable function with this copied code?

context-js/tests/parse_and_validate.test.ts
```diff
@@ -80,6 +81,7 @@ describe("context-js", () => {
 
     let parse_result = junit_parse(Buffer.from(validJunitXml, "utf-8"));
     let report = parse_result.report;
+    assert(report, "Report from parsing report should not be null");
```
dfrankland, line 84: Why not use `expect`?

pv72895: `expect` doesn't type narrow like `assert` does so it was just a way to not have to slap a `?` on everything

bundle/src/bundler.rs
pv72895: OK I pushed code that takes in a list of files and goes through the tarball once to extract them all (and stops early if it finds everything it needs) - please let me know if I missed the mark here

context-js/src/lib.rs
pv72895: Still have not done this yet - will make another pass tomorrow

gnalh (approved):
Thank you for adding the tests 🙏

PR #623, opened 2025-06-04
[TRUNK-15071] Surface file validation issues in upload

cmillar-trunk:
Adds a file validation report to the upload command's superconsole output.

### DIFF
--- a/cli/src/upload_command.rs
+++ b/cli/src/upload_command.rs
@@ -1,3 +1,4 @@
+use std::collections::BTreeMap;
 use std::env;
 
 use api::client::ApiClient;
@@ -7,6 +8,9 @@ use bundle::{BundleMeta, BundlerUtil};
 use clap::{ArgAction, Args};
 use constants::EXIT_SUCCESS;
 use context::bazel_bep::common::BepParseResult;
+use context::junit::validator::{
+    JunitReportValidation, JunitReportValidationFlatIssue, JunitValidationLevel,
+};
 use pluralizer::pluralize;
 use superconsole::{
     style::{style, Attribute, Color, Stylize},
@@ -224,6 +228,7 @@ pub struct UploadRunResult {
     pub error_report: Option<ErrorReport>,
     pub quarantine_context: QuarantineContext,
     pub meta: BundleMeta,
+    pub validations: BTreeMap<String, anyhow::Result<JunitReportValidation>>,
 }
 
 pub async fn run_upload(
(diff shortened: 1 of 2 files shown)


cli/src/upload_command.rs
```diff
@@ -416,6 +424,118 @@ impl Component for UploadRunResult {
...
+                            }
+                        }
+                        (true, true) => {
+                            output.push(Line::from_iter([
+                                Span::new_unstyled("✅ ")?,
+                                Span::new_styled(
+                                    format!("{file_name} Is Perfect").attribute(Attribute::Bold),
+                                )?,
+                            ]));
+                        }
```
gnalh, line 533: The log gave me a chuckle, but we can skip this, there may be a lot of "perfect files".

cli/src/upload_command.rs
```diff
@@ -416,6 +424,118 @@ impl Component for UploadRunResult {
...
+        ]));
+        output.push(Line::from_iter([Span::new_styled(
+            format!(
+                "  {}",
+                pluralize("file", self.validations.len() as isize, true)
+            )
+            .attribute(Attribute::Italic),
+        )?]));
+
+        for (file_name, validation_reports) in self.validations.iter() {
```
gnalh, line 440: I think it makes sense to have these validations show up at the top (above the test report). What do you think?

gnalh, line 440: Let's also only show the first 10 or so validation issues (dealer's choice on the exact count of issues). I worry someone is going to glob a bunch of unrelated xml and we spam the output. We do something similar in the test report above for test failures.

gnalh (commented):
Love it, just a few comments here.

cli/src/upload_command.rs
cmillar-trunk: Fair point, I'll put up a count of the perfect files (since I'm showing the number of files we check, I want to represent those files somehow so we don't have people reaching out about how 'the number of files doesn't match the number listed').

cli/src/upload_command.rs
cmillar-trunk: Will do, I'll go for first 5 just because there's fewer kinds of error (and we dedupe them in the parser).

cli/src/upload_command.rs
```diff
@@ -416,6 +426,125 @@ impl Component for UploadRunResult {
...
+                    }
+                }
+            }
+        }
+
+        if perfect_files > 0 {
+            output.push(Line::from_iter([
+                Span::new_unstyled("✅ ")?,
+                Span::new_styled(
+                    pluralize("perfect file", perfect_files as isize, true)
```
gnalh, line 541: Nit:
```suggestion
                    pluralize("valid file", perfect_files as isize, true)
```

gnalh (approved):

cli/src/upload_command.rs
cmillar-trunk: `valid` seems to imply that the files with warnings are not used; is there an alternative?

gnalh: We could also do
`X fully correct files`.

I don't have strong feelings here, we can leave this as is if you just want to merge it.

PR #630, opened 2025-06-05
Deprecate hide_banner flag

cmillar-trunk:
The hide_banner flag does nothing in the superconsole world, so it can be deprecated and have its logic removed.


gnalh (approved):
Thank you!

PR #633, opened 2025-06-06
[TRUNK-14885] Add test_reports argument

cmillar-trunk:
Adds a test_reports argument to upload & validate arguments, which attempts to parse as all valid types of file, falling back to junit if nothing else works. Since we can't know if an error is because there are no tests in a bazel file or it's actually invalid, files passed in test_reports will not raise if they have no test cases.


.github/actions/run_tests/action.yaml
```diff
@@ -24,4 +24,4 @@ runs:
 
     - name: Run tests
       shell: bash
-      run: cargo nextest run --features=bindings --workspace ${{ contains(inputs.target, 'musl') && '--exclude context-js --exclude context-py --exclude context_ruby' || '' }} --profile=ci
+      run: cargo nextest run --features=bindings --workspace ${{ contains(inputs.target, 'musl') && '--exclude context-js --exclude context-py --exclude context_ruby' || '' }} --profile=ci --nocapture
```
gnalh, line 27: why add this flag?

cli/src/context.rs
```diff
@@ -357,7 +370,7 @@ fn coalesce_junit_path_wrappers(
             &temp_dir,
             xcresult_path,
             repo,
-            org_url_slug,
+            org_url_slug.clone(),
```
gnalh, line 373: I wonder if we can update this to be a reference to remove this clone.

cli/src/context.rs
```diff
@@ -373,6 +386,50 @@ fn coalesce_junit_path_wrappers(
         }
     }
 
+    if !test_reports.is_empty() {
+        for test_report in test_reports {
+            let mut was_other_than_junit = false;
+
+            if let Ok(bazel_result) = parse_as_bep(test_report.clone()) {
```
gnalh, line 393: These set of conditionals seem brittle. It might be good to split it off into a separate function that explicitly identifies the goal. Consider also moving to using `if else` so that each code path is independent.

My main concern here is that we may end up editing this code path such that it has a bad assumption around what's being parsed.

.github/actions/run_tests/action.yaml
cmillar-trunk: It was added for debugging, deleted.

cli/src/context.rs
cmillar-trunk: embarrassingly easily.

cli/test_fixtures/bep_binary_file.bin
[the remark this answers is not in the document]
cmillar-trunk: I don't think so, I'm mirroring the file that was used in the context crate [here](https://github.com/trunk-io/analytics-cli/blob/main/context/test_fixtures/bep_binary_file.bin), and I don't think it's in lfs, is it?

PR #634, opened 2025-06-06
wip

dfrankland:


bundle/src/files.rs
```diff
@@ -172,22 +175,77 @@ pub struct FileSet {
...
-    pub resolved_status: Option<JunitReportStatus>,
+    pub resolved_status: TestRunnerReportStatus,
+    /// Added in v0.9.3. Populated when parsing from BEP, not from junit globs
+    #[cfg_attr(feature = "wasm", tsify(type = "number"))]
+    #[serde(default, with = "ts_milliseconds")]
+    pub resolved_start_time_epoch_ms: DateTime<Utc>,
+    /// Added in v0.9.3. Populated when parsing from BEP, not from junit globs
+    #[cfg_attr(feature = "wasm", tsify(type = "number"))]
+    #[serde(default, with = "ts_milliseconds")]
+    pub resolved_end_time_epoch_ms: DateTime<Utc>,
```
dfrankland, line 195: Break down of some of these proc-macro configs:

For wasm:
* `tsify(type = "number")` set the type to `number` instead of being auto-inferred as `DateTime<Utc>` which doesn't exist as a type in TypeScript
* `serde(default)` sets the default for this if it's not provided from JSON—it will default to `0` which is equivilant to the unix epoch
* `serde(with = "ts_milliseconds")` uses `chrono::serde::ts_milliseconds` for (de)serialization making managing the data much nicer

For python:
* `gen_stub_pyclass` now generates the type as `datetime.datetime` automatically and `pyclass` does the conversions

bundle/src/files.rs
```diff
@@ -172,22 +175,77 @@ pub struct FileSet {
...
+
+impl From<FileSetTestRunnerReport> for TestRunnerReport {
+    fn from(test_runner_report: FileSetTestRunnerReport) -> Self {
+        Self {
+            status: test_runner_report.resolved_status,
+            start_time: test_runner_report.resolved_start_time_epoch_ms,
+            end_time: test_runner_report.resolved_end_time_epoch_ms,
+        }
+    }
 }
```
dfrankland, line 235: Interop for `meta.json` and `context/junit` types. The split isn't just for field naming, but for (de)serialization implementation

cli-tests/src/upload.rs
```diff
@@ -235,14 +236,31 @@ async fn upload_bundle_using_bep() {
...
+                    - report.timestamp.unwrap().to_utc())
+                .abs()
+                    <= TimeDelta::milliseconds(1)
+            );
+            assert!(
+                (test_runner_report.resolved_end_time_epoch_ms
+                    - (report.timestamp.unwrap().to_utc() + report.time.unwrap()))
+                .abs()
+                    <= TimeDelta::milliseconds(1)
+            );
```
dfrankland, line 263: I split these out to prevent any issues with small nanosecond rounding issues when roundtripping gRPC timstamps

cli/src/context.rs
```diff
@@ -245,8 +245,13 @@ pub fn generate_internal_file(
...
                     if reports.len() == 1 {
-                        junit_validations
-                            .insert(file.original_path.clone(), Ok(validate(&reports[0])));
+                        junit_validations.insert(
+                            file.original_path.clone(),
+                            Ok(validate(
+                                &reports[0],
+                                file_set.test_runner_report.map(|t| t.into()),
+                            )),
+                        );
```
dfrankland, line 254: Most things in this file are just renaming / plumbing to pass the test runner report to `validate`

cli/src/context_quarantine.rs
dfrankland, line 1: Everything in here should just be renaming, moving around fields. No logic changes.

cli/src/validate_command.rs
```diff
@@ -75,20 +75,26 @@ pub async fn run_validate(validate_args: ValidateArgs) -> anyhow::Result<i32> {
...
-type JunitFileToReport = BTreeMap<String, Report>;
+type JunitFileToReportAndParseIssues = BTreeMap<
+    String,
+    (
+        anyhow::Result<Option<Report>>,
+        Vec<JunitParseIssue>,
+        Option<FileSetTestRunnerReport>,
+    ),
+>;
+type JunitFileToReport = BTreeMap<String, (Report, Option<FileSetTestRunnerReport>)>;
```
dfrankland, line 92: Everything in this file should just be plumbing to pass around the test runner report

context/src/bazel_bep/common.rs
```diff
@@ -2,44 +2,54 @@ use std::collections::HashMap;
...
+            bep_test_events: Vec<BuildEvent>,
+        }
+        let Acc {
+            errors,
+            test_results,
+            test_runner_reports,
+            bep_test_events,
+        } = events
+            .into_iter()
+            .fold(Acc::default(), |mut acc, parse_event| {
```
dfrankland, line 48: Using a tuple was getting a bit unwieldy here. I switched to a simple local struct since nothing needs access to the this struct and it helps with readability

context/src/bazel_bep/common.rs
```diff
@@ -49,12 +59,13 @@ impl BepParseResult {
                                 Some(Payload::TestSummary(test_summary)),
                                 Some(Id::TestSummary(id)),
                             ) => {
-                                if let Result::Ok(status) =
-                                    JunitReportStatus::try_from(test_summary.overall_status())
+                                if let Result::Ok(test_runner_report) =
+                                    TestRunnerReport::try_from(test_summary)
```
dfrankland, line 63: Updated to use `impl TryFrom<&TestSummary> for TestRunnerReport` below

context/src/bazel_bep/common.rs
```diff
@@ -2,44 +2,54 @@ use std::collections::HashMap;
...
 const FILE_URI_PREFIX: &str = "file://";
 
 #[derive(Debug, Clone, Default)]
-pub struct TestResult {
+pub struct BepTestResult {
     pub label: String,
     pub cached: bool,
     pub xml_files: Vec<String>,
-    pub summary_status: Option<JunitReportStatus>,
+    pub test_runner_report: Option<TestRunnerReport>,
```
dfrankland, line 20: Below changes are to plumb through the test runner report

context/src/junit/validator.rs
```diff
@@ -39,6 +42,15 @@ pub enum JunitValidationIssue<SO, I> {
...
 }
 
+impl<SO: fmt::Display, I: fmt::Display> fmt::Display for JunitValidationIssue<SO, I> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::SubOptimal(i) => write!(f, "{i}"),
+            Self::Invalid(i) => write!(f, "{i}"),
+        }
+    }
+}
```
dfrankland, line 52: Trunk Check was failing on a Clippy lint that disallows direct implementation of the `ToString`:
https://rust-lang.github.io/rust-clippy/master/index.html#to_string_trait_impl

It requires implementing `Display` instead which I did so in a generic way here so that we don't need to repeat ourselves below for each validation enum type

context/src/junit/validator.rs
```diff
@@ -201,23 +209,26 @@ pub fn validate(report: &Report) -> JunitReportValidation {
...
-            JunitValidationIssueType::Report(i) => i.to_string(),
-            JunitValidationIssueType::TestSuite(i) => i.to_string(),
-            JunitValidationIssueType::TestCase(i) => i.to_string(),
+            JunitValidationIssueType::Report(i) => write!(f, "{i}"),
+            JunitValidationIssueType::TestRunnerReport(i) => write!(f, "{i}"),
+            JunitValidationIssueType::TestSuite(i) => write!(f, "{i}"),
+            JunitValidationIssueType::TestCase(i) => write!(f, "{i}"),
         }
     }
 }
```
dfrankland, line 234: Same issue with Clippy as described above

context/src/junit/validator.rs
```diff
@@ -149,32 +174,15 @@ pub fn validate(report: &Report) -> JunitReportValidation {
...
-            }
+            validate_test_case_timestamp(
+                test_case.timestamp,
+                test_suite.timestamp,
+                report.timestamp,
+                test_runner_report,
+                now,
+            )
+            .into_iter()
+            .for_each(|i| test_case_validation.add_issue(JunitValidationIssue::SubOptimal(i)));
```
dfrankland, line 185: This logic was getting a little large for this function body, so it's been split out

bundle/Cargo.toml
dfrankland, line 1: All of these updates are to facilitate having `DateTime` binding interop declaratively defined. See `FileSetTestRunnerReport` in `bundle/src/files.rs` for more details.

cli/src/context_quarantine.rs
dfrankland, line 1: Everything in here should just be renaming, moving around fields. No logic changes.

PR #635, opened 2025-06-06
Loosen code owners parsing

gnalh:
If a relative path is provided, we should attempt to match it relative to the root of the repo. This will allow us to parse the corresponding code owner when a test reporter provides relative paths.

### DIFF
--- a/codeowners/src/github.rs
+++ b/codeowners/src/github.rs
@@ -96,8 +96,9 @@ impl OwnersOfPath for GitHubOwners {
                 if pattern.matches_path_with(path.as_ref(), opts) {
                     Some(owners)
                 } else {
-                    // if the path is relative, we need to strip the leading dot
-                    // to match the pattern
+                    // NOTE: if the path is relative, we need to strip the leading dot
+                    // to match the pattern. We are doing this as a workaround for
+                    // cases where the provided path is relative, like `./foo/bar/baz.rs`
                     if let Ok(simplified_path) = path.as_ref().strip_prefix("./") {
                         if pattern.matches_path_with(simplified_path, opts) {
                             return Some(owners);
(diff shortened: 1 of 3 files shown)


cmillar-trunk (approved):

codeowners/src/gitlab/file.rs
```diff
@@ -64,8 +64,14 @@ impl File {
     }
 
     pub fn entries_for_path(&self, path: String) -> Vec<Entry> {
+        let stripped_path = path.strip_prefix("./");
```
dfrankland, line 67: Shouldn't `.` be the only thing stripped since you are re-adding `/`?

codeowners/src/github.rs
```diff
@@ -368,6 +383,21 @@ mod tests {
...
+
+    #[test]
+    fn relative_path_owners() {
+        let owners = GitHubOwners::from_reader("foo/bar @doug".as_bytes()).unwrap();
+        assert_eq!(
+            owners.of("./foo/bar/baz.rs"),
+            Some(vec![GitHubOwner::Username("@doug".into())])
+        );
+        assert_eq!(owners.of(".foo/bar/baz.rs"), None,);
+        assert_eq!(owners.of("./"), None,);
```
dfrankland, line 397:
```suggestion
        assert_eq!(owners.of(".foo/bar/baz.rs"), None);
        assert_eq!(owners.of("./"), None);
```

dfrankland (approved):
It's possible this will over match because of the possibility of nested file structures. I can see this leading someone to confusion at a later point, so I think it might be worth it to document.

codeowners/src/gitlab/file.rs
dfrankland: I'm forgetful. I forgot hidden files

gnalh:
> It's possible this will over match because of nested file structures. I can see this leading someone to confusion at a later point, so I think it might be worth it to document.

Added a note inline to the comments about this being a workaround.

PR #636, opened 2025-06-06
part 1 update validation to mark timestamps as old when older than 24 hours

dfrankland:
First of a couple of PRs to update how we validate timestamps. The threshold has been reduced to 24 hours.

### DIFF
--- a/context/src/junit/validator.rs
+++ b/context/src/junit/validator.rs
@@ -15,7 +15,7 @@ use crate::string_safety::{validate_field_len, FieldLen};
 
 pub const MAX_FIELD_LEN: usize = 1_000;
 
-const TIMESTAMP_OLD_DAYS: u32 = 30;
+const TIMESTAMP_OLD_HOURS: u32 = 24;
 const TIMESTAMP_STALE_HOURS: u32 = 1;
 
 #[cfg_attr(feature = "pyo3", gen_stub_pyclass_enum, pyclass(eq, eq_int))]
@@ -161,7 +161,7 @@ pub fn validate(report: &Report) -> JunitReportValidation {
                     test_case_validation.add_issue(JunitValidationIssue::SubOptimal(
                         JunitTestCaseValidationIssueSubOptimal::TestCaseFutureTimestamp(timestamp),
                     ));
-                } else if time_since_timestamp.num_days() > i64::from(TIMESTAMP_OLD_DAYS) {
+                } else if time_since_timestamp.num_hours() > i64::from(TIMESTAMP_OLD_HOURS) {
                     test_case_validation.add_issue(JunitValidationIssue::SubOptimal(
                         JunitTestCaseValidationIssueSubOptimal::TestCaseOldTimestamp(timestamp),
                     ));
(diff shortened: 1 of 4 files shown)


context/src/junit/validator.rs
```diff
@@ -161,7 +161,7 @@ pub fn validate(report: &Report) -> JunitReportValidation {
                     test_case_validation.add_issue(JunitValidationIssue::SubOptimal(
                         JunitTestCaseValidationIssueSubOptimal::TestCaseFutureTimestamp(timestamp),
                     ));
-                } else if time_since_timestamp.num_days() > i64::from(TIMESTAMP_OLD_DAYS) {
+                } else if time_since_timestamp.num_days() > i64::from(TIMESTAMP_OLD_HOURS) {
```
epes, line 164: should this be `num_hours()`? if that even exists

context/src/repo/validator.rs
```diff
@@ -191,7 +194,7 @@ pub fn validate(bundle_repo: &BundleRepo) -> RepoValidation {
             repo_validation.add_issue(RepoValidationIssue::SubOptimal(
                 RepoValidationIssueSubOptimal::RepoCommitFutureTimestamp(timestamp),
             ));
-        } else if time_since_timestamp.num_days() > i64::from(TIMESTAMP_OLD_DAYS) {
+        } else if time_since_timestamp.num_days() > i64::from(TIMESTAMP_OLD_HOURS) {
```
epes, line 197: `num_hours` here as well

context/src/junit/validator.rs
```diff
@@ -15,7 +15,7 @@ use crate::string_safety::{validate_field_len, FieldLen};
 
 pub const MAX_FIELD_LEN: usize = 1_000;
 
-const TIMESTAMP_OLD_DAYS: u32 = 30;
+const TIMESTAMP_OLD_HOURS: u32 = 24;
```
gnalh, line 18: Nit: TIMESTAMP_MAX_HOURS

context/src/junit/validator.rs
gnalh: +1, looks like this needs to be updated

gnalh (commented):
Can you add a test for this?

context/src/junit/validator.rs
dfrankland: Woops, I didn't migrate this change correctly from my larger change :/

dfrankland:
@gnalh this is already tested in the `cli-tests`

dfrankland:
Hmmm, I'll reply to myself regarding tests: "If it's so well tested, Dylan, then why didn't it catch the change to use `.num_hours()`?"

That's a great question. Not good enough tests 👀 

I'll update them.

gnalh:
> @gnalh this is already tested in the `cli-tests`

We aren't testing that a day triggers this, just that old timestamps trigger it. A test for >24 hours specifically would have caught the issue raised in this PR around the if statement that was not updated appropriately for hours.

epes (approved):

gnalh (approved):

PR #638, opened 2025-06-06
part 2 update BEP tests to check for resolved status in meta.json

dfrankland:
A lot of tests were not passing the path to the BEP file which made the results incorrect. In addition, I added more checks to API calls and checked that `meta.json` contained a resolved status.

The next PRs will be adding more properties to the `meta.json` that will get used for validation.


cli-tests/src/test.rs
```diff
@@ -252,14 +249,14 @@ async fn test_command_succeeds_with_bundle_using_bep() {
     .failure();
 
     let requests = state.requests.lock().unwrap().clone();
-    assert_eq!(requests.len(), 4);
-
-    let tar_extract_directory = assert_matches!(&requests[2], RequestPayload::S3Upload(d) => d);
+    assert_eq!(requests.len(), 3);
+    assert_matches!(requests[0], RequestPayload::CreateBundleUpload(_));
+    let tar_extract_directory = assert_matches!(&requests[1], RequestPayload::S3Upload(d) => d);
+    assert_matches!(requests[2], RequestPayload::TelemetryUploadMetrics(_));
```
dfrankland, line 255: When there's only 3-4 requests, it makes sense to assert each of them. In this case I was confused about what changed and it would be nice to have a historical reference of the changes in the requests themselves.

cli-tests/src/test.rs
```diff
@@ -252,14 +249,14 @@ async fn test_command_succeeds_with_bundle_using_bep() {
...
 
     let junit_file = fs::File::open(tar_extract_directory.join("junit/0")).unwrap();
     let junit_reader = BufReader::new(junit_file);
 
-    // Uploaded file is a junit, even when using BEP
```
dfrankland, line 262: Most comments in tests have been for self-explanatory code adding to the noise. Additionally, some of them have been incorrect, stale, or added non-relevant context.

Removing them and making the test assertions clearer is better.

cli-tests/src/upload.rs
```diff
@@ -7,14 +7,14 @@ use api::message::{
...
+use axum::{extract::State, http::StatusCode, Json};
 use bundle::{BundleMeta, FileSetType, INTERNAL_BIN_FILENAME};
 use codeowners::CodeOwners;
 use constants::EXIT_FAILURE;
-use context::repo::{BundleRepo, RepoUrlParts};
 use context::{
-    bazel_bep::parser::BazelBepParser, junit::parser::JunitParser, repo::RepoUrlParts as Repo,
+    bazel_bep::parser::BazelBepParser,
+    junit::{junit_path::JunitReportStatus, parser::JunitParser},
+    repo::{BundleRepo, RepoUrlParts as Repo},
```
dfrankland, line 17: I've combined as many imports as I could.

cli-tests/src/upload.rs
```diff
@@ -210,28 +210,40 @@ async fn upload_bundle() {
...
 
     let state = MockServerBuilder::new().spawn_mock_server().await;
 
     let assert = CommandBuilder::upload(temp_dir.path(), state.host.clone())
-        .bazel_bep_path("./bep.json")
+        .bazel_bep_path(bep_path.to_str().unwrap())
         .command()
         .assert()
-        .failure();
+        .success();
```
dfrankland, line 221: It seems we weren't generating BEP files correctly and so the upload failed when it shouldn't have

cli-tests/src/upload.rs
```diff
@@ -331,57 +376,42 @@ async fn upload_bundle_success_preceding_failure() {
...
 
-    // HINT: View CLI output with `cargo test -- --nocapture`
-    println!("{assert}");
-}
-
-#[tokio::test(flavor = "multi_thread")]
-async fn upload_bundle_success_timestamp_status_code() {
-    let temp_dir = tempdir().unwrap();
-    generate_mock_git_repo(&temp_dir);
-    let test_bep_path = get_test_file_path("test_fixtures/bep_retries_timestamp");
```
dfrankland, line 355: This tests doesn't assert anything different from the other tests using the `bep_retries` file. Additionally, it didn't include `testSummary` which is present in BEP normally. I couldn't see a reason for keeping this test.

cli-tests/src/upload.rs
```diff
@@ -1059,9 +1079,11 @@ async fn test_uncloned_repo_requires_manual_settings() {
     let state = MockServerBuilder::new().spawn_mock_server().await;
 
     let assert = CommandBuilder::upload(temp_dir.path(), state.host.clone())
+        .bazel_bep_path(bep_path.to_str().unwrap())
         .use_uncloned_repo(true)
         .command()
         .assert()
+        .code(predicate::eq(2))
```
dfrankland, line 1086: Asserting only on `failure` is bad practice since we aren't checking what the `failure` is for.

cli-tests/src/utils.rs
```diff
@@ -33,49 +33,99 @@ pub fn generate_mock_git_repo<T: AsRef<Path>>(directory: T) {
...
+                        total_num_cached: 0,
+                        first_start_time: test_start_time.clone(),
+                        last_stop_time: Some(
+                            (report.timestamp.unwrap() + report.time.unwrap())
+                                .to_utc()
+                                .into(),
+                        ),
+                        total_run_duration: test_duration,
+                        ..Default::default()
+                    })),
```
dfrankland, line 113: Every BEP I've inspected includes the summary which we use to set the resolved status for tests. Additionally, this will be used for the start and end times in follow up PRs.

context-js/tests/parse_compressed_bundle.test.ts
```diff
@@ -27,6 +27,20 @@ const generateBundleMeta = (): TestBundleMeta => ({
...
+        {
+          original_path: "/abs/path/junit.xml",
+          original_path_rel: "junit.xml",
+          path: "0.xml",
+          owners: ["owner"],
+          team: "team",
+        },
+      ],
+      glob: "**/*.xml",
+      resolved_status: "Passed",
```
dfrankland, line 42: Added to check for positive case (negative being `null`)

cli-tests/src/upload.rs
```diff
@@ -247,8 +259,6 @@ async fn upload_bundle_success_status_code() {
     let temp_dir = tempdir().unwrap();
     generate_mock_git_repo(&temp_dir);
     let test_bep_path = get_test_file_path("test_fixtures/bep_retries");
-    // The test cases need not match up or have timestamps, so long as there is a testSummary
-    // That indicates a flake or pass
```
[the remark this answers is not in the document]
dfrankland: It's still true, it just doesn't help explain the test—it's not relevant

cli-tests/src/test.rs
```diff
@@ -252,14 +249,14 @@ async fn test_command_succeeds_with_bundle_using_bep() {
     .failure();
 
     let requests = state.requests.lock().unwrap().clone();
-    assert_eq!(requests.len(), 4);
-
-    let tar_extract_directory = assert_matches!(&requests[2], RequestPayload::S3Upload(d) => d);
+    assert_eq!(requests.len(), 3);
```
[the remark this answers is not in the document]
dfrankland: Looks like previously this called `GetQuarantineBulkTestStatus`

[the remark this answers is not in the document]
gnalh: Where was that logic changed? It looks like this PR only updates tests?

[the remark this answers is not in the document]
gnalh: Ah that makes sense. We skip the quarantine check because bazel said all is well.

gnalh (approved):

cli-tests/src/upload.rs
dfrankland: `exitcode` doesn't define 2 as a constant:
https://github.com/benwilber/exitcode/blob/5d18ab34e680180b0c032707c5a42592a075af95/src/lib.rs

This code comes directly from `clap` and isn't publicly accessible:
https://github.com/clap-rs/clap/blob/d117f7acdeedebaf5fd7847debb15c834423f159/clap_builder/src/util/mod.rs#L25-L30

PR #639, opened 2025-06-06
remove non-null assertions

dfrankland:
Sorry, Phil! I like your answer about narrowing types :/

### DIFF
--- a/context-js/tests/parse_and_validate.test.ts
+++ b/context-js/tests/parse_and_validate.test.ts
@@ -1,9 +1,8 @@
-/* eslint-disable @typescript-eslint/no-non-null-assertion */
 import dayjs from "dayjs";
 import fs from "fs";
 import path from "path";
 import utc from "dayjs/plugin/utc";
-import { describe, expect, it } from "vitest";
+import { assert, describe, expect, it } from "vitest";
 
 import {
   BundleRepo,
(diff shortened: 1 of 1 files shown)


context-js/tests/parse_and_validate.test.ts
```diff
@@ -214,7 +205,7 @@ describe("context-js", () => {
   it("parses internal.bin", () => {
     expect.hasAssertions();
 
-    const file_path = path.resolve(__dirname, "../tests/test_internal.bin");
+    const file_path = path.resolve(__dirname, "./test_internal.bin");
```
dfrankland, line 208: redundant pathpart

context-js/tests/parse_and_validate.test.ts
```diff
@@ -42,9 +41,7 @@ describe("context-js", () => {
     };
 
     const ciInfo = env_parse(env_vars, ["main", "master"]);
-    // NOTE: Need to narrow type here
-    // eslint-disable-next-line vitest/no-conditional-in-test
-    if (!ciInfo) throw Error("ciInfo is undefined");
+    assert(ciInfo);
```
gnalh, line 44: This is fine as is, just want to note that asserts without a message are harder to debug. Since this is just in a test, and we aren't consistent about it in tests, this is fine as is.

gnalh (approved):

context-js/tests/parse_and_validate.test.ts
dfrankland: I would never use `assert` in the actual code. IMO, it only belongs in tests and in this instance the error message is the same as what the `assert` will show anyways.

PR #647, opened 2025-06-09
[tui] Update logging

gnalh:
* add logs for the case when no failures found, but exit code was 0
* use `terminal_size` to try and find the terminal size

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -3,6 +3,7 @@ use std::env;
 use clap::{Parser, Subcommand};
 use clap_verbosity_flag::{log::LevelFilter, InfoLevel, Verbosity};
 use superconsole::{Dimensions, SuperConsole};
+use terminal_size::{terminal_size, Height, Width};
 use third_party::sentry;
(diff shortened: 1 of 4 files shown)


cli/src/main.rs
```diff
@@ -110,10 +111,25 @@ fn main() -> anyhow::Result<()> {
...
-                        height: 24,
-                    })
+                    let size = terminal_size();
+                    if let Some((Width(width), Height(height))) = size {
+                        SuperConsole::forced_new(Dimensions {
+                            width: width as usize,
+                            height: height as usize,
+                        })
+                    } else {
+                        tracing::warn!("Falling back to default dimensions for superconsole");
```
cmillar-trunk, line 121: Will these actually show to the user? When I've had issues with superconsole, tracing logs just didn't show up at all, and the docs warn against using multiple console outputs.

gnalh: These will only show when using the verbose flag. -v

cmillar-trunk (approved):

PR #650, opened 2025-06-09
[wip] Add missing code coverage

gnalh:


context-py/src/lib.rs
```diff
@@ -61,7 +61,7 @@ fn branch_class_to_string(branch_class: env::parser::BranchClass) -> String {
 #[gen_stub_pyfunction]
 #[pyfunction]
 fn ci_platform_to_string(ci_platform: env::parser::CIPlatform) -> String {
-    ci_platform.to_string()
+    ci_platform.to_string().into()
```
gnalh, line 64: Was seeing this error when running the code coverage report
```
error[E0308]: mismatched types
  --> context-py/src/lib.rs:64:5
   |
63 | fn ci_platform_to_string(ci_platform: env::parser::CIPlatform) -> String {
   |                                                                   ------ expected `String` because of return type
64 |     ci_platform.to_string()
   |     ^^^^^^^^^^^^^^^^^^^^^^^ expected `String`, found `&str`
   |
help: call `Into::into` on this expression to convert `&str` into `String`
   |
64 |     ci_platform.to_string().into()
   |                            +++++++
```

PR #651, opened 2025-06-09
part 3 add test runner report

dfrankland:
TRUNK-15041

Split from #634

Plumbs through the test runner report (Bazel BEP test summary) to allow more accurate validation of JUnit reports.


bundle/Cargo.toml
dfrankland, line 1: All of these updates are to facilitate having `DateTime` binding interop declaratively defined. See `FileSetTestRunnerReport` in `bundle/src/files.rs` for more details.

gnalh (approved):

PR #661, opened 2025-06-11
[xcresult] Do not show xml warnings

gnalh:
When using xcresult, we parse it into xml and then upload that. The xml generated has filepaths missing for tests that succeeded (we only get them on failures). Users can't control this, so we should hide these warnings for xcresult.


dfrankland:
Not sure if it's possible, but I think we can use the same proc macros for platform specific tests if it's easy enough to add

dfrankland:
My bad, I wasn't paying attention

PR #662, opened 2025-06-11
[TRUNK-15076] Support unlimited rows for end output

cmillar-trunk:
Superconsole lets you emit rows, in which case they get outputted as a normal bit of console output, which lets us bypass the issues we were seeing with long rows and more rows than the tty has visible. Creates an EndOutput trait we can use for these in order to nicely display the end reports.

### DIFF
--- a/cli/src/upload_command.rs
+++ b/cli/src/upload_command.rs
@@ -14,12 +14,12 @@ use context::junit::validator::{
 use pluralizer::pluralize;
 use superconsole::{
     style::{style, Attribute, Color, Stylize},
-    Line, Span,
+    Line, Lines, Span,
 };
-use superconsole::{Component, Dimensions, DrawMode, Lines};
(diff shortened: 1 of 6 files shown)


gnalh (approved):
Nice!!

cli/src/upload_command.rs
```diff
@@ -418,14 +418,14 @@ async fn upload_bundle(
     Ok(())
 }
 
-impl Component for UploadRunResult {
-    fn draw_unchecked(&self, dimensions: Dimensions, mode: DrawMode) -> anyhow::Result<Lines> {
+impl EndOutput for UploadRunResult {
+    fn output(&self) -> anyhow::Result<Vec<Line>> {
```
gnalh, line 422: Since it'll probably merge in first, we'll need to update the new validate display to use this as well.

PR #671, opened 2025-06-12
Loading Indicators

cmillar-trunk:
Adds the ability to show loading indicators in the superconsole world. Also extracts out a bunch of the logic for rendering to a separate thread with channels to send messages so that we can handle messages from multiple threads (necessary because we have threads spun up to track progress on api calls).

progressMessage


cli/src/main.rs
```diff
@@ -108,65 +125,49 @@ fn main() -> anyhow::Result<()> {
...
-                    if let Err(e) = render_result {
-                        tracing::error!("Failed to render test display: {}", e);
-                    }
-                    guard.flush(None);
-                    std::process::exit(run_result.quarantine_context.exit_code)
+                    let result_ptr = Arc::new(run_result);
+                    send_message(
+                        DisplayMessage::Final(result_ptr.clone(), String::from("test display")),
+                        &render_sender,
+                    );
```
gnalh, line 149: Was this for debugging?

PR #672, opened 2025-06-13
[TRUNK-14788] Separate out test command output

cmillar-trunk:
Uses our outputters to format and display test command output more distinctly from the rest of the cli output.

Screenshot from 2025-06-17 12-49-17


cli/src/test_command.rs
```diff
@@ -44,16 +53,54 @@ pub struct TestRunResult {
...
+
+        for stdout_line in self.command_stdout.lines() {
+            output.push(Line::from_iter([Span::new_unstyled(stdout_line)?]));
+        }
+
+        output.extend(vec![
+            Line::default(),
+            Line::from_iter([Span::new_unstyled(String::from("📙 Stderr outputs"))?]),
+            Line::default(),
+        ]);
```
gnalh, line 80: We don't need the distinction of stderr vs stdout with different outputs. It often isn't very standardized (as an example: superconsole outputs to stderr). What I think would help make the test output more clear is having an identifying header (`Running command [command]`) with a dividing line between it an the "regular" output. What do you think?

cmillar-trunk: I think the identifying header makes sense, but I think we probably need to show some distinction between these two output types, just because the command crate forces it to be split - we can't merge them because there's no way to interleave them the way it'd happen with a terminal output (gonna admit, superconsole is the first I've seen that randomly uses stderr, and it was wild to me that it could engage in that misuse).

gnalh: >  I think we probably need to show some distinction between these two output types, just because the command crate forces it to be split 

If they need to be split then that's fine, I just don't think we necessarily need to be explicit about which one is stderr vs stdout in our output.

> gonna admit, superconsole is the first I've seen that randomly uses stderr, and it was wild to me that it could engage in that misuse

Unfortunately it's not super uncommon. I remember there being a few linters that use stderr to write their output when I was working on check.

cmillar-trunk: Sure, will remove the titles.

PR #676, opened 2025-06-16
Add TS Bindings For Getting Both Meta and Internal Bin

pv72895:


bundle/src/bundle_meta.rs
```diff
@@ -204,7 +204,7 @@ impl From<BundleMetaV0_7_7> for BundleMetaV0_7_6 {
     }
 }
 
-#[derive(Serialize, Deserialize, Debug, PartialEq, Eq)]
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone)]
```
cmillar-trunk, line 207: I don't see a call to `.clone()` here, is this needed?

context-js/src/lib.rs
```diff
@@ -113,10 +115,9 @@ pub fn repo_validate(bundle_repo: repo::BundleRepo) -> repo::validator::RepoVali
     repo::validator::validate(&bundle_repo)
 }
 
-#[wasm_bindgen()]
-pub async fn parse_meta_from_tarball(
+async fn get_buf_reader_from_stream(
     input: sys::ReadableStream,
-) -> Result<VersionedBundle, JsError> {
+) -> BufReaderAsync<Either<impl futures::io::AsyncRead, impl futures::io::AsyncRead>> {
```
cmillar-trunk, line 120: This feels like it warrants a separate type to disambiguate the two kinds of AsyncRead that might be returned - there's no way to tell from the signature what special cases apply to the Left vs Right case, which is very difficult to read or code against.

bundle/src/bundle_meta.rs
pv72895: Left over from previous attempts at doing this for python

cmillar-trunk: Ok, but VersionedBundleWithBindingsReport also doesn't seem to have clone called on it here, does it?

cmillar-trunk (approved):

bundle/src/bundle_meta.rs
pv72895: Resolved - it is because I'm using `getter_with_clone`

PR #682, opened 2025-06-17
Expose variant on bindings reports

pv72895:
As it says in the title - expose the variant in the context (`BindingsReport`)

### DIFF
--- a/context-js/tests/parse_compressed_bundle.test.ts
+++ b/context-js/tests/parse_compressed_bundle.test.ts
@@ -302,7 +302,7 @@ describe("context-js", () => {
 
     expect(result?.tests).toBe(13);
     expect(result?.test_suites).toHaveLength(2);
-    expect(result?.variant).toBeUndefined();
+    expect(result?.variant).toBe("");
 
     const contextRubySuite = result?.test_suites.find(
       ({ name }) => name === "context_ruby",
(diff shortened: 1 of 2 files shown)


proto/proto/test_context.proto
```diff
@@ -32,9 +32,10 @@ message TestCaseRun {
 }
 
 message UploaderMetadata {
-    string version = 1;
-    string origin = 2; // RSpec, jest, etc..
-    google.protobuf.Timestamp upload_time = 3;
+  string version = 1;
+  string origin = 2; // RSpec, jest, etc..
+  google.protobuf.Timestamp upload_time = 3;
+  optional string variant = 4;
```
gnalh, line 38: Everything in proto is optional, this isn't needed

gnalh:
You also need to update the ruby binding definitions

test_report/src/report.rs
```diff
@@ -486,7 +489,7 @@ impl From<MutTestReport> for String {
 #[cfg(feature = "ruby")]
 pub fn ruby_init(ruby: &magnus::Ruby) -> Result<(), magnus::Error> {
     let status = ruby.define_class("Status", ruby.class_object())?;
-    status.define_singleton_method("new", magnus::function!(Status::new, 1))?;
+    status.define_singleton_method("new", magnus::function!(Status::new, 2))?;
```
gnalh, line 492: This is the wrong one. It should be the one on line 495
```
test_report.define_singleton_method("new", magnus::function!(MutTestReport::new, 3))?;
```

gnalh: You also need to update the TestRepo invocation in `context-ruby/lib/trunk_spec_helper.rb`

PR #691, opened 2025-06-18
Add Code And Bindings For Generating Test Case CheckSums

pv72895:
## Context 

Test Case IDs act as a checksum for tests - they're V5 UUID generated using aspects of the test like its filepath and name. If nothing about the test changes (including the variant used when running it), the ID remains fixed

## What Does The PR Do

Currently, generating the ID / Checksum for test runs is done [in the ETL when it runs](https://github.com/trunk-io/trunk/blob/5117073037075cc5f51c2bf59fe5422c8360c8fd/trunk/py/metrics_data_pipeline/trunk_etl_common/id_schema.py#L30). This PR moves that functionality over to rust bindings so that both the ETL and other processes (like our root cause AI agent) can use it

## Testing

This PR adds rust tests and tests for the TS and Python bindings. Also verified this against a recently uploaded test in staging, specifically https://app.trunk-staging.io/trunk-staging-org/flaky-tests/test/3f507aef-e834-523b-a8ad-edaba6b137be?repo=trunk-io%2Ftrunk (see actual rust test in `id.rs`

### DIFF
--- a/context-js/src/lib.rs
+++ b/context-js/src/lib.rs
@@ -187,8 +187,8 @@ pub fn gen_info_id(
     name: Option<String>,
     info_id: Option<String>,
     variant: String,
-) -> Result<String, JsError> {
-    let info_id = gen_info_id_impl(
+) -> String {
+    gen_info_id_impl(
         &org_url_slug,
         &repo_full_name,
         file.as_deref(),
@@ -197,6 +197,5 @@ pub fn gen_info_id(
         name.as_deref(),
         info_id.as_deref(),
         &variant,
-    );
-    Ok(info_id)
+    )
 }
--- a/context-js/src/lib.rs
+++ b/context-js/src/lib.rs
@@ -6,7 +6,7 @@ use bundle::{
     parse_meta_from_tarball as parse_tarball_meta, FileSetTestRunnerReport, VersionedBundle,
     VersionedBundleWithBindingsReport,
 };
-use context::{env, junit, repo};
+use context::{env, junit, meta::id::gen_info_id as gen_info_id_impl, repo};
 use futures::{future::Either, io::BufReader as BufReaderAsync, stream::TryStreamExt};
 use js_sys::Uint8Array;
 use prost::Message;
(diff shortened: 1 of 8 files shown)


context-py/src/lib.rs
```diff
@@ -362,6 +361,42 @@ fn associate_codeowners_multithreaded_impl(
     Ok(results)
 }
 
+#[gen_stub_pyfunction]
+#[pyfunction]
+#[pyo3(signature = (
```
pv72895, line 366: This feels wrong to me? This came out of what looks like there being too many `Option<String>`s

context-js/src/lib.rs
```diff
@@ -175,3 +175,28 @@ pub async fn parse_internal_bin_and_meta_from_tarball(
...
+
+#[wasm_bindgen]
+// trunk-ignore(clippy/too_many_arguments)
+pub fn gen_info_id(
+    org_url_slug: String,
+    repo_full_name: String,
+    file: Option<String>,
+    classname: Option<String>,
+    parent_fact_path: Option<String>,
+    name: Option<String>,
```
pv72895, line 187: I switch from `&str` to `String` since I'm getting errors that wasm doesn't know how to create an equivalent for `&str` but it did for `String` - if there was something else I should have done please let me know

context-js/src/lib.rs
```diff
@@ -175,3 +175,28 @@ pub async fn parse_internal_bin_and_meta_from_tarball(
...
+    info_id: Option<String>,
+    variant: String,
```
TylerJang27, line 189: Is empty string variant expected here? Should this be optional?

TylerJang27 (commented):
Looks good, deferring to rust experts

context-js/src/lib.rs
pv72895: Just a note that I made it not optional here because it isn't optional in the ETL implementation - https://github.com/trunk-io/trunk/blob/5117073037075cc5f51c2bf59fe5422c8360c8fd/trunk/py/metrics_data_pipeline/trunk_etl_common/id_schema.py#L38

I'm fine either way (no reason to make this optional and then default to `""`) but figured there was a reason it was required in the ETL

cmillar-trunk: Empty strings are considered 'no variant', so it doesn't need to be optional.

context-js/src/lib.rs
```diff
@@ -175,3 +175,28 @@ pub async fn parse_internal_bin_and_meta_from_tarball(
...
+        &org_url_slug,
+        &repo_full_name,
+        file.as_deref(),
+        classname.as_deref(),
+        parent_fact_path.as_deref(),
+        name.as_deref(),
+        info_id.as_deref(),
+        &variant,
+    );
+    Ok(info_id)
```
cmillar-trunk, line 201: Nit: it looks like this is always `Ok`, can we skip the Result, or is it needed for where this is used? (Ditto for the context-py version of this).

cmillar-trunk (approved):

PR #696, opened 2025-06-19
Lossify test command output

cmillar-trunk:
Since the test command is untrusted, it can contain text that looks like html, which superconsole chokes on. Switching to the lossy variant of the functions that output those to prevent this. Also substituting carriage returns and tab characters, which superconsole also cannot handle.

### DIFF
--- a/cli/src/test_command.rs
+++ b/cli/src/test_command.rs
@@ -57,6 +57,18 @@ pub struct TestRunResult {
     pub command_stderr: String,
 }
 
+fn lines_of(output: String) -> impl Iterator<Item = String> {
+    output
+        .lines()
+        .flat_map(|lf_line| {
+            lf_line
+                .split('\r')
+                .map(|crlf_line| crlf_line.replace('\t', "    "))
+        })
+        .collect::<Vec<String>>()
+        .into_iter()
+}
+
 impl EndOutput for TestRunResult {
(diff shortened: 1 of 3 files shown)


cli/src/upload_command.rs
```diff
@@ -450,13 +450,13 @@ impl EndOutput for UploadRunResult {
...
             return Ok(output);
         }
-        if !self.validations.validations.is_empty() {
-            output.extend(
-                self.validations
-                    .output_with_report_limits(&self.validation_report)?,
-            );
-            output.push(Line::default());
-        }
+        //if !self.validations.validations.is_empty() {
```
gnalh, line 453: Did you mean to leave this commented?

cli/src/test_command.rs
```diff
@@ -57,6 +57,18 @@ pub struct TestRunResult {
     pub command_stderr: String,
 }
 
+fn lines_of(output: String) -> impl Iterator<Item = String> {
+    output
+        .lines()
+        .flat_map(|lf_line| {
+            lf_line
+                .split('\r')
+                .map(|crlf_line| crlf_line.replace('\t', "    "))
```
gnalh, line 66: Are we missing new line here too?

gnalh (commented):
Can we add a cli test for this as well? I think having a command that echos with special space characters would work.

gnalh (approved):

PR #699, opened 2025-06-25
Swap stdout and stderr

cmillar-trunk:
Bazel outputs its final summary to stdout, but progress logs to stderr. With stdout>stderr, we end up logging the summary above the progress. Since bazel shows up in so many places, we'll swap the order so it's more consistent with expectations.


gnalh (approved):
LGTM, make sure to update the summary + title to match the new intent.

PR #715, opened 2025-07-09
Collect bazel target information

gnalh:


proto/proto/test_context.proto
```diff
@@ -38,7 +38,29 @@ message UploaderMetadata {
...
 
+message BazelBuildInformation {
+  string label = 1; // test label for the execution 
+  TestBuildResult result = 2; // result of the build
+}
+
 message TestResult {
   repeated TestCaseRun test_case_runs = 1;
+  // use the uplaoder metadata in test report instead
+  UploaderMetadata uploader_metadata = 2 [deprecated = true];;
```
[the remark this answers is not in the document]
gnalh: We'll get uploads with the old version on older versions of the CLI and need to support grabbing that information from both. Once we no longer have the v1 version of these used anymore, we can remove this.

cmillar-trunk (approved):

gnalh:
Really need to get code coverage working on context-* and cli. The current report is misleading.

cli/src/context.rs
```diff
@@ -211,6 +214,112 @@ pub fn gather_post_test_context<U: AsRef<Path>>(
...
+            test_case_runs,
+            // trunk-ignore(clippy/deprecated)
+            uploader_metadata: Some(UploaderMetadata {
+                variant: variant.clone().unwrap_or_default(),
+                ..Default::default()
+            }),
+            test_build_information: Some(BazelBuildInformationEnum(BazelBuildInformation {
+                label,
+                result: TestBuildResult::Success as i32,
+            })),
```
TylerJang27, line 293: ✅ I think this is what we need. Per run what the target's final result was. Am I reading this right?

context/src/bazel_bep/parser.rs
```diff
@@ -188,4 +188,55 @@ mod tests {
...
+        assert_eq!(first_label, "//trunk/hello_world/cc:hello_test");
+        let first_label_files = uncached_labels.get(first_label).unwrap();
+        assert_eq!(first_label_files.len(), 1);
+
+        let first_label_file = &first_label_files[0];
+        assert_eq!(first_label_file.junit_path, "/tmp/hello_test/test.xml");
+        assert!(first_label_file.test_runner_report.is_some());
+        if let Some(report) = &first_label_file.test_runner_report {
+            assert_eq!(report.status, TestRunnerReportStatus::Flaky);
+        }
```
TylerJang27, line 215: 🙌

TylerJang27 (approved):
LGTM on first pass. I still need to take it for a spin for our use case in AI land. Will do on Tues, feel free to land beforehand

proto/proto/test_context.proto
```diff
@@ -38,7 +38,29 @@ message UploaderMetadata {
   string variant = 4;
 }
 
+message BazelBuildInformation {
+  string label = 1; // test label for the execution 
+  TestBuildResult result = 2; // result of the build
+}
```
TylerJang27, line 44: Can we get this information per test case/suite? Having it top-level isn't enough for us to dedupe the bazel reruns issue we've been having

PR #718, opened 2025-07-11
[TRUNK-15323] Recurse on paths that look file-ish

cmillar-trunk:
If a path ends without a slash, github will recurse into it for codeowners, so we add a fallback path to recurse in any directory that matches, as well as matching on any files that match.

### DIFF
--- a/codeowners/src/github.rs
+++ b/codeowners/src/github.rs
@@ -197,7 +197,7 @@ impl PatternWithFallback {
         // Matches anything that ends with neither a slash nor a period nor an asterisk,
         // see test pattern_with_fallback for cases
         static FALLBACK_NEEDED_REGEX: Lazy<Regex> =
-            Lazy::new(|| Regex::new(r"^\/?([a-zA-Z0-9_*\-\.]*\/)*(\.?[a-zA-Z0-9_\-]+)$").unwrap());
+            Lazy::new(|| Regex::new(r"^\/?(.*\/)*((\.?)[^\/\.\*]+)$").unwrap());
 
         let base_pattern = Pattern::new(base).map_err(anyhow::Error::msg)?;
         let mut fallback_pattern = None;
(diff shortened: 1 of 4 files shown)


gnalh (approved):
This looks reasonable to me and thoroughly tests the various cases we can hit. Any concerns with this approach @dfrankland ?

dfrankland:
I'm going to check this branch out and see if there's an edge case we're missing

codeowners/src/github.rs
```diff
@@ -187,7 +185,54 @@ impl BindingsGitHubOwners {
...
+pub struct PatternWithFallback {
+    base: Pattern,
+    fallback: Option<Pattern>,
+}
+
+impl PatternWithFallback {
+    pub fn new(base: &str) -> anyhow::Result<Self> {
+        // Matches anything that ends with neither a slash nor a period nor an asterisk,
+        // see test pattern_with_fallback for cases
+        let fallback_needed_regex = Regex::new(r"^\/?([a-zA-Z0-9_*\-]*\/)*([a-zA-Z0-9_\-]+)$")?;
```
dfrankland, line 198: This regex doesn't work for hidden paths which is something that is used by customers. For example this fails:

```rs
assert_eq!(
    PatternWithFallback::new("/.buildkite").unwrap(),
    PatternWithFallback {
        base: Pattern::new("/.buildkite").unwrap(),
        fallback: Some(Pattern::new("/.buildkite/**").unwrap()),
    },
);
```

Additionally, this won't support any paths with a period in it. Although this might be an edge-case we don't support, I think we should document that we don't support it yet. We can do that through a test or otherwise. For example this fails:

```rs
assert_eq!(
    PatternWithFallback::new("/top.level/second-level").unwrap(),
    PatternWithFallback {
        base: Pattern::new("/top.level/second-level").unwrap(),
        fallback: Some(Pattern::new("/top.level/second-level/**").unwrap()),
    },
);
```

dfrankland, line 198: We should use `lazy_static` for this regex to avoid recompiling it on every invocation. Docs:
https://docs.rs/regex/latest/regex/#avoid-re-compiling-regexes-especially-in-a-loop

cmillar-trunk: Nice catch. I added logic to support those paths and that extraction.

dfrankland (approved):

codeowners/src/github.rs
```diff
@@ -187,7 +186,55 @@ impl BindingsGitHubOwners {
...
+    base: Pattern,
+    fallback: Option<Pattern>,
+}
+
+impl PatternWithFallback {
+    pub fn new(base: &str) -> anyhow::Result<Self> {
+        // Matches anything that ends with neither a slash nor a period nor an asterisk,
+        // see test pattern_with_fallback for cases
+        static FALLBACK_NEEDED_REGEX: Lazy<Regex> =
+            Lazy::new(|| Regex::new(r"^\/?(.*\/)*((\.?)[^\/\.\*]+)$").unwrap());
```
dfrankland, line 200: This looks fine. In my humble opinion, I'd simplify this to just looking for `/\*{0,2}$` (look for a literal `/` character plus zero to two literal `*` characters at the end of the string):
https://regex101.com/r/gFPdbL/1

Then I would check for `if !FALLBACK_NEEDED_REGEX.is_match(base) { ... }` below.

That's up to you if you like that or not.

dfrankland (approved):

codeowners/src/github.rs
cmillar-trunk: I like the simplicity, but I'd like to avoid trying to recurse into files with extensions, which requires a more complex pattern.

PR #724, opened 2025-07-21
add codeowners debug cli

dfrankland:
This helped me figure out that we are associating GitLab files differently than GitHub files and that because we default to GitLab we get bad results.

Example:
**`CODEOWNERS`**
```
my_top_level/my_second_level @dylan
```

**Attmpting to match using GitLab**

```bash
cargo run -- --codeowners-type gitlab --codeowners-path ./CODEOWNERS --test-case-path my_top_level/my_second_level/src/file.js
```

```
No owners found for my_top_level/my_second_level/src/file.js
```

**Attmpting to match using GitHub**

```bash
cargo run -- --codeowners-type github --codeowners-path ./CODEOWNERS --test-case-path my_top_level/my_second_level/src/file.js
```

```
Owners found for my_top_level/my_second_level/src/file.js:
@dylan
```

Places in code we try to parse GitLab, then as GitHub:
* https://github.com/trunk-io/analytics-cli/blob/24666bbbbe53044d537be447b92f0a5b87c7b640/codeowners/src/codeowners.rs#L83-L91
* https://github.com/trunk-io/analytics-cli/blob/24666bbbbe53044d537be447b92f0a5b87c7b640/codeowners/src/codeowners.rs#L112-L116
*

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -899,6 +899,7 @@ version = "0.1.3"
 dependencies = [
  "anyhow",
  "assert_matches",
+ "clap",
  "const_format",
  "constants",
  "fancy-regex",
--- a/codeowners/Cargo.toml
+++ b/codeowners/Cargo.toml
@@ -3,12 +3,17 @@ name = "codeowners"
 edition = "2018"
 version = "0.1.3"
 
+[[bin]]
+name = "check-codeowners"
+path = "src/main.rs"
+
 [dev-dependencies]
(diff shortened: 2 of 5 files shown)


gnalh (approved):
This is super useful to have. Though it doesn't exactly emulate what the ETL uses since it isn't calling the same top level function.

dfrankland:
Totes, we can update to improve as we go along

PR #733, opened 2025-07-31
Allow users to specify GitHub job url manually

gnalh:
Update the ETL bindings to support using the `JOB_URL` environment variable.

### DIFF
--- a/context/src/env/parser.rs
+++ b/context/src/env/parser.rs
@@ -303,18 +303,22 @@ impl<'a> CIInfoParser<'a> {
 
         self.ci_info.actor = self.get_env_var("GITHUB_ACTOR");
         self.ci_info.title = self.get_env_var("PR_TITLE");
-        if let Some(job_url) = self.get_env_var("JOB_URL") {
-            self.ci_info.job_url = Some(job_url);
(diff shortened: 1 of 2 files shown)


context/src/env/parser.rs
```diff
@@ -303,7 +303,9 @@ impl<'a> CIInfoParser<'a> {
 
         self.ci_info.actor = self.get_env_var("GITHUB_ACTOR");
         self.ci_info.title = self.get_env_var("PR_TITLE");
-        if let (Some(repo_name), Some(run_id)) = (
+        if let Some(job_url) = self.get_env_var("JOB_URL") {
+            self.ci_info.job_url = Some(job_url);
+        } else if let (Some(repo_name), Some(run_id)) = (
```
dfrankland, line 308: nit, totally optional: I think it might be better to use some mapping such that the `if` and `else` clauses don't diverge in setting 1 property on the struct

dfrankland (approved):

context/src/env/parser.rs
gnalh: Updated to use a match statement

PR #739, opened 2025-08-07
[2] Update the context to support grouping by label

gnalh:


context/src/bazel_bep/common.rs
```diff
@@ -92,20 +142,31 @@ impl BepParseResult {
...
-                                    xml_files,
-                                    test_runner_report: None,
-                                });
+                                if let Some(existing_result) =
+                                    acc.test_results.iter_mut().find(|r| r.label == id.label)
+                                {
+                                    existing_result.xml_files.extend(xml_files);
+                                    let new_status =
+                                        map_i32_test_status_to_bep_test_status(test_result.status);
+                                    existing_result.build_status = Some(new_status);
```
gnalh, line 158: Updates the build_status to represent the most recent status

context/src/bazel_bep/common.rs
```diff
@@ -12,12 +12,60 @@ use crate::junit::junit_path::{
...
+    if test_status == TestStatus::RemoteFailure as i32 {
+        return BepTestStatus::RemoteFailure;
+    }
+    if test_status == TestStatus::FailedToBuild as i32 {
+        return BepTestStatus::FailedToBuild;
+    }
+    if test_status == TestStatus::ToolHaltedBeforeTesting as i32 {
+        return BepTestStatus::ToolHaltedBeforeTesting;
+    }
+    BepTestStatus::NoStatus
```
cmillar-trunk, line 40: Is this an expected case, or would getting a number that doesn't line up with the proto enum be an error? I'm wondering if this should be Result or Option.

context/src/junit/bindings.rs
```diff
@@ -29,6 +29,26 @@ pub struct BindingsParseResult {
...
+#[derive(Clone, Debug, Copy, PartialEq, Eq)]
+pub enum BindingsTestBuildResult {
+    Unspecified,
+    Success,
+    Failure,
+    Skipped,
+    Flaky,
+}
+
+// Ideally this would be an enum, but enums are not directly supportted by wasm conversions, so we have to manually map out the options. See: https://github.com/rustwasm/wasm-bindgen/issues/2407
```
cmillar-trunk, line 43: Does this mean parameterised enums (or adts or product types or whatever the rust-space terminology is), since there's an enum used directly above this?

context/src/junit/bindings.rs
```diff
@@ -43,6 +63,23 @@ pub struct BindingsReport {
...
+    } else if result == TestBuildResult::Success as i32 {
+        BindingsTestBuildResult::Success
+    } else if result == TestBuildResult::Failure as i32 {
+        BindingsTestBuildResult::Failure
+    } else if result == TestBuildResult::Skipped as i32 {
+        BindingsTestBuildResult::Skipped
+    } else if result == TestBuildResult::Flaky as i32 {
+        BindingsTestBuildResult::Flaky
+    } else {
+        BindingsTestBuildResult::Unspecified
```
cmillar-trunk, line 81: As above, I'm wondering if this should be Option or Result to capture i32's that don't map onto the proto enum.

context/src/bazel_bep/common.rs
gnalh: It would be an error. It should never happen but prost converts all of the values into integers, even if they were originally enums.

context/src/junit/bindings.rs
gnalh: It's for complex wrapped enums, not simple ones. Ideally the build information would be an enum, with bazel being an option.

cmillar-trunk (approved):

TylerJang27 (approved):

PR #743, opened 2025-08-07
[4] Update test report to use new proto format

gnalh:


TylerJang27 (approved):
I tested bindings from this branch btw and that's what worked, with internal.bin generated from [5]

PR #746, opened 2025-08-07
[5] Fix parsing the number of tests 

gnalh:
We have been double counting in the TUI. We need to only count the unique tests because retries don't impact results (only the final attempt counts).


cmillar-trunk:
Do we also need to change how we're counting success/failure/skipped counts for this to also avoid duplicates? I'm worried about making sure that our reported success + failure + skipped  should equal the total, or else it looks like the cli is obviously wrong.

PR #764, opened 2025-08-28
[TRUNK-16209] Limit quarantine disabling to only fetching context

cmillar-trunk:
Moves the check for disabling quarantining to only apply when we're fetching quarantine state, otherwise building up context. This means that failing tests will be compared against a default quarantining state, ie no quarantined tests, allowing us to report on them and fail them.

### DIFF
--- a/cli-tests/src/upload.rs
+++ b/cli-tests/src/upload.rs
@@ -1765,10 +1765,12 @@ async fn reports_failing_tests_but_succeeds_when_quarantine_disabled() {
         .disable_quarantining(true)
         .command();
 
+    // Our junit library generates random amounts of test failures (but at least 1),
+    // so we check that we don't have 0 failures.
     let assert = command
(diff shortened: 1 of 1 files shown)


cli-tests/src/upload.rs
gnalh, line 1: This is a pretty big change in terms of functionality. Correct me if I'm wrong, but anyone not using quarantining will suddenly see their uploads fail if we have failures in the bundle.

gnalh: Yeah we definitely don't want to break CI for customers. Right now the expectation for users is: "if quarantining is enabled in the repo then we should fail on un-quarantined test failures. If quarantining is disabled in the repo then we should only fail on upload failures"

cmillar-trunk: Gotcha, I'll change to that.

PR #771, opened 2025-09-03
[TRUNK-16320] Switch to sanitising user values

cmillar-trunk:
Superconsole has a bunch of things it crashes when attempting to display, but has _lossy variants of functions which will sanitise any text given to them. Switching to those for anything that can contain user-provided values.


cmillar-trunk:
@gnalh Good point, will add.

PR #780, opened 2025-09-26
[TRUNK-16582] Fix timezone offset handling when converting JUnit timestamp to ChronoDateTime

blx:


context/src/junit/date_parser.rs
```diff
@@ -45,7 +45,7 @@ impl JunitDateParser {
             .map(|dt| {
                 self.date_type = Some(DateType::DateTime);
                 TimestampAndOffset {
-                    timestamp_secs_micros: Some((dt.timestamp(), dt.time.microsecond)),
+                    timestamp_secs_micros: Some((dt.timestamp_tz(), dt.time.microsecond)),
```
blx, line 48:
```rs
    /// Unix timestamp (seconds since epoch, 1970-01-01T00:00:00) omitting timezone offset
    /// (or equivalently comparing to 1970-01-01T00:00:00 in the same timezone as self)
    pub fn timestamp(&self) -> i64 { ... }

    /// Unix timestamp assuming epoch is in zulu timezone (1970-01-01T00:00:00Z) and accounting for
    /// timezone offset.
    ///
    pub fn timestamp_tz(&self) -> i64 { ... }
```

cmillar-trunk (approved):

PR #1086, opened 2026-05-05
(Feat): Support X-Trunk-Public-Repo-Id auth for fork PRs

mmatheson:
## Summary

Adds support for the new `X-Trunk-Public-Repo-Id` header so fork-PR runs that lack `TRUNK_API_TOKEN` can still upload bundles and fetch the quarantine list. See the integration guide for the server-side contract.

- New `--public-repo-id` flag / `TRUNK_PUBLIC_REPO_ID` env. Token-first: when both are set, the token wins. When neither is set, `upload` exits with a clear message naming both env vars.
- Auth is sent on the two affected endpoints (`createBundleUpload`, `getQuarantineConfig`); telemetry is silently skipped on the public-id path because the telemetry endpoint does not accept the new header. S3 PUT is unchanged.
- Logs `Using X-Trunk-Public-Repo-Id auth (TRUNK_API_TOKEN not set; assuming fork PR)` once at info level when the public-id path is taken; the id value itself is never logged.
- Same fallback applied to the rspec/`test_report` flow (`is_quarantined` and `publish`).

## Implementation

- `api/src/auth.rs` (new): `TrunkApiCredential::{Token, PublicRepoId}` plus `resolve(token, public_repo_id)` with token-first ordering. Empty / whitespace strings count as absent.
- `api/src/client.rs`: `ApiClient::new` now takes a `TrunkApiCredential`; the trunk-API client carries either `x-api-token` or `x-trunk-public-repo-id`. Telemetry client is `Option<Client>` and `telemetry_upload_metrics` short-circuits with `Ok(())` when no token is available.
- `cli/src/upload_command.rs`: `--token` no longer required by clap; `run_upload` resolves the credential and `bail!`s with a clear message when neither is set.
- `test_report/src/report.rs`: both call sites resolve via `TrunkApiCredential` and forward `public_repo_id` into `UploadArgs`.

The integration guide also names a renamed `/v1/flaky-tests/list-quarantined-tests` path; the server has wired the new auth into the existing `/v1/metrics/getQuarantineConfig` endpoint as well, so no endpoint migration is needed in this PR. GitHub Action (`action.yml`) changes live in `trunk-io/analytics-uploader` and are out of scope.

## Test plan

- [x] `cargo build --workspace` — clean
- [x] `cargo test -p api` — 20/20 (incl. 7 new `auth::*` tests covering token-only, public-id-only, both → token wins, empty/whitespace fall-through, neither → None)
- [x] `cargo test -p test_report` — 6/6
- [x] `cargo test --workspace` — all pass except `upload_bundle_using_xcresult`, which fails on this developer machine before any HTTP call ("failed to parse xcresulttool version" — local Xcode env issue, unrelated to this change)
- [ ] Staging integration test: `RepoForkPrUploadsConfig.enabled = true` → CLI with only `TRUNK_PUBLIC_REPO_ID` set → presigned URL returned
- [ ] Staging integration test: toggle to `enabled = false` → 401 surfaced cleanly
- [ ] Real fork PR end to end (verify bundle appears in dashboard)


dfrankland:
superseded by #1101

PR #1091, opened 2026-05-12
(Fix): Strip ansi colors from rspec exceptions

TylerJang27:
The recent changes to use the native rspec `RSpec::Core::Formatters::ExceptionPresenter` were an improvement, but we got a report that it was including ANSI color codes. I was able to repro iff `RSpec.configuration.color_mode = :on` is set.

Additionally,
> We had to strip ANSI codes at the final serialization boundary because PlainColorizer only stops ExceptionPresenter from adding color in its wrap(...) step. RSpec can still inject ANSI earlier when it syntax-highlights the Failure/Error: source snippet, and that path is controlled by global RSpec color settings plus CodeRay, not by our custom colorizer. Since those escapes can already be embedded by the time we format test_output.message and test_output.text, the only reliable fix is to sanitize the final stored strings.

i.e. stripping the codes this way is the safest approach. I was able to repro and verify the fix. Meta note that we're getting to the point where we need a bit more code organization for the rspec gem. I will undertake that when the next fix comes up

### DIFF
--- a/.github/actions/test_ruby_gem_uploads/Gemfile
+++ b/.github/actions/test_ruby_gem_uploads/Gemfile
@@ -3,6 +3,7 @@
 source 'https://rubygems.org'
 
 gem 'rspec'
+gem 'coderay'
 gem 'knapsack_pro'
 gem 'rake'
 gem 'rake-compiler'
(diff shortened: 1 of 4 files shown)


TylerJang27:
@claude review

dfrankland (approved):
woof

PR #1094, opened 2026-05-18
Fix xcresult snapshot test file attribution

dfrankland:
## Summary
- Prefer structured failure summary file paths before falling back to call-stack inspection.
- Preserve the existing legacy `documentLocationInCreatingWorkspace` fallback path.
- Add a Swift SnapshotTesting xcresult fixture plus parameterized unit and integration coverage.

## Rationale
Swift SnapshotTesting trait failures can include framework files later in the call stack, such as `SnapshotsTestTrait.swift`. The previous experimental failure-summary path selected the last Swift or Obj-C call-stack frame, so the reported JUnit `file` could point at SnapshotTesting internals instead of the test source that asserted the snapshot.

The xcresult failure summary already contains better structured attribution fields. Although the legacy schema calls the preferred field `fileName`, the inspected xcresult objects store a full absolute source path there, not just a basename. In the SnapshotTesting fixture, `fileName` and `sourceCodeContext.location.filePath` both point to `/tmp/xcresult-snapshot-repro.sPh62f/Tests/SnapshotReproTests/SnapshotReproTests.swift`; the call stack also includes full paths, but later frames point at SnapshotTesting internals.

This change therefore prefers `fileName`, then `sourceCodeContext.location.filePath`, and only falls back to the old call-stack heuristic when those structured fields are absent. The non-experimental legacy path remains available and unchanged in behavior.

## Testing
- `cargo test -p xcresult`
- Temporarily reverted the source change and confirmed `test_swift_snapshot_testing_trait_failure_uses_assertion_file::case_1_experimental_failure_summary` fails by pointing at `SnapshotsTestTrait.swift`; restored the fix and confirmed it passes.


xcresult/src/xcresult_legacy.rs
```diff
@@ -34,44 +34,112 @@ impl XCResultTestLegacy {
...
+                            .map(|ext| ext == "swift" || ext == "m")
+                            .unwrap_or(false)
+                    })
+                    // use the last valid swift / obj-c file-path in the stack
+                    .last()
+            })
+    }
+
+    fn normalize_file_path(file_path: Option<&String>) -> Option<String> {
+        file_path.map(|file_path| file_path.replace(' ', "%20"))
```
TylerJang27, line 97: Is this necessary?

TylerJang27 (approved):

xcresult/src/xcresult_legacy.rs
dfrankland: Unfortunately, yes. We didn't correct for this before when parsing XCResult `file://` URLS, so for test identity backwards compat, we need to continue percent encoding spaces to ensure we don't create new test cases.

PR #1095, opened 2026-05-19
(Chore): Include trunk_envs in meta.json

TylerJang27:
For our ease of debugging customer issues.
Screenshot 2026-05-19 at 3 14 06 PM


cli/src/context.rs
```diff
@@ -67,12 +67,25 @@ lazy_static! {
 // This function is used to gather debug properties for the bundle meta.
 // It will trigger EXC_BAD_ACCESS on arm64-darwin builds when compiled under cdylib
 pub fn gather_debug_props(args: Vec<String>, token: String) -> BundleMetaDebugProps {
+    let trunk_envs: HashMap<String, String> = env::vars()
+        .filter(|(key, _)| key.starts_with("TRUNK_"))
```
dfrankland: I would err on the side of collecting specific env vars so we don't accidentally capture secrets our users are storing with a `TRUNK_` prefix

cli/src/context.rs
```diff
@@ -67,12 +67,25 @@ lazy_static! {
...
+        .map(|(key, value)| {
+            let masked_value = if key == TRUNK_API_TOKEN_ENV {
+                "***".to_string()
```
dfrankland: No reason to capture this if we just replace it with `***`

dfrankland (approved):
v cool

PR #1096, opened 2026-05-27
(Chore): Include message when quarantine lookup fails

TylerJang27:
Leans on https://github.com/trunk-io/analytics-cli/pull/966 and improves the audit story when the call to get quarantined tests fails in RSpec. We already do this in the analytics-cli (`We were unable to determine the quarantine status for tests...`), but it was missing on the RSpec side.

Screenshot 2026-05-27 at 2 27 59 AM
Screenshot 2026-05-27 at 2 25 24 AM


max-trunk (approved):
Looks good 👍 

A unit test in `tests/report.rs` would be great too

PR #1099, opened 2026-05-28
(Chore): Reorganize report tests

TylerJang27:
Base for https://github.com/trunk-io/analytics-cli/pull/1100

Some reorganization of domains for the report testing (used by RSpec plugin)

### DIFF
--- a/test_report/tests/common/env.rs
+++ b/test_report/tests/common/env.rs
@@ -1,0 +1,46 @@
+use std::env;
+
+use constants::{
+    TRUNK_ALLOW_EMPTY_TEST_RESULTS_ENV, TRUNK_API_CLIENT_RETRY_COUNT_ENV, TRUNK_API_TOKEN_ENV,
+    TRUNK_CODEOWNERS_PATH_ENV, TRUNK_DISABLE_QUARANTINING_ENV, TRUNK_DRY_RUN_ENV,
+    TRUNK_ORG_URL_SLUG_ENV, TRUNK_PR_NUMBER_ENV, TRUNK_PUBLIC_API_ADDRESS_ENV,
+    TRUNK_QUARANTINED_TESTS_DISK_CACHE_TTL_SECS_ENV, TRUNK_REPO_HEAD_AUTHOR_NAME_ENV,
+    TRUNK_REPO_HEAD_BRANCH_ENV, TRUNK_REPO_HEAD_COMMIT_EPOCH_ENV, TRUNK_REPO_HEAD_SHA_ENV,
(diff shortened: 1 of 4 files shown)


test_report/tests/common/env.rs
```diff
@@ -0,0 +1,46 @@
...
+        env::remove_var(TRUNK_DRY_RUN_ENV);
+        env::remove_var(TRUNK_CODEOWNERS_PATH_ENV);
+        env::remove_var("CI");
+        env::remove_var("GITHUB_JOB");
+        env::remove_var(TRUNK_QUARANTINED_TESTS_DISK_CACHE_TTL_SECS_ENV);
+        env::remove_var(TRUNK_API_CLIENT_RETRY_COUNT_ENV);
+    }
+}
+
+#[allow(dead_code)]
```
TylerJang27, line 38: I realize this is an antipattern in rust-world, but it matches the pattern of the rest of this repo, and I didn't want to undertake crate explosion right now. Having these common utils are necessary for the other "integration" tests in this crate.

test_report/tests/publish.rs
TylerJang27, line 1: Undertook this split mainly to keep the follow-up PR sane. But also just to make it clear whether we're testing the upload flow or the quarantine flow for rspec

TylerJang27: These tests are ported directly with no meaningful changes, just renames and commonizing a couple small pieces

dfrankland (approved):

test_report/tests/common/env.rs
acatxnamedvirtue: what about this is an anti-pattern? new to rust

acatxnamedvirtue (approved):

test_report/tests/common/env.rs
dfrankland: Ideally you don't define re-usable functions for tests within the `tests` directory. It forces one to mark the functions as `dead_code` because code (the test functions themselves) within `tests` files only materializes via compiler options.

TylerJang27: The compiler only executes this function inside of tests, which are excluded from the normal compilation cycle, so it thinks this is dead code. The better pattern afaik is to setup separate, flat crates for test_report, test_report_tests, test_report_test_utils, that way the functions are used either directly where they're defined or exposed as public in the crate (and therefore not unused). That's what I tried to do for the rust integration tests in trunk2. But doing so right now in analytics-cli would require a substantial refactor, so I'm just using the macro here to bypass the warning

PR #1100, opened 2026-05-28
(Chore): Better telemetry for quarantine query

TylerJang27:
Stacked on https://github.com/trunk-io/analytics-cli/pull/1099. Adds missing observability for the result of querying quarantined tests. 

Inputs to quarantine query status:
Screenshot 2026-05-28 at 9 19 50 AM

### DIFF
--- a/cli/src/context_quarantine.rs
+++ b/cli/src/context_quarantine.rs
@@ -21,6 +21,7 @@ use context::{
 };
 use pluralizer::pluralize;
 use prost::Message;
+use proto::upload_metrics::trunk::QuarantineQueryResult;
 
 #[derive(Debug)]
 pub enum QuarantineFetchStatus {
@@ -68,6 +69,21 @@ impl QuarantineContext {
             org_url_slug: String::default(),
             fetch_status: QuarantineFetchStatus::FetchFailed(error),
         }
+    }
+}
+
+pub fn quarantine_query_result(
+    disable_quarantining: bool,
+    ctx: &QuarantineContext,
+) -> proto::upload_metrics::trunk::QuarantineQueryResult {
+    if disable_quarantining || ctx.quarantine_status.quarantining_disabled_for_repo {
+        return QuarantineQueryResult::Disabled;
+    }
+
+    match &ctx.fetch_status {
+        QuarantineFetchStatus::FetchFailed(_) => QuarantineQueryResult::Failure,
+        QuarantineFetchStatus::FetchSkipped => QuarantineQueryResult::Skipped,
+        QuarantineFetchStatus::FetchSucceeded => QuarantineQueryResult::Success,
     }
 }
 
@@ -334,7 +350,10 @@ pub async fn gather_quarantine_context(
         tracing::info!("Quarantining is not enabled, not quarantining any tests");
         return Ok(QuarantineContext {
             exit_code,
-            quarantine_status: QuarantineBulkTestStatus::default(),
+            quarantine_status: QuarantineBulkTestStatus {
+                quarantining_disabled_for_repo: true,
+                ..Default::default()
+            },
             failures: failed_tests_extractor.failed_tests().to_vec(),
             repo: request.repo.clone(),
             org_url_slug: request.org_url_slug.clone(),
@@ -625,6 +644,53 @@ mod tests {
         multi_failures.sort_by(|a, b| a.name.cmp(&b.name));
         assert_eq!(multi_failures.len(), 1);
         assert_eq!(multi_failures[0].name, "Hello");
+    }
+
+    #[test]
+    fn test_quarantine_query_result_mapper() {
+        use proto::upload_metrics::trunk::QuarantineQueryResult;
+
+        let success_ctx = QuarantineContext {
+            exit_code: 0,
+            quarantine_status: QuarantineBulkTestStatus::default(),
+            failures: vec![],
+            repo: RepoUrlParts::default(),
+            org_url_slug: String::new(),
+            fetch_status: QuarantineFetchStatus::FetchSucceeded,
+        };
+        assert_eq!(
+            quarantine_query_result(false, &success_ctx),
+            QuarantineQueryResult::Success
+        );
+        assert_eq!(
+            quarantine_query_result(true, &success_ctx),
+            QuarantineQueryResult::Disabled
+        );
+
+        let skipped_ctx = QuarantineContext {
+            fetch_status: QuarantineFetchStatus::FetchSkipped,
+            ..success_ctx
+        };
+        assert_eq!(
+            quarantine_query_result(false, &skipped_ctx),
+            QuarantineQueryResult::Skipped
+        );
+
+        let repo_disabled_ctx = QuarantineContext {
+            quarantine_status: QuarantineBulkTestStatus {
+                quarantining_disabled_for_repo: true,
+                ..Default::default()
+            },
+            fetch_status: QuarantineFetchStatus::FetchSucceeded,
+            exit_code: 0,
+            failures: vec![],
+            repo: RepoUrlParts::default(),
+            org_url_slug: String::new(),
+        };
+        assert_eq!(
+            quarantine_query_result(false, &repo_disabled_ctx),
+            QuarantineQueryResult::Disabled
+        );
     }
 
     #[test]
--- a/cli/src/upload_command.rs
+++ b/cli/src/upload_command.rs
@@ -18,7 +18,7 @@ use superconsole::{
 };
 use tempfile::TempDir;
 
-use crate::context_quarantine::QuarantineContext;
+use crate::context_quarantine::{QuarantineContext, quarantine_query_result};
 use crate::validate_command::JunitReportValidations;
 use crate::{
     context::{
@@ -367,6 +367,7 @@ pub async fn run_upload(
     pre_test_context: Option<PreTestContext>,
     test_run_result: Option<TestRunResult>,
     render_sender: Option<Sender<DisplayMessage>>,
+    quarantine_query_result_override: Option<proto::upload_metrics::trunk::QuarantineQueryResult>,
 ) -> anyhow::Result<UploadRunResult> {
     // grab the exec start if provided (`test` subcommand) or use the current time
     let cli_started_at = if let Some(test_run_result) = test_run_result.as_ref() {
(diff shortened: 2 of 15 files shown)


cli/src/context_quarantine.rs
```diff
@@ -627,6 +646,53 @@ mod tests {
         assert_eq!(multi_failures[0].name, "Hello");
     }
 
+    #[test]
+    fn test_quarantine_query_result_mapper() {
```
TylerJang27, line 650: unit test for this

cli/src/upload_command.rs
```diff
@@ -360,6 +360,7 @@ pub async fn run_upload(
     pre_test_context: Option<PreTestContext>,
     test_run_result: Option<TestRunResult>,
     render_sender: Option<Sender<DisplayMessage>>,
+    quarantine_query_result_override: Option<proto::upload_metrics::trunk::QuarantineQueryResult>,
```
TylerJang27: RSpec calls the quarantine config at a separate time than the upload flow. There's a gate above this where we conditionally call GetQuarantineConfig depending on if there are JUnits (which there aren't in RSpec). That's a smell, but I didn't want to undertake that refactor right now, and it's not a meaningful increase in burden to wire in a quarantine query result override here

cli/tests/upload.rs
```diff
@@ -1308,11 +1313,126 @@ async fn telemetry_upload_metrics_on_upload_success() {
...
+        QuarantineQueryResult::try_from(telemetry_request.quarantine_query_result).unwrap(),
+        QuarantineQueryResult::Disabled
+    );
 
     // HINT: View CLI output with `cargo test -- --nocapture`
     println!("{assert}");
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn telemetry_upload_metrics_quarantine_query_result_success() {
```
TylerJang27, line 1371: These CLI integration tests require a bit more custom setup because of the scenarios they test

proto/proto/upload_metrics.proto
```diff
@@ -5,6 +5,16 @@ package trunk.oss.flakytests_cli.v1;
...
 
+enum QuarantineQueryResult {
+  QUARANTINE_QUERY_RESULT_UNSPECIFIED = 0;
+  QUARANTINE_QUERY_RESULT_SUCCESS = 1;
+  QUARANTINE_QUERY_RESULT_FAILURE = 2;
+  QUARANTINE_QUERY_RESULT_DISABLED = 3;
+  QUARANTINE_QUERY_RESULT_SKIPPED = 4;
+  // rspec only
+  QUARANTINE_QUERY_RESULT_CACHED = 5;
+}
```
TylerJang27, line 21: - Unspecified: default, shouldn't be sent from new CLI versions in practice
- Success: we successfully got quarantined results back
- Failure: something went wrong. This could also be an auth error (403/404), but we don't differentiate rn here. Realistically we want to track both
- Disabled: either disabled via the command line/env var, or via the repo setting (from our db)
- Skipped: no tests actually failed, no need to look up quarantine config
- Cached: rspec only, quarantined results served from file on disk

test_report/src/report.rs
```diff
@@ -40,12 +40,21 @@ struct QuarantineConfigDiskCacheEntry {
     cached_at_secs: u64,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum QuarantineLookupSource {
```
TylerJang27, line 53: Special enum for the rspec/report case

test_report/tests/telemetry.rs
TylerJang27, line 1: "Integration" tests underlying the rspec flow. Meaningfully distinct from the CLI integration tests above

test_report/tests/telemetry.rs
```diff
@@ -0,0 +1,281 @@
...
+        env::set_var(TRUNK_USE_UNCLONED_REPO_ENV, "true");
+        env::set_var(TRUNK_REPO_HEAD_SHA_ENV, "");
+        env::set_var(TRUNK_REPO_HEAD_BRANCH_ENV, "");
+        env::set_var(TRUNK_REPO_HEAD_AUTHOR_NAME_ENV, "");
+        env::set_var("CI", "1");
+        env::set_var("GITHUB_JOB", "test-job");
+    }
+}
+
+fn publish_minimal_success_test(test_report: &MutTestReport) {
```
TylerJang27, line 59: This helper keeps things simple for most of the tests below. Note that test_report.publish() is what calls upload and the telemetry call

proto/proto/upload_metrics.proto
acatxnamedvirtue: Worth making any of those points an inline comment alongside any of those?

acatxnamedvirtue (approved):

cli/src/upload_command.rs
dfrankland: These arguments are getting kind of long and hard to decipher at call sites. I think we should consider creating an options struct.

proto/proto/upload_metrics.proto
dfrankland: agreed, I think those comments are nice :)

test_report/tests/telemetry.rs
```diff
@@ -0,0 +1,281 @@
...
+}
+
+#[tokio::test(flavor = "multi_thread")]
+#[serial]
+async fn telemetry_query_result_success_on_publish() {
+    cleanup_env_vars();
+    let temp_dir = tempdir().unwrap();
+    let repo_setup_res = setup_repo_with_commit(&temp_dir);
+    assert!(repo_setup_res.is_ok());
+    assert!(env::set_current_dir(&temp_dir).is_ok());
```
dfrankland: nit, optional: I'd say to just unwrap these since it's functionality equivalent and more succinct. Same goes for any other tests.

test_report/tests/telemetry.rs
```diff
@@ -0,0 +1,281 @@
...
+    let state = MockServerBuilder::new().spawn_mock_server().await;
+    set_mock_api_env(&state.host);
+    unsafe {
+        env::set_var(TRUNK_ORG_URL_SLUG_ENV, "test-org");
+        env::set_var(TRUNK_QUARANTINED_TESTS_DISK_CACHE_TTL_SECS_ENV, "0");
+    }
+
+    thread::spawn(|| {
```
dfrankland, line 93: is there a specific reason to do this in a separate thread? same goes for the other tests

test_report/tests/telemetry.rs
```diff
@@ -0,0 +1,281 @@
...
+    let requests = state.requests.lock().unwrap().clone();
+    assert_telemetry_quarantine_query_result(&requests, QuarantineQueryResult::Success);
+}
+
+#[tokio::test(flavor = "multi_thread")]
+#[serial]
+async fn telemetry_query_result_skipped_without_lookup_on_publish() {
+    cleanup_env_vars();
+    let temp_dir = tempdir().unwrap();
+    let _ = env::set_current_dir(&temp_dir);
```
dfrankland: nit, optional: I'd just unwrap on this rather than drop it. If the env isn't set, this test will fail, no? same goes for the other tests

test_report/tests/telemetry.rs
```diff
@@ -0,0 +1,281 @@
...
+#[serial]
+async fn telemetry_query_result_cached_on_publish() {
+    cleanup_env_vars();
+    clean_up_cache_files();
+
+    let temp_dir = tempdir().unwrap();
+    let _ = env::set_current_dir(&temp_dir);
+
+    let repo_url = "https://github.com/test-org/test-repo-cache-telemetry.git";
+    set_uncloned_repo_publish_env(repo_url);
```
dfrankland: nit, optional: Some tests define `repo_url` others just put the string inline for the `set_uncloned_repo_publish_env` call. IMO, make it consistent and put this string inline

dfrankland (approved):

cli/src/upload_command.rs
TylerJang27: I waffled on that. Will do. I think we have the clippy warning if it gets to be one more

proto/proto/upload_metrics.proto
TylerJang27: Fair enough. Will do

test_report/tests/telemetry.rs
TylerJang27: According to Cursor, our `publish` call in rspec does:

```rs
tokio::runtime::Builder::new_multi_thread()
    .build()
    .unwrap()
    .block_on(run_upload(...))
```

Without the thread spawn we get
```rs
thread 'telemetry_query_result_skipped_without_lookup_on_publish' (3635475) panicked at /home/tyler/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.48.0/src/runtime/scheduler/multi_thread/mod.rs:86:9:
Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test telemetry_query_result_skipped_without_lookup_on_publish ... FAILED
```

In summary:

> MutTestReport::publish() (and is_quarantined()) create their own tokio runtime internally via Runtime::new().block_on(...). Our tests use #[tokio::test] to run the mock server async, so calling MutTestReport directly inside the test body nests runtimes on the same thread.

> thread::spawn puts MutTestReport on a separate OS thread with no existing runtime, which matches how RSpec/Ruby calls it synchronously. Same pattern as the other test_report integration tests (quarantine.rs, publish.rs).

I presume that means we're doing something wrong in the report/rspec side, but not sure

dfrankland: It's kind of confusing, but I see why. I think it's because we can't do anything async due to Rust<->Ruby interop so we use a blocking Tokio executor nested in a function. And you cannot use a Tokio executor within another Tokio executor.

Call from Ruby:
https://github.com/trunk-io/analytics-cli/blob/8e5ffb4f97be103b306e5731f86c655ed74e979c/rspec-trunk-flaky-tests/lib/trunk_spec_helper.rb#L122

Place that blocks in Rust:
https://github.com/trunk-io/analytics-cli/blob/8e5ffb4f97be103b306e5731f86c655ed74e979c/test_report/src/report.rs#L506-L510

Don't feel like you need to make any code changes for this, but if I were to attempt to improve this, I would:
* Convert everything internal to Rust to use `async`
* Expose a Ruby function which _only_ calls the internal function and does the blocking

This way we could create a test which utilizes the `async` internal function directly and not have confusing multi-threaded tests :)

PR #1101, opened 2026-05-29
Support uploading from forked-repo PRs via x-trunk-public-repo-id

mmatheson:
## Summary

Wires up the CLI side of the server's [`x-trunk-public-repo-id`](https://github.com/trunk-io/trunk/pull/31970) support, so fork-PR runs — which can't read repo secrets — can authenticate to the flaky-tests endpoints with a **non-secret per-repo identifier** instead of an API token.

Ingest treats the identifier like a Mixpanel/Segment SDK key (tenant-routing, not a credential), sent as the `x-trunk-public-repo-id` header on the only two authenticated endpoints a fork-PR run hits: `POST /metrics/createBundleUpload` and `POST /metrics/getQuarantineConfig`.

## Changes

- **`constants`** — new `TRUNK_PUBLIC_REPO_ID_ENV` (`TRUNK_PUBLIC_REPO_ID`), captured in `TRUNK_ENVS_TO_CAPTURE` (it's non-secret).
- **`ApiClient::new`** — now takes `Option<String>` for both the token and the public repo id; errors only when **both** are absent/blank. The `trunk_api_client` (used by exactly those two endpoints) sends `x-api-token` and/or `x-trunk-public-repo-id` based on what's present. Telemetry still carries only the token (the server didn't extend the id there; token-less runs simply send no auth header).
- **`UploadArgs`** — `token` is now `Option<String>` with `required_unless_present = "public_repo_id"`, plus a new `--public-repo-id` / `TRUNK_PUBLIC_REPO_ID` arg with the inverse requirement. Threaded through `run_upload`, `test_command`, and `main`.

## Tests

- `requires_a_token_or_public_repo_id` — covers neither (error), blank-both (error), token-only (ok), public-repo-id-only (ok).
- `sends_public_repo_id_header_in_place_of_token` — asserts the header is sent and `x-api-token` is absent in public-repo-id mode.

## Verification

- `cargo check --workspace --all-targets` clean; `cargo test -p api` and the `upload` integration tests pass.
- Manual run against `api.trunk-staging.io` with `--public-repo-id` (and no token) reaches the endpoints and resolves the tenant from the id.
- Confirmed clap's `required_unless_present` counts env-provided values, so the env-only fork-PR path (credential set in workflow YAML) parses correctly.


api/src/client.rs
```diff
@@ -67,19 +67,40 @@ impl ApiClient {
...
+
+        // Either a secret API token or a non-secret public repo id authenticates the
+        // request. Fork PRs can't read repo secrets, so they rely on the public repo id.
+        let api_token = api_token.filter(|token| !token.trim().is_empty());
+        let public_repo_id = public_repo_id.filter(|id| !id.trim().is_empty());
+        if api_token.is_none() && public_repo_id.is_none() {
+            return Err(anyhow::anyhow!(
+                "Either a Trunk API token or a public repo id is required."
+            ));
         }
```
dfrankland, line 88: We should enforce only the API token or the public repo ID is passed at the CLI arguments level so there's no confusion. Instead of accepting `api_token: Option<String>` and `public_repo_id: Option<String>` we should only accept an enum of:
```rs
#[derive(Debug, Clone)]
enum AuthHeader {
  ApiToken(String),
  PublicRepoId(String),
}
```

The comment is much better suited as documentation on the CLI arguments than this deeply-nested runtime validation.

I know this previously was having validation in here, but that isn't a great pattern when we could just validate inputs at the time of CLI invocation.

api/src/client.rs
```diff
@@ -451,11 +483,14 @@ fn does_not_add_settings_if_domain_absent() {
 #[cfg(test)]
 mod tests {
```
dfrankland, line 484: optional: I would love if we could move all of these tests to the `tests` dir

test_utils/src/mock_server.rs
```diff
@@ -44,6 +45,9 @@ impl IntoResponse for RespError {
 #[derive(Debug, Default)]
 pub struct MockServerState {
     pub requests: Mutex<Vec<RequestPayload>>,
+    /// Request headers captured per endpoint (keyed by endpoint name, e.g.
+    /// "createBundleUpload"), so tests can assert on auth headers.
+    pub request_headers: Mutex<HashMap<String, HeaderMap>>,
```
dfrankland, line 50: I know its more work, but it would make a lot more sense to just capture the headers for each request and include it in `pub requests: Mutex<Vec<RequestPayload>>,` than adding `pub request_headers: Mutex<HashMap<String, HeaderMap>>`. Especially so if there are repeated API endpoint calls with different headers.

You might be able to prevent a large test rewrite by creating  a struct which [`Deref`](https://doc.rust-lang.org/std/ops/trait.Deref.html)s into its inner payload. For example:
```rs
use std::ops::Deref;

#[derive(Debug, Clone, PartialEq)]
pub struct HttpRequest<T> {
  payload: T,
  headers: HeaderMap,
}

impl<T> Deref for HttpRequest<T> {
    type Target = T;

    fn deref(&self) -> &Self::Target {
        &self.payload
    }
}
```

and then change the above to be

```rs
#[derive(Debug, Clone, PartialEq)]
pub enum RequestPayload {
    CreateBundleUpload(HttpRequest<CreateBundleUploadRequest>),
    GetQuarantineConfig(HttpRequest<GetQuarantineConfigRequest>),
    S3Upload(HttpRequest<PathBuf>),
    TelemetryUploadMetrics(HttpRequest<UploadMetrics>),
}
```

If that doesn't work, I recommend to just split this PR and do the test refactor separately.

cli/src/upload_command.rs
```diff
@@ -81,11 +81,18 @@ pub struct UploadArgs {
...
     )]
-    pub token: String,
+    pub token: Option<String>,
+    #[arg(
+        long,
+        required_unless_present = "token",
+        env = constants::TRUNK_PUBLIC_REPO_ID_ENV,
+        help = "Non-secret per-repo identifier, usable instead of --token on fork PRs where repo secrets are unavailable."
+    )]
+    pub public_repo_id: Option<String>,
```
dfrankland, line 95: I don't think this the best way of handling this. We really should have an enum that requires one or the other rather than doing validation in deeply-nested code (in the API client).

We can do our own manual parsing and validation via instrumenting `FromStr` on an enum. We will lose some things like the `env` option here, but we can document the env vars in the `help` string.

cli/tests/common/command_builder.rs
```diff
@@ -49,167 +50,168 @@ impl UploadArgs {
...
-                    }
-                }),
-        )
-        .chain(self.allow_empty_test_results.into_iter().flat_map(
-            |allow_empty_test_results: bool| {
-                if allow_empty_test_results {
-                    vec![String::from("--allow-empty-test-results")]
+        }
+        auth_args
+            .into_iter()
```
dfrankland, line 72: It's possible to keep the same pattern as before with chaining. `if` are expressions, not statements in Rust, so you can do `.chain(if {...} else {...})` no problem

cli/tests/common/command_builder.rs
```diff
@@ -522,6 +524,19 @@ impl<'b> CommandBuilder<'b> {
...
 
+    pub fn public_repo_id(&mut self, new_value: &str) -> &mut Self {
+        match &mut self.command_type {
+            CommandType::Upload { upload_args, .. } => {
+                upload_args.public_repo_id = Some(new_value.to_string());
+            }
+            CommandType::Test { upload_args, .. } => {
+                upload_args.public_repo_id = Some(new_value.to_string());
+            }
+            CommandType::Validate { .. } => {}
```
dfrankland, line 535: Why not do this for validate?

dfrankland (approved):
I'd rather we fix some of the things in this PR before merging / releasing, but there's no actual blockers

api/src/client.rs
```diff
@@ -114,8 +143,11 @@ impl ApiClient {
             header::CONTENT_TYPE,
             HeaderValue::from_static("application/x-protobuf"),
         );
-        telemetry_client_default_headers
-            .append(Self::TRUNK_API_TOKEN_HEADER, api_token_header_value);
+
+        if let Some(api_token_header_value) = api_token_header_value {
+            telemetry_client_default_headers
+                .append(Self::TRUNK_API_TOKEN_HEADER, api_token_header_value);
```
TylerJang27, line 149: This means that telemetry isn't being tracked for `TRUNK_PUBLIC_REPO_ID_HEADER`. This is a blindspot and needs a follow-up

cli/src/upload_command.rs
```diff
@@ -81,11 +81,18 @@ pub struct UploadArgs {
...
         env = constants::TRUNK_API_TOKEN_ENV,
         help = "Organization token. Defaults to TRUNK_API_TOKEN env var."
     )]
-    pub token: String,
+    pub token: Option<String>,
+    #[arg(
+        long,
+        required_unless_present = "token",
+        env = constants::TRUNK_PUBLIC_REPO_ID_ENV,
+        help = "Non-secret per-repo identifier, usable instead of --token on fork PRs where repo secrets are unavailable."
```
TylerJang27, line 93: Should clarify that this needs to be enabled as a repo setting too

constants/src/lib.rs
```diff
@@ -54,6 +55,7 @@ pub const TRUNK_DEBUG_ENV: &str = "TRUNK_DEBUG";
 pub const TRUNK_ENVS_TO_CAPTURE: &[&str] = &[
     TRUNK_PUBLIC_API_ADDRESS_ENV,
     TRUNK_API_CLIENT_RETRY_COUNT_ENV,
+    TRUNK_PUBLIC_REPO_ID_ENV,
```
TylerJang27, line 58: 👍

TylerJang27 (commented):
Some notes and +1 on Dylan's comments
👍

PR #1102, opened 2026-05-29
(Chore): Fix org slug for smoke tests

TylerJang27:
Staging only https://github.com/trunk-io/analytics-cli/actions/runs/26658683845/job/78584066328


dfrankland (approved):
Love it

PR #1107, opened 2026-06-02
fix(circleci): omit workflow field since CircleCI has no human-readable workflow name

acatxnamedvirtue:
## Problem

The **Job Run** column on the uploads page was displaying a raw UUID for CircleCI customers instead of a human-readable name. For example:

```
03e6d2b6-93c4-4f61-9edb-6bb3cc32f663 - feature-tests ↗
```

## Root Cause

`parse_circleci()` was setting `ci_info.workflow` from `CIRCLE_WORKFLOW_ID`, which is always a UUID. Unlike GitHub Actions (`GITHUB_WORKFLOW`) or Semaphore (`SEMAPHORE_PROJECT_NAME`), CircleCI has no built-in environment variable for a human-readable workflow name.

## Fix

Remove the `ci_info.workflow` assignment from `parse_circleci()`. The `workflow` field is left as `None`, so the Job Run column displays just the job name (e.g., `feature-tests ↗`) without a UUID prefix.

`CIRCLE_WORKFLOW_ID` is still used to construct the `job_url` — that is unchanged.

## Safety: downstream usages of `ci_info.workflow`

We searched both the `trunk` and `trunk2` repos for all usages of the `workflow` field to verify this change is safe:

**Storage** — `workflow` is stored as a nullable/defaulted column in two places: `bundle_meta.workflow` (Postgres) and `flaky_tests.upload_summary.workflow` (ClickHouse). Neither has an index, unique constraint, or foreign key on this column.

**Display** — The primary consumer is the Job Run column, assembled as `[workflow, jobName].filter(Boolean).join(" - ")` (`quarantining.ts`). With `workflow: None`, CircleCI jobs will display as `feature-tests ↗` instead of `{uuid} - feature-tests ↗`, which is the desired behavior.

**Outdated CLI version detection** — `getOutdatedVersionSources()` in `upload.ts` groups upload sources by `workflow + jobName` to detect stale CLI versions. Setting `workflow` to `None` means CircleCI sources group by job name alone. The UUID was meaningless as a grouping key, so this is an improvement.

**Proto** — `BundleUpload.workflow` is declared as `optional string` in the flaky tests service proto. An absent/null value is valid.

No filtering, sorting, deduplication, or indexing logic depends on `workflow` being populated for CircleCI specifically.

## Testing

All 32 env tests pass. 4 CircleCI-specific tests updated to assert `workflow: None`.

### DIFF
--- a/context/src/env/parser.rs
+++ b/context/src/env/parser.rs
@@ -527,7 +527,6 @@ impl<'a> CIInfoParser<'a> {
             });
         self.ci_info.actor = self.get_env_var("CIRCLE_USERNAME");
 
-        self.ci_info.workflow = self.get_env_var("CIRCLE_WORKFLOW_NAME");
         self.ci_info.job = self.get_env_var("CIRCLE_JOB");
     }
 
--- a/context/src/env/parser.rs
+++ b/context/src/env/parser.rs
@@ -527,7 +527,7 @@ impl<'a> CIInfoParser<'a> {
             });
         self.ci_info.actor = self.get_env_var("CIRCLE_USERNAME");
 
-        self.ci_info.workflow = self.get_env_var("CIRCLE_WORKFLOW_ID");
+        self.ci_info.workflow = self.get_env_var("CIRCLE_WORKFLOW_NAME");
         self.ci_info.job = self.get_env_var("CIRCLE_JOB");
     }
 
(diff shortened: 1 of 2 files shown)


TylerJang27:
Sadly not a real env var. Just something that some users do as a hack (I think same thing with the job id)
https://circleci.com/docs/reference/variables/#c-consent-modal

TylerJang27:
Should we be using `CIRCLE_JOB` instead? Can you audit some of the values we have stored for that in bundle_meta?

TylerJang27 (approved):
LGTM, we'll need to uprev our ref

PR #1108, opened 2026-06-02
(Feat): Support early exit in rspec when quarantine lookup fails

TylerJang27:
Respect new flag `TRUNK_QUARANTINE_QUERY_FAILURE_EXIT` in rspec. Tested with verifying that we exit immediately when set, but continue as normal otherwise


acatxnamedvirtue (approved):

dfrankland (commented):
lgtm

PR #1123, opened 2026-06-17
(Chore): Cleanup bazel labels

TylerJang27:
This PR makes Bazel target labels canonical on `internal.bin` through `bazel_run_information` and `bazel_build_information`, and removes production use of the deprecated `file_sets.resolved_label` path.

**Codeowners and Bazel run info**

`IntoTestCaseRunsOptions` in `context/src/junit/parser.rs` now takes an optional `TestRunnerConfig::Bazel` instead of a separate `bazel_label` argument. When `--use-bazel-target-for-codeowners` is enabled, `into_test_case_runs` uses `bazel_run_information.label` as a fallback for codeowners lookup and sets `test_runner_information` on each test case in the same pass.

On the upload path, `generate_internal_file_from_bep` in `cli/src/context.rs` builds `BazelRunInformation` from BEP XML and passes `TestRunnerConfig::Bazel`. The non-BEP `generate_internal_file` path no longer reads `resolved_label` from `file_sets`.

**meta.json and BEP parsing**

We stopped writing Bazel labels into `meta.json`. `TestRunnerReport.label` was removed, new bundles always serialize `resolved_label` as empty/omitted, and the field is kept only so older bundles still deserialize. BEP parsing no longer populates labels on `TestRunnerReport`; `LabelledTestSummary` was replaced with `TryFrom<&TestSummary> for TestRunnerReport` in `context/src/bazel_bep/common.rs`.

Upload tests now assert labels and codeowners on decoded `internal.bin` rather than on `file_sets` in `meta.json`.

## Context

A bit of a history lesson. Originally when we added bazel support, we didn't have `internal.bin`, and we were parsing everything raw in ingestion. At the time I/we wired bazel labels through the `file_sets` in `meta.json`, and we reassociated those in the pipeline.

When we switched to `internal.bin`, we added `bazel_build_information` to the proto, but we left `file_sets` for backwards compatibility. The new ingestion copied over some of this behavior, as did https://github.com/trunk-io/analytics-cli/pull/1116. https://github.com/trunk-io/trunk2/pull/4239 fixes the ingestion gap, this PR fixes the codeowners piece and cleans things up.

Existing tests pass as applicable for integration tests and codeowners association.

### DIFF
--- a/cli/src/context.rs
+++ b/cli/src/context.rs
@@ -26,14 +26,13 @@ use context::{
     junit::{
         bindings::BindingsReport,
         junit_path::JunitReportFileWithTestRunnerReport,
-        parser::{IntoTestCaseRunsOptions, JunitParser},
+        parser::{IntoTestCaseRunsOptions, JunitParser, TestRunnerConfig},
         validator::{JunitReportValidation, validate},
     },
     repo::{BundleRepo, RepoUrlParts},
 };
 use github_actions::extract_github_external_id;
 use lazy_static::lazy_static;
-use proto::test_context::test_run::test_case_run::TestRunnerInformation;
 use proto::test_context::test_run::{
     BazelAttemptNumber, BazelBuildInformation, BazelRunInformation, TestBuildResult, TestReport,
     TestResult, UploaderMetadata,
@@ -380,6 +379,24 @@ fn get_build_result_from_bep(bep_result: &BepParseResult, label: &str) -> BuildR
     }
 }
 
+fn bazel_run_information_from_bep_xml(
+    label: &str,
+    xml_file: &context::bazel_bep::common::BepXMLFile,
+) -> BazelRunInformation {
+    BazelRunInformation {
+        label: label.to_string(),
+        attempt_number: xml_file.attempt,
+        started_at: xml_file.start_time.map(|time| prost_wkt_types::Timestamp {
+            seconds: time.timestamp(),
+            nanos: time.timestamp_subsec_nanos() as i32,
+        }),
+        finished_at: xml_file.end_time.map(|time| prost_wkt_types::Timestamp {
+            seconds: time.timestamp(),
+            nanos: time.timestamp_subsec_nanos() as i32,
+        }),
+    }
+}
+
 pub fn generate_internal_file_from_bep(
     bep_result: &BepParseResult,
     temp_dir: &TempDir,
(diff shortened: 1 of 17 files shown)


cli/src/context.rs
```diff
@@ -380,6 +379,24 @@ fn get_build_result_from_bep(bep_result: &BepParseResult, label: &str) -> BuildR
     }
 }
 
+fn bazel_run_information_from_bep_xml(
```
TylerJang27, line 382: You may be asking, since we're removing labels? Where does it actually come from now? This bazel BEP parsing map returns it: https://github.com/trunk-io/analytics-cli/blob/787cccd36811ea36d3539979bf27747a29375617/cli/src/context.rs#L415-L418

cli/src/context.rs
```diff
@@ -418,36 +435,22 @@ pub fn generate_internal_file_from_bep(
...
-                                }),
-                                finished_at: xml_file.end_time.map(|time| {
-                                    prost_wkt_types::Timestamp {
-                                        seconds: time.timestamp(),
-                                        nanos: time.timestamp_subsec_nanos() as i32,
-                                    }
-                                }),
+                            test_runner_config: Some(TestRunnerConfig::Bazel {
+                                bazel_run_information: &bazel_run_information,
+                                use_bazel_target_for_codeowners,
```
TylerJang27, line 450: Rather than postprocessing and setting test_runner_information, we now funnel this through to the parsing (and have it be attached at construction)

cli/src/context.rs
```diff
@@ -466,14 +469,24 @@ pub fn generate_internal_file_from_bep(
                         continue;
                     };
 
+                    let bazel_run_information = BazelRunInformation {
+                        label: label.clone(),
+                        attempt_number: 0,
+                        started_at: None,
+                        finished_at: None,
+                    };
```
TylerJang27, line 477: This shouldn't happen in practice unless we fail parsing/an old BEP version comes through. The above branch should capture most of the BEP idiosyncracies

context/src/junit/parser.rs
```diff
@@ -140,14 +141,22 @@ enum CurrentReportState {
     Opened,
 }
 
+#[derive(Debug, Clone)]
+pub enum TestRunnerConfig<'a> {
+    Bazel {
+        bazel_run_information: &'a BazelRunInformation,
+        use_bazel_target_for_codeowners: bool,
+    },
+}
```
TylerJang27, line 150: yay enums

context/src/junit/parser.rs
```diff
@@ -360,8 +369,14 @@ impl JunitParser {
                     if codeowners.is_some() {
                         let codeowners_path = if !file.is_empty() {
                             Some(file.clone())
+                        } else if let Some(TestRunnerConfig::Bazel {
+                            bazel_run_information,
+                            use_bazel_target_for_codeowners: true,
+                        }) = test_runner_config
+                        {
+                            bazel_label_to_package_path(&bazel_run_information.label)
```
TylerJang27, line 377: This is the codeowners association piece that Dylan added. I considered changing this codeowners to all be post-processing, but I think that's antithetical to the current way this is all organized and would be a bigger regression risk

cli/src/context.rs
```diff
@@ -380,6 +379,24 @@ fn get_build_result_from_bep(bep_result: &BepParseResult, label: &str) -> BuildR
     }
 }
 
+fn bazel_run_information_from_bep_xml(
+    label: &str,
+    xml_file: &context::bazel_bep::common::BepXMLFile,
+) -> BazelRunInformation {
+    BazelRunInformation {
+        label: label.to_string(),
```
dfrankland: If you immediately allocate for a reference, you should just pass the value directly as an argument.

```suggestion
    label: String,
    xml_file: &context::bazel_bep::common::BepXMLFile,
) -> BazelRunInformation {
    BazelRunInformation {
        label,
```

context/src/junit/parser.rs
```diff
@@ -140,14 +141,22 @@ enum CurrentReportState {
     Opened,
 }
 
+#[derive(Debug, Clone)]
+pub enum TestRunnerConfig<'a> {
+    Bazel {
+        bazel_run_information: &'a BazelRunInformation,
```
dfrankland: This doesn't seem to need a reference, unless I'm missing something. We're already allocating the label and the timestamp is essentially `Copy` even if it doesn't implement that trait. Also, below we end up cloning it anyways.

context/src/bazel_bep/common.rs
```diff
@@ -284,38 +281,28 @@ impl BepParseResult {
     }
 }
 
-struct LabelledTestSummary<'a> {
-    test_summary: &'a TestSummary,
-    label: Option<String>,
-}
-
-impl<'a> TryFrom<&LabelledTestSummary<'a>> for TestRunnerReport {
+impl TryFrom<&TestSummary> for TestRunnerReport {
```
dfrankland: Seems like we clone almost everything internal to `TestSummary`, so there's not really a good reason to have `TryFrom<&TestSummary>` instead of `TryFrom<TestSummary>`

cli/tests/upload.rs
```diff
@@ -393,6 +389,15 @@ async fn upload_bundle_using_bep() {
...
 
     let test_case_run = &report.test_case_runs[0];
+    let bazel_run_information = match test_case_run.test_runner_information.as_ref() {
+        Some(
+            proto::test_context::test_run::test_case_run::TestRunnerInformation::BazelRunInformation(
+                bazel_run_information,
+            ),
+        ) => bazel_run_information,
+        _ => panic!("Expected BazelRunInformation"),
+    };
```
dfrankland: Assert matches does this really nicely (and it's already used in several workspace crates):
https://docs.rs/assert_matches/latest/assert_matches/index.html

```suggestion
    let bazel_run_information = assert_matches!(
        test_case_run.test_runner_information.as_ref(),
        Some(
            proto::test_context::test_run::test_case_run::TestRunnerInformation::BazelRunInformation(
                bazel_run_information,
            ),
        ) => bazel_run_information
    );
```

same with the below instances using `match` and `panic`

dfrankland (approved):

PR #1131, opened 2026-06-29
Add TRUNK_API_CLIENT_RETRY_DEADLINE_SECS to bound API retry time

dfrankland:
## Problem

The API client's retry loop (`api/src/call_api.rs`) only honored a retry *count* (`TRUNK_API_CLIENT_RETRY_COUNT`). Total wall-clock time was otherwise uncapped — per-request timeouts plus up to ~62s of exponential backoff (2/4/8/16/32s) across retries — so a slow/failing endpoint could spin for minutes (this is what surfaced the repeated `…is taking longer than expected` messages customers reported).

## Change

- New optional env var **`TRUNK_API_CLIENT_RETRY_DEADLINE_SECS`**: an overall wall-clock budget for a call's retry loop. The loop gives up once `elapsed + next_backoff >= deadline`, returning the most recent error. Pairs with `TRUNK_API_CLIENT_RETRY_COUNT` to tune how fast/often the CLI gives up.
- Replaces `tokio-retry`'s `RetryIf` with an explicit loop (behavior-identical when the deadline is unset) so the deadline can short-circuit while still returning a real error — the generic `A::Error` includes `reqwest::Error`, which has no public constructor, so a `timeout()`/`select!` wrapper couldn't synthesize one. We still use `tokio-retry`'s `ExponentialBackoff` + `Action`.
- Read the env vars once via `LazyLock` (`DEFAULT_DELAY`, `RETRY_DEADLINE`).
- Captured `TRUNK_API_CLIENT_RETRY_DEADLINE_SECS` in `TRUNK_ENVS_TO_CAPTURE` for bundle-metadata debugging.

## Notes

- The deadline is checked **between attempts**, so a single in-flight request can overshoot by up to its own reqwest per-request timeout (30s API / 1s telemetry); each attempt is already bounded by that timeout. Immediate mid-request abort on CI cancellation is handled separately in #1132 (signal handling).
- Customer mitigations enabled by this: `TRUNK_API_CLIENT_RETRY_COUNT=0` and/or a small `TRUNK_API_CLIENT_RETRY_DEADLINE_SECS`.

## Testing

- Added two deterministic unit tests (paused clock, explicit deadlines): `stops_retrying_once_deadline_would_be_exceeded` and `retries_full_count_when_no_deadline_set`.
- All 9 `call_api` tests pass; existing retry-count/backoff/`should_retry` tests unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1132, opened 2026-06-29
Abort the CLI immediately on SIGTERM/SIGINT (CI cancellation)

dfrankland:
## Problem

GitHub Actions (and most CI) cancel a job by signaling the process tree. The CLI installed **no** signal handlers and blocked on its work, so a cancellation couldn't fully complete until slow operations (e.g. the API retry loop or telemetry upload) finished on their own — customers reported jobs that wouldn't cancel until the slow upload completed.

## Change

- `main` now races the command against termination signals via `tokio::select!`. On SIGTERM/SIGINT (Unix) or Ctrl-C (Windows), the in-flight work future is **dropped** — cancelling outstanding requests and the retry loop immediately — and the process exits with the conventional `128 + signal` code.
- If handler installation fails, falls back to the OS default disposition, so behavior is no worse than before.
- Signal handling lives in its own module (`cli/src/shutdown.rs`) to keep `main.rs` focused; exit codes are named constants (`EXIT_CODE_SIGINT` = 130, `EXIT_CODE_SIGTERM` = 143) with a comment noting these come from the POSIX `128 + signal` convention, which the `exitcode`/sysexits crate doesn't define.
- Adds the `signal` feature to the `cli` crate's tokio dependency (pulls `signal-hook-registry` into `Cargo.lock`).

## Notes

This is the complementary fix to #1131 (retry deadline): the deadline bounds *total retry time* but is checked between attempts; this PR provides true *immediate* mid-request abort when CI actually cancels. The two are independent and can merge in any order.

## Testing

- `cargo build -p trunk-analytics-cli` clean on this branch standalone (off `main`, without the #1131 changes).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5792,6 +5792,7 @@ dependencies = [
  "libc",
  "mio 1.1.0",
  "pin-project-lite",
+ "signal-hook-registry",
  "socket2 0.6.1",
  "tokio-macros",
  "windows-sys 0.61.2",
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -32,6 +32,7 @@ prost-wkt-types = { version = "0.5.1", features = ["vendored-protox"] }
 tokio = { version = "*", default-features = false, features = [
   "rt-multi-thread",
   "macros",
+  "signal",
 ] }
 tempfile = "3.2.0"
 tracing = "0.1.41"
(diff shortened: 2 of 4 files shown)


max-trunk (approved):

acatxnamedvirtue (approved):

TylerJang27 (commented):
👍

PR #1135, opened 2026-07-06
fix(test_report): write trunk_output.bin atomically via temp file + rename

acatxnamedvirtue:
## Summary

When `TRUNK_LOCAL_UPLOAD_DIR` is set, the rspec plugin saves `trunk_output.bin` at suite end via `TestReport::save`, which used `fs::write` — a truncate-then-write of the final path. That is not atomic: concurrent writers (e.g. parallel rspec shards saving to the same directory) can interleave so the tail of a longer report survives past a shorter one's EOF, corrupting the protobuf.

We observed exactly this in production: a customer bundle contained a fully valid `TestReport` followed by 942 bytes of a stale, larger report (with multiple layered `uploader_metadata` trailers). Ingestion deterministically fails on the stale bytes with `decode error: failed to decode Protobuf message: groups not supported`, and the upload retries until it is dead-lettered — silently losing the run's results. See TRUNK-18745 for the full byte-level analysis and evidence.

## Change

`TestReport::save` now writes the serialized report to a unique temp file (`tempfile::NamedTempFile`) in the destination directory and atomically renames it into place. The final path only ever holds one complete report — a reader or competing writer can never observe a partially-overwritten file. The temp file lives in the same directory so the rename cannot cross filesystems, and the existing mtime freshening is preserved (applied to the temp file before the rename).

An earlier revision of this PR deleted stale `trunk_output.bin` files at rspec suite start (Ruby side); that is reverted here in favor of the atomic write, which covers both the sequential-staleness and concurrent-writer cases at the source.

## Remaining gap

With N parallel shards writing the same path, atomic writes guarantee a *valid* file but still mean last-writer-wins — N−1 shards' results are dropped. Distinct per-process output paths are the follow-up if shard-level completeness is needed (tracked in TRUNK-18745).

## Testing

- `cargo check -p test_report` clean (no new warnings), `cargo fmt`, `trunk check` no new issues.
- Existing `test_report` suite passes, including `publish_try_save`.
- New regression test `try_save_leaves_decodable_report_under_concurrent_writers`: 8 threads with very differently sized reports race 20 saves each on one path; the surviving file must decode cleanly. Deterministically green with atomic rename; the old truncate-then-write could interleave and fail it.

### DIFF
--- a/test_report/tests/publish.rs
+++ b/test_report/tests/publish.rs
@@ -1,6 +1,15 @@
 mod common;
 
-use std::{env, fs, io::BufReader, path::Path, thread};
+use std::{
+    env, fs,
+    io::BufReader,
+    path::Path,
+    sync::{
+        Arc,
+        atomic::{AtomicBool, Ordering},
+    },
+    thread,
+};
 
 use assert_matches::assert_matches;
 use bundle::{BundleMeta, FileSetType};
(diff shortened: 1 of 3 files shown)


acatxnamedvirtue:
@claude review please

test_report/tests/publish.rs
```diff
@@ -274,6 +274,59 @@ fn publish_try_save() {
     assert_eq!(test_result.test_case_runs.len(), 0);
 }
 
+#[test]
+fn try_save_leaves_decodable_report_under_concurrent_writers() {
```
dfrankland, line 287: Does this test complete in a reasonable amount of time?

dfrankland (approved):

test_report/tests/publish.rs
acatxnamedvirtue: locally: 0.08s
CI: 0.95s

test_report/tests/publish.rs
```diff
@@ -274,6 +274,59 @@ fn publish_try_save() {
...
+    cleanup_env_vars();
+    let temp_dir = tempdir().unwrap();
+    setup_quarantine_disk_cache_dir(&temp_dir);
+    let dir = temp_dir.path().to_str().unwrap().to_string();
+
+    // Writers with very different report sizes race on the same path. A
+    // truncate-then-write save can interleave so the tail of a longer report
+    // survives past a shorter one's EOF, corrupting the protobuf; the atomic
+    // temp-file + rename must always leave a complete report.
```
TylerJang27: Were you able to repro with this?

TylerJang27 (approved):

test_report/tests/publish.rs
dfrankland: Sounds good to me :)

test_report/tests/publish.rs
acatxnamedvirtue: restructured it a bit, but yes

PR #1136, opened 2026-07-06
fix(rspec): report correct status for pending/skipped examples

dfrankland:
## Problem

The RSpec plugin didn't handle `pending` examples correctly, because RSpec expresses a pending example's outcome as the **inverse** of its body:

| pending example | RSpec `status` | build | plugin reported (before) |
|---|---|---|---|
| body **still fails** (expectation met) | `:pending` | green | **skipped** ❌ |
| body **now passes** (fixed → `PendingExampleFixedError`) | `:failed` | red | failure (unhelpful error) |

Reporting the raw status meant a still-broken pending test was recorded as **skipped** — never counted as failing, so it could never be quarantined — while a fixed one was recorded as a failure. Genuine `skip`/`xit` (which RSpec also reports as `:pending`) couldn't be told apart from a failing pending test.

## Fix

Split the status decision into `TrunkAnalyticsListener#status_and_exception`, which reports the outcome RSpec actually decided (which also matches the build result):

- `skip` / `xit` (never ran) → **skipped** — detected explicitly via `metadata[:skip]`, up front, since RSpec reports skips as `:pending` too
- `pending`, body still failing → **success** (the "should fail" expectation was met)
- `pending`, body now passing → **failure** (`PendingExampleFixedError`; error pulled from `example.exception`)

RSpec's own build pass/fail behavior is left untouched (`#set_exception`) — a fixed pending example still fails the run; we only change what status/error is reported to Trunk.

## Tests

Adds `test/pending_status_spec.rb` to the gem suite. It runs **real** pending/skip examples inside an `RSpec::Core::Sandbox` (so the fixed-pending's failure doesn't fail this suite) and asserts the mapping against genuine RSpec behavior.

## Verification

Built the native extension and ran the full gem suite locally: **18 examples, 0 failures** (13 existing + 5 new). Also confirmed end-to-end that the reported statuses serialize as `skipped / success / failure` as expected.

### DIFF
--- a/rspec-trunk-flaky-tests/lib/trunk_spec_helper.rb
+++ b/rspec-trunk-flaky-tests/lib/trunk_spec_helper.rb
@@ -336,7 +336,7 @@ class TrunkAnalyticsListener
 
   # trunk-ignore(rubocop/Metrics/CyclomaticComplexity,rubocop/Metrics/AbcSize,rubocop/Metrics/MethodLength)
   def add_test_case(example)
-    exception = example.exception || example.metadata[:quarantined_exception]
+    status, exception = status_and_exception(example)
     failure_message = ''
     backtrace = ''
     if exception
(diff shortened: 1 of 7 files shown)


rspec-trunk-flaky-tests/lib/trunk_spec_helper.rb
```diff
@@ -355,22 +355,60 @@ def add_test_case(example)
...
-    when :pending
-      status = Status.new('skipped')
-    end
     parent_name = example.example_group.metadata[:description]
     parent_name = parent_name.empty? ? 'rspec' : parent_name
     @testreport.add_test(id, name, classname, file, parent_name, line, status, attempt_number,
                          started_at, finished_at, failure_message || '', backtrace || '', is_quarantined)
   end
+
+  # Determine the Trunk status to report for an example, plus the exception (if
```
TylerJang27, line 366: Note that this could be a breaking change for users of `TRUNK_LOCAL_UPLOAD_DIR`, but I agree with your choices here

rspec-trunk-flaky-tests/lib/trunk_spec_helper.rb
```diff
@@ -355,22 +355,60 @@ def add_test_case(example)
...
+    return [Status.new('skipped'), nil] if example.metadata[:skip]
+
+    case result.status
+    when :passed
+      # A quarantined failure is recorded as :passed by RSpec (see #set_exception),
+      # so report it as a failure but carry the original quarantined exception.
+      quarantined_exception ? [Status.new('failure'), quarantined_exception] : [Status.new('success'), nil]
+    when :failed
+      # Includes a pending example whose body unexpectedly passed: RSpec reports it
+      # as :failed with a PendingExampleFixedError (on example.exception) and breaks
```
TylerJang27, line 400: 🙄

rspec-trunk-flaky-tests/test/pending_status_spec.rb
TylerJang27, line 1: Sanity check: this should rspec exit code 0, but we would see that is as nonzero nonquarantined failures.

Let's make sure this runs for a few cycles of smoke tests

You'll need to enumerate this file in `.github/actions/test_ruby_gem_uploads/action.yaml`, probs in `run-regular-tests`

TylerJang27 (approved):
One blocker on the smoke tests

rspec-trunk-flaky-tests/test/pending_status_spec.rb
dfrankland: I've configured the smoke test action to optionally run the unit tests with the compiled gem. The ruby.yaml won't run the unit tests, but the smoke_test.yaml and smoke_test_main.yaml will.

PR #1137, opened 2026-07-07
Add musl Linux gem support via -linux-gnu/-linux-musl split (Alpine)

TylerJang27:
## Summary

Publish the precompiled Linux gems as explicit **`x86_64-linux-gnu` / `aarch64-linux-gnu`** and **`x86_64-linux-musl` / `aarch64-linux-musl`** variants — replacing the bare `x86_64-linux` / `aarch64-linux` builds — so musl-based distros (Alpine, etc.) get a gem that actually loads. This raises the floor to **Ruby >= 3.1**.

## Background: the bug

The bare `*-linux` gems are glibc-linked. As of 0.13.0 the native extension references `__res_init`, a glibc-exclusive resolver symbol musl lacks, so loading on Alpine fails:

```
LoadError: Error relocating .../rspec_trunk_flaky_tests.so: __res_init: symbol not found
```

0.12.0 happened to load on musl by luck; 0.13.x broke it, and the `ruby`-platform gem intentionally raises, so Alpine users were pinned to 0.12.0.

## Why two gems (`-gnu` + `-musl`) and not one cross-compatible `-linux`

A single bare `-linux` gem cannot safely serve both libcs for *this* extension:

- **glibc-linked → doesn't load on musl.** A glibc build carries glibc-versioned symbol references (`__res_init`, `…@GLIBC_x.y`) that musl doesn't provide. The 0.12→0.13 breakage is exactly this — "it works on Alpine" was incidental, not a guarantee, and any dependency bump can reintroduce a glibc-only symbol.
- **musl-static labeled `-linux` → unsafe on glibc.** Tagging a static-musl build as bare `-linux` makes glibc hosts select it too, which means dlopening a musl object into a glibc Ruby process (two libcs in one address space — UB: cross-libc `malloc`/`free`, TLS, `errno`, plus static-TLS `dlopen` failures).
- **This gem is a worst case for a universal build:** it does networking (`reqwest`), TLS (vendored OpenSSL in `context`), and DNS / git-over-HTTPS. DNS resolution (`getaddrinfo`/NSS/`__res_init`) is precisely where glibc and musl diverge, and glibc's resolver can't be statically linked — so the one subsystem that would need to be portable fundamentally isn't.

### This is the documented, sanctioned approach — and the bare `-linux` + `-musl` combo is explicitly broken

The maintainer of **both Nokogiri and rake-compiler-dock** ([@flavorjones](https://github.com/flavorjones)) documented the rule in [rake-compiler-dock#117](https://github.com/rake-compiler/rake-compiler-dock/issues/117):

> Either ship a `-linux` platform gem that works for both gnu and musl systems, or ship both `-linux-gnu` and `-linux-musl` platform gems. **Do NOT ship `-linux` and `-linux-musl` platform gems for the same version.**

Concretely, shipping bare `-linux` alongside `-linux-musl` (what this PR's first pass did) causes:

- **`bundle install` hard-fails** with `Could not find gems matching '…' valid for all resolution platforms` — bare `-linux` is ambiguous (glibc *or* musl), so Bundler can't produce a consistent multi-platform resolution.
- On **Ruby 3.0**, musl systems install the *wrong* (glibc `-linux`) gem.

Since a universal `-linux` isn't achievable for this extension, `-linux-gnu` + `-linux-musl` is the correct — and only unambiguous — option. rake-compiler-dock supports these targets since [v1.5.0](https://github.com/rake-compiler/rake-compiler-dock/blob/main/CHANGELOG.md) (bare `*-linux` is now just an alias for `*-linux-gnu`); our `Gemfile.lock` pins 1.11.0.

## Why Ruby 3.0 is dropped

The gnu/musl platform distinction — including matching a `-gnu` gem to a host that reports itself as bare `x86_64-linux` — requires **RubyGems 3.3.22 / Bundler 2.3.21** (per rake-compiler-dock#117). That RubyGems generation first ships with **Ruby 3.1**. Stock Ruby 3.0 ships RubyGems 3.2.x, which does **not** understand `-gnu`, so a `-gnu` gem would fall through to the raising stub *even for glibc users* on Ruby 3.0. Ruby 3.0 has also been EOL since March 2024. So `required_ruby_version` is bumped to `>= 3.1` and 3.0 is removed from the build/test matrices.

## Prior art

This is the same path other native-extension gems took for musl support:

- **Nokogiri** switched from a single bare `-linux` gem to separate `x86_64-linux-gnu` + `x86_64-linux-musl` (and aarch64/arm) native gems in [v1.18.0](https://github.com/sparklemotion/nokogiri/releases/tag/v1.18.0), via rake-compiler-dock v1.5.1 (contributed by @flavorjones). Published artifacts, e.g. [`nokogiri 1.18.9-x86_64-linux-musl`](https://rubygems.org/gems/nokogiri/versions/1.18.9-x86_64-linux-musl).
- **grpc** — [grpc/grpc#39443](https://github.com/grpc/grpc/issues/39443) "Build separate binary gems for -linux-gnu and -linux-musl".

## Changes

- **Release matrix:** `x86_64-linux-gnu`, `x86_64-linux-musl`, `aarch64-linux-gnu`, `aarch64-linux-musl`, `arm64-darwin`, `x86_64-darwin`, `ruby`.
- **musl validation folded into the shared `test_ruby_gem_uploads` action** via an `alpine` input: when set, it skips the host `ruby/setup-ruby`/bundler steps and runs the *same* smoke specs inside a `ruby:*-alpine` container (docker) against the precompiled musl gem — exercising the full runtime (native load, quarantine lookup over TLS/DNS, result serialization, upload). `test-musl-gem` is a thin job that calls the action with `alpine: true`; `publish_ruby_gem` is gated on it.
- **Ruby >= 3.1:** `required_ruby_version` bumped; 3.0 dropped from cross-gem, `test-ruby-gem`, and the local `rake native` task.
- **Docs / loader:** README supported-platforms + prerequisites and the `LoadError` message updated to the gnu/musl names.

## Rollout note

Consumers on musl must be on **Bundler >= 2.5.6**; earlier Bundler omits the musl native gem from the lockfile ([ruby/rubygems#7432](https://github.com/ruby/rubygems/issues/7432)). Basic gnu/musl resolution needs RubyGems >= 3.3.22 / Bundler >= 2.3.21 (ships with Ruby >= 3.1).

## Sources

- [rake-compiler-dock#117](https://github.com/rake-compiler/rake-compiler-dock/issues/117) — canonical guidance: ship `-linux-gnu` + `-linux-musl`, never bare `-linux` + `-linux-musl`; documents the `bundle install` resolution failure and the RubyGems 3.3.22 / Bundler 2.3.21 (2.5.6 for musl) requirements.
- [rake-compiler-dock CHANGELOG](https://github.com/rake-compiler/rake-compiler-dock/blob/main/CHANGELOG.md) — `-gnu`/`-musl` targets added in 1.5.0; bare `-linux` is an alias for `-linux-gnu`.
- [Nokogiri v1.18.0 release notes](https://github.com/sparklemotion/nokogiri/releases/tag/v1.18.0) — reference implementation adopting the `-gnu`/`-musl` split.
- [grpc/grpc#39443](https://github.com/grpc/grpc/issues/39443) — same split in another large native gem.
- [ruby/rubygems#7432](https://github.com/ruby/rubygems/issues/7432) — musl native gem missing from lockfile on Bundler < 2.5.6.
- [Andrew Nesbitt — Platform Strings](https://nesbitt.io/2026/02/17/platform-strings.html) — overview of RubyGems linux platform matching.

https://claude.ai/code/session_01RumF5roNQnj2XsBS6erPVn

### DIFF
--- a/.github/workflows/release_ruby_gem.yml
+++ b/.github/workflows/release_ruby_gem.yml
@@ -156,9 +156,7 @@ jobs:
           artifact-pattern: cross-gem-${{ matrix.platform.name }}
           knapsack-pro-test-suite-token-rspec: ${{ secrets.KNAPSACK_PRO_TEST_SUITE_TOKEN_RSPEC }}
 
-  # test-ruby-gem covers glibc/darwin via ruby/setup-ruby, which has no Alpine
-  # support. The same action runs the musl gems inside a ruby:*-alpine
-  # container (alpine: true) so they are validated on a musl host.
+  # Validate the musl gems inside Alpine (ruby/setup-ruby has no Alpine support).
   test-musl-gem:
     name: Test ${{ matrix.platform.name }} gem on Alpine (ruby ${{ matrix.ruby-version }})
     runs-on: ${{ matrix.platform.os }}
--- a/.github/workflows/release_ruby_gem.yml
+++ b/.github/workflows/release_ruby_gem.yml
@@ -156,12 +156,12 @@ jobs:
           artifact-pattern: cross-gem-${{ matrix.platform.name }}
           knapsack-pro-test-suite-token-rspec: ${{ secrets.KNAPSACK_PRO_TEST_SUITE_TOKEN_RSPEC }}
 
-  # The glibc-linked *-linux gems fail to load on Alpine/musl (e.g.
-  # "Error relocating ... __res_init: symbol not found"). The musl-targeted gems
-  # must therefore be validated inside an Alpine container rather than on the
-  # glibc-based ubuntu runners used by test-ruby-gem. ruby/setup-ruby does not
-  # support Alpine, so we download the artifact on the host and exercise the gem
-  # inside an official ruby:*-alpine image via docker.
+  # test-ruby-gem covers glibc/darwin via ruby/setup-ruby, which has no Alpine
+  # support. So we validate the musl gems in an official ruby:*-alpine image:
+  # the artifact is downloaded on the host and the smoke specs run inside the
+  # container via docker. One Ruby version per arch is enough — the __res_init
+  # failure is a libc-linkage issue shared by every per-Ruby .so in a gem, and
+  # per-Ruby coverage already lives in test-ruby-gem.
   test-musl-gem:
     name: Test ${{ matrix.platform.name }} gem on Alpine (ruby ${{ matrix.ruby-version }})
     runs-on: ${{ matrix.platform.os }}
@@ -169,11 +169,6 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # The __res_init failure is a libc-linkage issue: every per-Ruby .so in
-        # a musl gem is linked the same way, so one Ruby version proves the
-        # linkage for all. Per-Ruby load-path/ABI coverage is already handled by
-        # test-ruby-gem. We do exercise both arches, since each is a separately
-        # built artifact.
         ruby-version: ["3.4"]
         platform:
           - name: x86_64-linux-musl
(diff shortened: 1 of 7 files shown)


rspec-trunk-flaky-tests/Gemfile.lock
```diff
@@ -30,11 +30,13 @@ GEM
     rspec-support (3.13.2)
 
 PLATFORMS
+  aarch64-linux-musl
   arm64-darwin
   arm64-linux
   ruby
   x86_64-darwin
   x86_64-linux
```
dfrankland, line 37: Do these need gnu?

.github/workflows/release_ruby_gem.yml
```diff
@@ -154,9 +156,69 @@ jobs:
...
+
+          # Forward the (non-secret) GitHub CI context so the gem detects CI and
+          # derives repo metadata just as it does in test-ruby-gem on the host.
+          env | grep -E '^GITHUB_' > "${RUNNER_TEMP}/ci.env" || true
+
+          # Run the same specs test-ruby-gem uses, inside Alpine, against the
+          # precompiled musl gem. This dlopens the native extension (where a
+          # glibc-only build dies on musl with __res_init), fetches the
+          # quarantine list over TLS/DNS, serializes results in the extension,
+          # and uploads — exercising the full runtime, not just gem loading.
```
dfrankland: Uh, this is not right. Since we're testing the musl compiled build, there's no dlopen. The problem before was a "gnu" build attempting to dlopen on alpine (and alpine doesn't have shared libraries to dlopen)

.github/workflows/release_ruby_gem.yml
```diff
@@ -154,9 +156,69 @@ jobs:
...
+          path: pkg
+          merge-multiple: true
+
+      - name: Run smoke specs against the musl gem on Alpine
+        shell: bash
+        env:
+          TRUNK_PUBLIC_API_ADDRESS: https://api.trunk-staging.io
+          TRUNK_ORG_URL_SLUG: trunk-staging-org
+          TRUNK_API_TOKEN: ${{ secrets.TRUNK_STAGING_ORG_API_TOKEN }}
+          TRUNK_TEST_COLLECTION_ID: T5yKSn9h
```
dfrankland: Maybe this should be moved to the smoke test action and made a conditional step that can be enabled to run?

dfrankland (approved):

rspec-trunk-flaky-tests/Gemfile.lock
TylerJang27: No — reverted. These build-time deps are all pure Ruby, and the rake-compiler-dock containers cross-compile from a glibc x86_64 host, so `Gem::Platform.local` inside is bare `x86_64-linux` (the documented alias for `-gnu`) regardless of the target. So none of the extra `-gnu`/`-musl` entries were actually needed here — I've reverted `PLATFORMS` to the original set.

---
_Generated by [Claude Code](https://claude.ai/code)_

.github/workflows/release_ruby_gem.yml
TylerJang27: Good catch — that comment was describing the *old* gnu-on-musl load failure, not what this step does. The musl-compiled build loads fine; the point of the step is just to exercise it end to end. Removed the dlopen/`__res_init` framing (in d62af6f).

---
_Generated by [Claude Code](https://claude.ai/code)_

.github/workflows/release_ruby_gem.yml
TylerJang27: Done (8d654f7). Folded it into `test_ruby_gem_uploads` behind a new `alpine` input: when `alpine: true`, the action skips the host `ruby/setup-ruby`/bundler steps and instead runs the smoke specs inside a `ruby:*-alpine` container via docker against the precompiled musl gem. `test-musl-gem` is now just a thin job that calls the action with `alpine: "true"`, gated on the musl matrix entries — so both glibc and musl go through the same action.

---
_Generated by [Claude Code](https://claude.ai/code)_

TylerJang27:
Release plan: bump a minor (not patch version), release a beta, validate in the load test repo and with the customer, and do a prod release with the minor bump

PR #1146, opened 2026-07-22
fix(cli): fold variant into report-supplied test ids for local quarantine checks

TylerJang27:
## Context

We were erroneously ignoring the passed variant when generating the test case ids we used to calculate quarantining when an ID was provided by the test runner (i.e. XCResult). This was not an issue before because our existing iOS users were not uploading with variants. Fixed the bug and added regression tests.

## Summary

- `convert_case_to_test` (in `cli/src/context_quarantine.rs`) used a report source's own stable test identifier as-is whenever a JUnit/xcresult test case carried a non-empty `"id"` extra attribute (e.g. xcresult's per-test-method identifier from `XCResult::generate_id`), bypassing `gen_info_id` entirely.
- That identifier does not encode `--variant`. Every other id derivation in this codebase (`JunitParser::into_test_case_runs` for JUnit-sourced tests, and bundle ingestion server-side) folds `variant` into the id via `gen_info_id`/`gen_info_id_base`.
- Net effect: for **any xcresult upload run with `--variant` set**, the CLI's local quarantine check compared a variant-less id against the server's variant-scoped quarantine list — a permanent mismatch, regardless of whether the test was actually quarantined. The "Learn more" link printed alongside an unquarantined failure (built from that same id via `url_for_test_case`) also pointed at the wrong test entity on the dashboard, since the dashboard identifies tests by the variant-scoped id.
- This reproduced as: a failing test that *was* quarantined server-side (confirmed present in that test's continuous history in the dashboard) still showed `Quarantined: 0` locally, and the printed dashboard link resolved to an unrelated test.

## Fix

Route the report-supplied id through `Test::generate_custom_uuid` — already used for exactly this purpose in `Test::new` — instead of assigning it directly to `test.id`. This folds `variant` into the id (matching `Test::set_id`'s existing behavior for tests with no report-supplied id, and matching `gen_info_id`'s behavior used elsewhere for JUnit/server-side ingestion). Behavior is unchanged when `variant` is empty (the common case for non-variant uploads), since `gen_info_id_base` returns the report-supplied id unchanged in that case.

I confirmed no other call site in the crate has this same pattern — `test_report::add_test` (used by the Python/JS/Ruby language-runtime bindings, a separate non-JUnit/non-xcresult path) just stores whatever id its caller already computed, so it's unaffected by this change and out of scope here.

## Test plan

- [x] Added `test_convert_case_to_test_folds_variant_into_report_supplied_id`: given a report-supplied id shaped like `XCResult::generate_id`'s output and a non-empty variant, asserts the resulting `Test.id` (a) differs from the raw report-supplied id, and (b) matches `gen_info_id` computed independently with the same inputs — i.e. matches what JUnit parsing/server-side ingestion would compute for the same test.
- [x] Added `test_convert_case_to_test_keeps_report_supplied_id_without_variant` as a companion/regression guard: with no variant, the report-supplied id is used unchanged (no behavior change for non-variant uploads).
- [x] Verified the new test fails against the pre-fix code (reverted the fix locally, confirmed `test_convert_case_to_test_folds_variant_into_report_supplied_id` fails with `left == right` on the raw id) and passes with the fix applied.
- [x] `cargo test -p trunk-analytics-cli --lib` — 20 passed, 0 failed.
- [x] `cargo fmt -p trunk-analytics-cli -- --check` clean.
- [x] `cargo clippy -p trunk-analytics-cli --lib --tests --no-deps` — no new findings introduced by this change (all output is pre-existing, unrelated lint debt elsewhere in the crate).

---
_Generated by [Claude Code](https://claude.ai/code/session_01G2HLZJPJWaHp7vFH3frr9k)_

### DIFF
--- a/cli/src/upload_command.rs
+++ b/cli/src/upload_command.rs
@@ -512,7 +512,7 @@ pub async fn run_upload(
             upload_args.xcresult_path.is_none(),
             #[cfg(not(target_os = "macos"))]
             true,
-            upload_args.variant,
+            meta.variant.clone(),
             &upload_args.org_url_slug,
             &quarantine_context.repo,
             &quarantined_test_ids,
@@ -528,7 +528,7 @@ pub async fn run_upload(
             upload_args.xcresult_path.is_none(),
             #[cfg(not(target_os = "macos"))]
             true,
-            upload_args.variant,
+            meta.variant.clone(),
             &upload_args.org_url_slug,
             &quarantine_context.repo,
             &quarantined_test_ids,
--- a/cli/src/context_quarantine.rs
+++ b/cli/src/context_quarantine.rs
@@ -127,7 +127,16 @@ fn convert_case_to_test<T: AsRef<str> + ToString>(
         if id.is_empty() {
             test.set_id(org_slug, repo, variant);
         } else {
-            test.id = id.clone();
+            // `id` here is a stable identifier supplied by the report source (e.g.
+            // xcresult's per-test-method identifier), which does not itself encode
+            // `variant`. Route it through `generate_custom_uuid` rather than assigning
+            // it as-is, so the id folds in `variant` the same way `gen_info_id` does
+            // for every other test (see `JunitParser::into_test_case_runs`, and
+            // `bundle-ingestion-lib` on the server) -- otherwise this locally computed
+            // id diverges from the variant-scoped id the server tracks, so quarantine
+            // lookups here never match and the printed test link resolves to the
+            // wrong (variant-less) test.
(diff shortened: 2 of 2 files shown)


cli/src/upload_command.rs
```diff
@@ -512,7 +512,7 @@ pub async fn run_upload(
             upload_args.xcresult_path.is_none(),
             #[cfg(not(target_os = "macos"))]
             true,
-            upload_args.variant,
+            meta.variant.clone(),
```
TylerJang27, line 515: Truncated to 64 chars, rather than the raw val we get from the upload_args. This shouldn't actually change anything downstream, but it matches our usage elsewhere

dfrankland (approved):

PR #1147, opened 2026-07-24
feat(quarantine): send test collection id and record resolution mode

max-trunk:
## What

Wires the CLI to the test-collection additions in the public API's `POST /metrics/getQuarantineConfig` endpoint (server side: trunk-io/trunk#33132), across **both** the traditional and rspec invocation paths:

1. Sends `--test-collection-id` as `testCollectionShortId` on the request.
2. Records the response's new `quarantineResolutionMode` (`repo` | `test_collection`) in telemetry and logs the resolved source.

The `--no-repo` flag / `noRepo` request param are intentionally **out of scope** for this PR.

## Changes

- **`api/src/message.rs`**
  - Add `test_collection_short_id` to `GetQuarantineConfigRequest` (`testCollectionShortId`, omitted when `None`).
  - Add a typed `QuarantineResolutionMode` enum (`Repo` | `TestCollection` | `Unspecified`) as the `quarantine_resolution_mode` field on `GetQuarantineConfigResponse` (`#[serde(default)]`). Its `Deserialize` is **tolerant**: any unknown, absent, or non-string value degrades to `Unspecified` rather than failing the whole response parse, so a future server value never breaks older clients. A single `From<QuarantineResolutionMode>` for the proto enum centralizes the mapping.
- **Request sourcing (both paths)** — the id comes from the `--test-collection-id` flag / `TRUNK_TEST_COLLECTION_ID` env var, threaded into the request on the traditional path (`cli/src/upload_command.rs` → `cli/src/context.rs`) and read directly on the rspec path (`test_report/src/report.rs`). Empty values are filtered so an empty env var doesn't send `""`.
- **Telemetry** — new proto `QuarantineResolutionMode` enum + `UploadMetrics.quarantine_resolution_mode = 9` in `proto/proto/upload_metrics.proto`; carried through `QuarantineContext` (traditional) and a new `RunUploadOptions.quarantine_resolution_mode_override` (rspec).
- **Logging** — on a successful lookup a human-readable `info` log names the resolved source (`Resolved quarantine status for test collection <id>` / `... for repo <name>`), emitted at most once per report (not per test); disk-cache hits are tagged `(cached)`.
- **rspec disk cache** — the resolution mode is stored in the cache entry and restored on a cache hit (so telemetry is correct without a re-fetch), and the collection id is part of the cache key so runs with different ids in the same repo can't collide.

## Compatibility

- **Response field is safe with older CLIs** — `GetQuarantineConfigResponse` has no `deny_unknown_fields`, so the extra key is ignored; the server PR can land first.
- **Unknown / non-string modes are safe** — the tolerant deserializer degrades anything unrecognized to `Unspecified`.
- **Telemetry field is safe with an un-updated server** — telemetry is protobuf-encoded, and a new field tag (`= 9`) is skipped by servers that don't know it.

## Testing

- ✅ `cargo check` and the full test suite pass across `api`, `trunk-analytics-cli`, and `test_report`.
- `cli/tests/upload.rs` asserts that `--test-collection-id` reaches the `getQuarantineConfig` request (guards the traditional-path wiring).

## Follow-up

- Confirm the rspec Ruby gem forwards `--test-collection-id` / sets `TRUNK_TEST_COLLECTION_ID` (no Rust-side change needed).

### DIFF
--- a/api/src/message.rs
+++ b/api/src/message.rs
@@ -24,16 +24,52 @@ pub struct CreateBundleUploadResponse {
     pub test_collection_bundle_meta_created_at: Option<String>,
 }
 
+/// Which source the server resolved quarantine status from, mapped from the
+/// getQuarantineConfig response's `quarantineResolutionMode` string. Unknown or
+/// absent values deserialize to `Unspecified` so a new server-side value never
+/// breaks deserialization on older clients.
+#[derive(Debug, Serialize, Clone, Copy, PartialEq, Eq, Default)]
+#[serde(rename_all = "snake_case")]
+pub enum QuarantineResolutionMode {
+    Repo,
+    TestCollection,
+    #[default]
+    Unspecified,
+}
+
+impl<'de> Deserialize<'de> for QuarantineResolutionMode {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        Ok(
+            match Option::<String>::deserialize(deserializer)?.as_deref() {
+                Some("repo") => Self::Repo,
+                Some("test_collection") => Self::TestCollection,
+                _ => Self::Unspecified,
+            },
+        )
+    }
+}
+
+impl From<QuarantineResolutionMode> for proto::upload_metrics::trunk::QuarantineResolutionMode {
+    fn from(mode: QuarantineResolutionMode) -> Self {
+        match mode {
+            QuarantineResolutionMode::Repo => Self::Repo,
+            QuarantineResolutionMode::TestCollection => Self::TestCollection,
+            QuarantineResolutionMode::Unspecified => Self::Unspecified,
+        }
+    }
+}
+
 #[derive(Debug, Serialize, Clone, Deserialize, Default)]
 #[serde(rename_all = "camelCase")]
 pub struct GetQuarantineConfigResponse {
     pub is_disabled: bool,
     #[serde(rename = "testIds")]
     pub quarantined_tests: Vec<String>,
-    /// Which source the server used to resolve quarantine status: "repo" or
-    /// "test_collection". Absent on older servers, hence `Option` + `default`.
     #[serde(default)]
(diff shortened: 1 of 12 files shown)


api/src/message.rs
```diff
@@ -30,6 +30,10 @@ pub struct GetQuarantineConfigResponse {
     pub is_disabled: bool,
     #[serde(rename = "testIds")]
     pub quarantined_tests: Vec<String>,
+    /// Which source the server used to resolve quarantine status: "repo" or
+    /// "test_collection". Absent on older servers, hence `Option` + `default`.
```
TylerJang27:
```suggestion
    /// "test_collection". Absent on older servers
```
:)

cli/src/context_quarantine.rs
```diff
@@ -336,14 +351,28 @@ pub async fn gather_quarantine_context(
     {
         tracing::info!("Checking if failed tests can be quarantined");
         match api_client.get_quarantining_config(request).await {
-            anyhow::Result::Ok(response) => (Some(response), QuarantineFetchStatus::FetchSucceeded),
+            anyhow::Result::Ok(response) => {
+                // Surface which source the server resolved quarantine status from. Sent to Sentry
+                // only (hidden from the console) to aid debugging without adding CLI noise.
```
TylerJang27: Imo would be preferable to include a visible log for it if test collection id is set

test_report/src/report.rs
```diff
@@ -551,6 +568,16 @@ impl MutTestReport {
                 for quarantined_test_id in response.quarantined_tests.iter() {
                     quarantined_tests.insert(quarantined_test_id.clone(), true);
                 }
+                // Surface which source the server resolved quarantine status from. Sent to Sentry
+                // only (hidden from the console) to aid debugging without adding CLI noise.
```
TylerJang27: Likewise

TylerJang27 (approved):

dfrankland (approved):

PR #1148, opened 2026-07-27
feat(quarantine): persist quarantine resolution mode in meta.json

max-trunk:
## Why

#1147 taught the CLI to parse `quarantineResolutionMode` off the `/metrics/getQuarantineConfig` response — whether the server resolved quarantine status from the repo or the test collection. That value currently reaches only a `tracing::info!` line and `UploadMetrics` telemetry, so it can't be tied back to a specific upload.

Writing it into `meta.json` lets the services side persist it per upload.

The config fetch completes before `upload_bundle` serializes `meta.json` into the tarball, so the value can ride along in the bundle rather than on the `createBundleUpload` request.

## What

- **Moves `QuarantineResolutionMode` from `api` into `bundle`** (new `bundle/src/quarantine_resolution_mode.rs`). `BundleMetaBaseProps` can't hold an `api` type — `api` depends on `bundle`, so that's a cycle — and consumers of the meta schema depend on `bundle`/`context`/`proto` without `api`. `api::message` re-exports it, so existing call sites are untouched; the `From<_> for proto::…` impl moves with the enum, since leaving it behind violates the orphan rule.
- **Adds `#[serde(default)] quarantine_resolution_mode` to `BundleMetaBaseProps`.**
- **`run_upload`** resolves the mode before `upload_bundle` and converts to the proto enum only at the telemetry site, so `UploadMetrics` stays byte-identical.
- **RSpec path**: `quarantine_resolution_mode_for_telemetry` returns the bundle-level enum and drops its now-inaccurate suffix, since it feeds `meta.json` too.

## Why no new `BundleMetaV0_7_9`

`#[serde(default)]` on the shared base props is load-bearing: `parse_meta` tries `V0_7_8` and falls back through six older structs that *all* flatten these same props, so a non-defaulted field fails every arm for bundles from older CLIs. Precedent is one line up — `test_collection` is a `#[serde(flatten, default)]` addition to the same struct. A new variant would cost a `VersionedBundle` arm plus a downgrade impl per version and buy nothing: `cli_version` already distinguishes a pre-field CLI from a genuine `Unspecified`. There's no `deny_unknown_fields` anywhere, so older readers ignore the new key.

## Testing

- `cli/tests/upload.rs` asserts a `test_collection` response reaches the extracted `meta.json` **and** that telemetry still carries the same value.
- Both bindings' fixtures round-trip the new field, verified by building the real artifacts (`wasm-pack build`, `maturin develop`):
  - `context-js` compares a parsed bundle against its input fixture. Without the fixture field, 6 tests fail — a *runtime* deep-equality diff, not a type error, since tsify renders the field optional because of `#[serde(default)]`. 28/28 pass.
  - `context-py`'s roundtrip test compares dumped JSON against an expected dict, so the defaulted key goes on the expected side (mirroring `test_collection_id`). 35/35 pass.
- Unrelated pre-existing flake: `test_report/tests/publish.rs` fails exactly one of `publish_environment_variable_overrides` / `publish_variant_priority_constructor_over_env` per run, and which one varies between runs — reproduced on unmodified `main`.

### DIFF
--- a/api/src/message.rs
+++ b/api/src/message.rs
@@ -1,3 +1,4 @@
+pub use bundle::QuarantineResolutionMode;
 use bundle::Test;
 use context::repo::RepoUrlParts;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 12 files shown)


api/src/message.rs
max-trunk, line 1: this file's diff is just moving code to the `bundle` crate to avoid circular import

bundle/src/quarantine_resolution_mode.rs
```diff
@@ -0,0 +1,65 @@
...
+                test_collection_id.unwrap_or("unknown"),
+            )),
+            Self::Repo => Some(format!(
+                "Resolved quarantine status for repo {}",
+                repo.repo_full_name()
+            )),
+            Self::Unspecified => None,
+        }
+    }
+}
```
TylerJang27, line 65: I know this is just ported but optional note on copy:
`Quarantine status applied from test collection {}`

`Quarantine status applied from repo {}`

Resolved imo doesn't make much sense to the user since that's a bit internal to us

bundle/src/quarantine_resolution_mode.rs
```diff
@@ -0,0 +1,65 @@
...
+    fn from(mode: QuarantineResolutionMode) -> Self {
+        match mode {
+            QuarantineResolutionMode::Repo => Self::Repo,
+            QuarantineResolutionMode::TestCollection => Self::TestCollection,
+            QuarantineResolutionMode::Unspecified => Self::Unspecified,
+        }
+    }
+}
+
+impl QuarantineResolutionMode {
```
TylerJang27, line 47: I love rust sometimes

TylerJang27 (approved):

